PR #27, opened 2022-07-20
Updated Schema

swalker2m:
This is not yet used, but I wanted to add the latest API for reference.  I left the original in place since it corresponds to the parts that are actually implemented.


tpolecat:
:+1:

PR #33, opened 2022-07-21
Fix unused warnings

swalker2m:
Is it okay to remove these compile warnings?


tpolecat:
haha wow, jinx

tpolecat:
In my PR I also enable fatal warnings and upgrade sbt, should we use that one instead?

swalker2m:
Sounds good, I'll close.

PR #43, opened 2022-07-28
Grease the wheels

swalker2m:
Makes a couple of updates to simplify running the service from SBT.

* Adds `sbt-revolver` to enable `service/reStart` and `service/reStop` for forking the service
* Rejiggers logging a bit because otherwise log messages were always being prepended with `[error]` after the previous change
* Adds an application flag `-skip-migration` to allow the service to be started without running migration.

If any of these changes rankle, maybe we can find better solutions or else I'll pull them out.


tpolecat (approved):
This all looks fine. I have never used revolver. :+1:

PR #44, opened 2022-07-28
move to modern schema

tpolecat:
Ok, this gets all the old tests running on the new schema, with the exception of the `programEdits` subscription test, which is disabled since it fails sporadically and needs closer attention.


modules/service/src/main/scala/lucuma/odb/graphql/snippet/input/TargetPropertiesInput.scala
```diff
@@ -0,0 +1,45 @@
...
+        ExistenceBinding.Option("existence", rExistence)
+      ) => (rName, rSidereal, rNonsidereal, rSourceProfile, rExistence).parTupled.flatMap {
+        case (name, sidereal, nonsidereal, sourceProfile, existence) =>
+          (name, sourceProfile, sidereal, nonsidereal) match {
+            case (None, _, _, _)             => Result.failure("Target name is required on creation.")
+            case (_, None, _, _)             => Result.failure("Source Profile is required on creation.")
+            case (Some(t), Some(p), Some(s), None) => Result(TargetPropertiesInput(t, Left(s), p, existence.getOrElse(Existence.Default)))
+            case (Some(t), Some(p), None, Some(n)) => Result(TargetPropertiesInput(t, Right(n), p, existence.getOrElse(Existence.Default)))
+            case (_, _, Some(_), Some(_))    => Result.failure("Found both sidereal and nonsidereal tracking; only one may be provided.")
+            case (_, _, None, None)          => Result.failure("Found neither sidereal noe nonsidereal tracking; exactly one must be provided.")
```
swalker2m, line 41: This is true when creating a new target, but not when editing since you could just be changing the name, source profile and/or existence.

swalker2m (approved):
Feels like we're making progress.  :+1:

modules/service/src/main/scala/lucuma/odb/graphql/snippet/input/TargetPropertiesInput.scala
tpolecat: Right, I will need to split it up for create/edit. Can't handle edits yet.

PR #45, opened 2022-07-28
WIP on updates to Program

tpolecat:
This adds more functionality for programs.

- Adds the `Program.plannedTime` field.
- Adds the `Program.observations` field.
- Adds the `Query.programs` query (details below).

Remaining things to do:

- [ ] `LIKE` and `NLIKE` don't work yet
- [ ] `LIMIT` doesn't work yet
- [ ] Return type is here is just `[Program!]!` because the `{ matches, hasMore }` struct needs some low-level support that @milessabin will probably need to look at. 

I will do `Mutation.updatePrograms` in another PR

#### Query.programs

I can now write this

```graphql
query {
  programs(
    OFFSET: "p-100"
    WHERE: {
      OR: [
        { name: { IS_NULL: false, IN: ["Foo", "Bar"] } }
        { id: { GTE: "p-101" } }
      ]
    }
  ) {
    id
    name
  }
}
```

and the generated SQL looks like this (I have indicated where the query arguments and defaults show up ... the remaining complexity is for the access control check).

```sql
SELECT
  t_program.c_name,
  t_program.c_program_id
FROM
  t_program
  INNER JOIN (
    SELECT
      DISTINCT ON ((t_program.c_program_id COLLATE "C"))(t_program.c_program_id COLLATE "C")
    FROM
      t_program
      LEFT JOIN t_program_user ON (
        t_program_user.c_program_id = t_program.c_program_id
      )
    WHERE
      (
        (
          (
            (t_program.c_program_id >= $1)     -- WHERE
            OR (                               -- WHERE
              t_program.c_name IS NOT NULL     -- WHERE
              AND t_program.c_name IN ($2, $3) -- WHERE
            )                                  -- WHERE
          )
          AND (
            (
              (t_program_user.c_user_id = $4)
              OR (t_program.c_pi_user_id = $5)
            )
            AND (
              (t_program.c_existence = $6)        -- INCLUDE DELETED
              AND (t_program.c_program_id >= $7)  -- OFFSET
            )
          )
        )
      )
      AND (t_program.c_program_id IS NOT NULL)
    ORDER BY
      (t_program.c_program_id COLLATE "C") NULLS LAST
  ) AS t_program_pred ON (
    t_program_pred.c_program_id = t_program.c_program_id
  )
```


modules/service/src/main/scala/lucuma/odb/graphql/snippet/input/WhereProgram.scala
```diff
@@ -0,0 +1,39 @@
...
+        ("proposal", _), // ignore for now
+      ) =>
+        (rAND, rOR, rNOT, rId, rName).parMapN { (AND, OR, NOT, id, name) =>
+          and(List(
+            AND.map(and),
+            OR.map(or),
+            NOT.map(Not(_)),
+            id,
+            name
+          ).flatten)
```
swalker2m, line 35: Wow, this is a lot less code than I would have guessed.

swalker2m (commented):
Looking great!

modules/service/src/main/scala/lucuma/odb/graphql/snippet/input/WhereProgram.scala
tpolecat: Yeah it maps directly onto the Grackle query structure, which is convenient.

tpolecat:
@swalker2m  I marked this as ready for review in case you want to merge it while I'm away.

swalker2m:
Excellent thanks, I'll take a look.

swalker2m (approved):
This looks great.  I know there are some unfinished steps but I'll merge since you'll be away for a while.

PR #53, opened 2022-08-03
Update Observation

swalker2m:
Work in progress on `updateObservations`.  It works (ish) but accepts a list of observation ids instead of a `WHERE` clause for observations.  I wrote the `WhereObservation` and used it to good effect in the `observations` query, but I'm not sure how to employ it in `updateObservations`.

It seems as though I'll need to write a `case class WhereObservation` to hold the various options in `WhereObservation` and an appropriate binding to fill it out (as opposed to the existing `Matcher[Predicate]` used with the `observations` query).  Then I'll need to update the `ObservationService` to do the select manually within the `ObservationService.updateObservations` query.

Since that seems wrong (?) I wanted to stop and get some feedback before going down that path.


modules/service/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -1773,7 +1773,10 @@ type Mutation {
   updateObservations(
     # Parameters for editing existing observations.
     input: UpdateObservationsInput!
-  ): UpdateObservationsResult!
+  ): [Observation!]!
```
swalker2m, line 1776: Pending a Miles update to Grackle, we just return a list here.

modules/service/src/main/scala/lucuma/odb/data/Nullable.scala
```diff
@@ -23,6 +23,15 @@ sealed trait Nullable[+A] extends Product with Serializable {
...
   def toOption: Option[A] = fold(None, None, Some(_))
 
+  def isNull: Boolean =
+    fold(ifNull = true, ifAbsent = false, _ => false)
+
+  def isAbsent: Boolean =
+    fold(ifNull = false, ifAbsent = true, _ => false)
+
+  def isPresent: Boolean =
+    fold(ifNull = false, ifAbsent = false, _ => true)
```
swalker2m, line 33: I didn't end up using these or the `orNull` or `orAbsent` constructors.  We can delete them if it meets opposition.

modules/service/src/main/scala/lucuma/odb/graphql/snippet/ObservationSnippet.scala
```diff
@@ -218,31 +270,59 @@ object ObservationSnippet {
...
+                  OFFSET.map(oid => GtEql(UniquePath(List("id")), Const(oid))).getOrElse(True),
+                  Predicates.includeDeleted(includeDeleted),
+                  Predicates.isVisibleTo(user),
+                  WHERE.getOrElse(True)
+                ),
+                child
+                // Limit(
+                //    LIMIT.foldLeft(1000)(_ min _.value),
+                //   child
+                // )
```
swalker2m, line 308: Pending a Grackle bug fix, I believe.

modules/service/src/main/scala/lucuma/odb/service/ObservationService.scala
```diff
@@ -169,6 +229,174 @@ object ObservationService {
...
+
+      val upExistence = sql"c_existence = $existence"
+      val upSubtitle  = sql"c_subtitle = ${text_nonempty.opt}"
+      val upStatus    = sql"c_status = $obs_status"
+      val upActive    = sql"c_active_status = $obs_active_status"
+
+      val upCloud     = sql"c_cloud_extinction = $cloud_extinction"
+      val upImage     = sql"c_image_quality = $image_quality"
+      val upSky       = sql"c_sky_background = $sky_background"
+      val upWater     = sql"c_water_vapor = $water_vapor"
```
swalker2m, line 333: I've also punted on updating the elevation range.  This is an instance where the update you supply has to be consistent with the existing elevation range.  For example, if you set just the minimum airmass then the observation must have a definition already for maximum airmass (as opposed to defining an hour angle range).  I'm not sure how to handle this.  Attempt the update and let the table constraint fail?  Filter out the observations where the update would fail?

modules/service/src/main/scala/lucuma/odb/graphql/snippet/ObservationSnippet.scala
tpolecat: You can fix this by swapping the order; i.e., `Limit(..., Filter(..., child))`

tpolecat (approved):
This looks good. Maybe fix the LIMIT thing and merge as-is before I start looking at Scala 3?

swalker2m:
> Maybe fix the LIMIT thing and merge as-is
Will do.

PR #56, opened 2022-08-08
Query.programs LIKE, NLIKE, LIMIT

tpolecat:
This fixes `LIMIT` on `Query.programs` and `LIKE`/`NLIKE` on `WhereOptionString`.

It also somewhat randomly fixes query logging so as not to cruft up the test output, and it re-enables a failing test that now passes.


tpolecat:
Oops didn't mean for this to be a draft. It's ready for judgment.

modules/service/src/main/scala/lucuma/odb/graphql/snippet/input/WhereOptionString.scala
```diff
@@ -24,18 +28,20 @@ object WhereOptionString {
...
         (rIsNull, rEq, rNeq, rIn, rNin, rLike, rNlike, rMatchCase).mapN {
-          (isNull, eq, neq, in, nin, _, _, _) =>
+          (isNull, eq, neq, in, nin, like, nlike, matchCase) =>
           and(List(
             isNull.map(IsNull(path, _)),
             eq.map(a => Eql(path, Const(a.some))),
             neq.map(a => NEql(path, Const(a.some))),
             in.map(as => In(path, as.map(_.some))),
             nin.map(as => Not(In(path, as.map(a => a.some)))),
+            like.map(s => Like(path, s.value, !matchCase.getOrElse(true))),
```
swalker2m, line 43: Is this what I expect with `%` and `?` wildcards?

modules/service/src/main/scala/lucuma/odb/graphql/snippet/BaseSnippet.scala
```diff
@@ -19,14 +19,20 @@ object BaseSnippet {
     val NonEmptyStringType = schema.ref("NonEmptyString")
     val ExistenceType = schema.ref("Existence")
     val BigDecimalType = schema.ref("BigDecimal")
+    val LongType = schema.ref("Long")
     val PosBigDecimalType = schema.ref("PosBigDecimal")
+    val HmsStringType = schema.ref("HmsString")
+    val DmsStringType = schema.ref("DmsString")
```
swalker2m, line 25: I'll have to do this too, I suspect.

swalker2m (approved):
Looks good. :+1:

modules/service/src/main/scala/lucuma/odb/graphql/snippet/input/WhereOptionString.scala
tpolecat: Yes, it compiles down to `LIKE` and `ILIKE` and the const argument is passed verbatim.

PR #57, opened 2022-08-08
WIP on target-env

swalker2m:
This is kind of a mishmash of progress on observations, but I think I'd like to get it in before working on creating/updating target environments.  I believe the way that observations are updated needs to be changed to take into account the target observation's existing state.

Here's a summary of what is included in this PR

1. Target environment explicit base in observation
2. Link table for observations/targets (asterisms)
3. Remove extraneous `includeDeleted` flag from observation -> program lookup
4. Shared RA/Dec mapping (for asterism explicit base + sidereal targets)
5. Target ID mapping bug fix
6. WHERE clause in observation update

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -59,9 +59,6 @@ lazy val service = project
       "com.github.vertical-blank" % "sql-formatter" % "2.0.3",
     ),
     reStart / envVars += "PORT" -> "8082",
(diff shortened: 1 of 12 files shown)


tpolecat:
My thinking was wrong. That looks good. :+1:

PR #58, opened 2022-08-09
Query.updatePrograms

tpolecat:
This adds bulk editing of programs. The user-specified `WHERE` predicate is compiled to SQL and is used directly in the update (as a subselect for an `IN` clause).


modules/service/src/main/scala/lucuma/odb/graphql/snippet/ProgramSnippet.scala
```diff
@@ -131,18 +137,30 @@ object ProgramSnippet {
...
+          // An applied fragment that selects all program ids that satisfy `filterPredicate`
+          val idSelect: Result[AppliedFragment] =
+            Result.fromOption(
+             MappedQuery(Filter(filterPredicate, Select("id", Nil, Query.Empty)), Cursor.Context(ProgramType)).map(_.fragment),
+              "Could not construct a subquery for the provided WHERE condition." // shouldn't happen
+            )
+
+          // Update the specified programs and then return a query for the same set of programs.
+          idSelect.traverse { which =>
+            pool.use(_.updatePrograms(input.SET, which)).as(Filter(filterPredicate, child))
```
swalker2m, line 160: Nice.

swalker2m (approved):
Looks great!

PR #59, opened 2022-08-09
Scala 3

tpolecat:


modules/service/src/main/scala/lucuma/odb/graphql/util/SnippetMapping.scala
```diff
@@ -45,7 +45,7 @@ trait SnippetMapping[F[_]] extends Mapping[F] {
   private implicit val SemigroupTypeMapping: Semigroup[TypeMapping] = (a, b) =>
     (a, b) match {
       case (a: ObjectMapping, b: ObjectMapping) => a |+| b
-      case (a: LeafMapping[_], _: LeafMapping[_]) => a
+      case (a: LeafMapping[x], _: LeafMapping[y]) => a
```
swalker2m, line 48: Hmm, what is this about?

swalker2m (approved):
Excellent.

modules/service/src/main/scala/lucuma/odb/graphql/util/SnippetMapping.scala
tpolecat: Scala 3 doesn't like existentials. There may be a better way to do this.

PR #62, opened 2022-08-11
Reorg!

tpolecat:
Ok this splits up the existing snippets into per-type mappings, which should help us keep complexity down a bit. The best way to review is probably going to be to pull the branch and look around.


tpolecat:
Oh I'm using some snapshots. Standby.

swalker2m (approved):
I did as suggested and looked through the branch. It seems much cleaner this way, so by all means :+1:

PR #65, opened 2022-08-12
remove graphql package object

tpolecat:
This removes `package object graphql` as follows:
- syntax for grackle `Result` and `Env` have been upstreamed to grackle (which is updated here)
- syntax for `Codec` has been moved into the `tables` package
- bindings have been moved into the `bindings` package


swalker2m (approved):
looks great, pending whatever import cleanup or whatnot that is missing

PR #66, opened 2022-08-12
Remove duplicate enums

tpolecat:
This removes `lucuma.odb.data.ObsStatus`/`ObsActiveStatus` in favor of the enums in `lucuma.core.enums`.

Waiting on updated core with permanent tag definitions.


swalker2m (approved):
This looks good, any reason not to merge?

tpolecat:
Don't want to merge this until the new core is released because the tags used in the disabled tests currently disagree with the serial format (which is based on `productPrefix` at the moment).

swalker2m:
Copied into #103 to avoid conflict resolution.

PR #67, opened 2022-08-15
WIP: edit events, full model

tpolecat:
This adds event type and event id to program update events and establishes a pattern for other event types (pass the object id, event id, and edit type in the channel event and stuff it in the query env).

In order to make the test more stable I lengthened the amount of time we wait for events to arrive. We'll see if it's sufficient.


swalker2m (approved):
Looks good. :+1:

PR #69, opened 2022-08-15
Query.target,targets

tpolecat:
Adds some more queries and basic tests.


tpolecat:
Ok this is ready.

We need to do some refactoring on tests to remove duplicated stuff like program creation, but that can be done as another PR.

swalker2m (approved):
:+1:

PR #73, opened 2022-08-17
Elevation range edit

swalker2m:
Adds elevation range editing.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/ObservationService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/ObservationService.scala
@@ -3,10 +3,11 @@
 
 package lucuma.odb.service
 
-import cats.data.NonEmptyList
+import cats.data.{Ior, NonEmptyList}
 import cats.effect.Sync
 import cats.syntax.all.*
-import edu.gemini.grackle.Predicate
+import edu.gemini.grackle.{Predicate, Problem, Result}
+import eu.timepit.refined.api.Refined.value
 import eu.timepit.refined.types.numeric.NonNegInt
 import eu.timepit.refined.types.numeric.PosBigDecimal
 import eu.timepit.refined.types.string.NonEmptyString
@@ -16,6 +17,8 @@ import lucuma.core.enums.SkyBackground
 import lucuma.core.enums.WaterVapor
 import lucuma.core.model.ConstraintSet
 import lucuma.core.model.ElevationRange
+import lucuma.core.model.ElevationRange.AirMass.DecimalValue
+import lucuma.core.model.ElevationRange.HourAngle.DecimalHour
 import lucuma.core.model.GuestRole
 import lucuma.core.model.Observation
 import lucuma.core.model.Program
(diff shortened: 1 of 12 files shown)


modules/service/src/main/scala/lucuma/odb/service/ObservationService.scala
```diff
@@ -29,38 +35,78 @@ import lucuma.odb.data.Nullable.NonNull
...
+
+  val ConstraintViolationMessages: List[(String, String)] =
+    List(
+      MissingAirMassConstraint   -> MissingAirMassConstraintMessage,
+      MissingHourAngleConstraint -> MissingHourAngleConstraintMessage
+    )
+
+  def constraintViolationMessage(ex: PostgresErrorException): String =
+    ConstraintViolationMessages
+      .find { case (c, _) => ex.message.contains(c) }
```
swalker2m, line 90: Couldn't see a better way to figure out which constraint was violated.

modules/service/src/main/scala/lucuma/odb/service/ObservationService.scala
```diff
@@ -197,16 +256,42 @@ object ObservationService {
...
+
+        val hourAngle: List[AppliedFragment] =
+          List(
+            e.hourAngle.flatMap(_.minBigDecimal).map(v => upHourAngleMin(v.some)),
+            e.hourAngle.flatMap(_.maxBigDecimal).map(v => upHourAngleMax(v.some))
+          ).flattenOption
+
+        (airMass, hourAngle) match {
+          case (Nil, Nil) => List.empty[AppliedFragment].asRight
+          case (am,  Nil) => (upHourAngleMin(None) :: upHourAngleMax(None) :: am).asRight
```
swalker2m, line 290: when you switch from one kind of elevation constraint to another, you have to wipe out the opposing constraint type

tpolecat (approved):
:+1:

modules/service/src/main/scala/lucuma/odb/service/ObservationService.scala
```diff
@@ -220,26 +305,33 @@ object ObservationService {
...
-    ): Option[AppliedFragment] =
-      updates(SET).map { us =>
-        void"UPDATE t_observation " |+|
-        void"SET " |+| us.intercalate(void", ") |+| void" " |+|
-        void"WHERE t_observation.c_observation_id IN (" |+| which |+| void")" |+|
-        void"RETURNING t_observation.c_observation_id"
+    ): Either[String, Option[AppliedFragment]] =
+      updates(SET).map {
+        _.map { us =>
+          void"UPDATE t_observation " |+|
```
cquiroz, line 330: Is this skunk syntax?

swalker2m: Yeah, `void` makes a Skunk `AppliedFragment` out of a bit of SQL that doesn't involve any parameters.  The `sql` interpolator is used when there are query arguments.  At least, that's my understanding.

PR #77, opened 2022-08-19
Target Environment Explicit Base

swalker2m:
This is the second (see #57) of several updates to add target environment.  In this installment, I add the optional `explicitBase` position.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/graphql/input/CoordinatesInput.scala
+++ b/modules/service/src/main/scala/lucuma/odb/graphql/input/CoordinatesInput.scala
@@ -1,0 +1,49 @@
+// Copyright (c) 2016-2022 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.odb.graphql
+package input
+
+import cats.data.ValidatedNec
+import cats.syntax.option.*
+import cats.syntax.parallel.*
+import cats.syntax.validated.*
+import edu.gemini.grackle.Result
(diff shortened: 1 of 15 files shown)


modules/service/src/main/scala/lucuma/odb/graphql/input/CoordinatesInput.scala
```diff
@@ -0,0 +1,52 @@
...
+  ra:  Option[RightAscension],
+  dec: Option[Declination]
+) {
+
+  def create: ValidatedNec[String, Option[Coordinates]] =
+    (ra, dec) match {
+      case (Some(r), Some(d)) => Coordinates(r, d).some.validNec
+      case (None, None)       => none.validNec
+      case _                  => CoordinatesInput.messages.BothRaAndDecNeeded.invalidNec
+    }
```
swalker2m, line 34: Potentially controversial, but `ValidatedNec` is closer to what I think we want for argument validation than `Ior`?

tpolecat: GraphQL allows you to return warnings along with the result, so I have been using `Result` for all of this stuff.

tpolecat: I guess it's not a huge deal either wat. Where is `toResult` defined?

swalker2m: Switched to `Result` and removed the little intermediate `CreateResult` and `UpdateResult` types since I just translated to those and then back again to `Result`.  That simplified it a bit.

PR #79, opened 2022-08-23
Target env part three

swalker2m:
Adds asterism creation upon observation creation.  Editing observation asterisms still pending.


modules/service/src/main/scala/lucuma/odb/service/AsterismService.scala
```diff
@@ -0,0 +1,110 @@
...
+      ): F[Result[Unit]] = {
+        val af = Statements.insertLinksAs(user, programId, observationId, targetIds.toList)
+        session.prepare(af.fragment.command).use { p =>
+          p.execute(af.argument)
+            .as(Result.unit)
+            .recoverWith {
+              // TODO: not catching the problem
+              case SqlState.ForeignKeyViolation(ex) =>
+                Result.failure(ex.getMessage).pure[F]
+            }
```
swalker2m, line 70: Not recovering ...

modules/service/src/main/scala/lucuma/odb/graphql/mapping/MutationMapping.scala
```diff
@@ -95,12 +101,34 @@ trait MutationMapping[F[_]: MonadCancelThrow]
...
+
+      pool.use { s =>
+        s.transaction.use { xa =>
+          for {
+            rTup  <- createObs
+            oid    = rTup.toOption.map(_._1)
+            rUnit <- insertAsterism(oid)
+            query  = (rTup, rUnit).parMapN { case ((_, query), _) => query }
+            // Fail altogether if there was an issue, say, creating the asterism
+            _     <- query.left.fold(Applicative[F].unit)(_ => xa.rollback.void)
```
swalker2m, line 129: Is this correct?  I assume I need an explicit transaction and rollback here.  I want to handle an exception generated when creating the asterism and turn it into a failure `Result` and yet have it rollback the whole observation creation.

Also, there's surely a cleaner way to do `Ior.left.fold(Applicative[F].unit)(_ => xa.rollback.void)`.

tpolecat (approved):
:+1: looks great!

modules/service/src/main/scala/lucuma/odb/graphql/mapping/MutationMapping.scala
swalker2m: Yep, thanks!

PR #86, opened 2022-08-26
Asterism update in observation edit

swalker2m:
Adds support for updating the `asterism` array in an observation edit.

```
mutation {
  updateObservations(input: {
    programId: "p-100"
    SET: {
      targetEnvironment: {
        asterism: [ "t-100", "t-101" ]
      }
      WHERE: {
        id: { EQ: "o-100" }
      }
    }
  }) {
    targetEnvironment {
      asterism {
        id
      }
    }
  }
}
```

Because the asterism array mentions specific target ids, and these targets must be associated with a single program, matching observations from multiple programs would always fail.  Accordingly I removed `programId` from the observation matching `WHERE` and put it into the `UpdateObservationsInput`.  Is this what we want to do?

### DIFF
--- a/modules/service/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/service/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -2469,6 +2469,10 @@ type UpdateDatasetsResult {
 
 # Observation selection and update description.  Use `SET` to specify the changes, `WHERE` to select the observations to update, and `LIMIT` to control the size of the return value.
 input UpdateObservationsInput {
+
+  # The program whose observations will be updated.
+  programId: ProgramId!
+
   # Describes the observation values to modify.
   SET: ObservationPropertiesInput!
 
@@ -5716,9 +5720,6 @@ input WhereObservation {
 
   # Matches the observation id.
   id: WhereOrderObservationId
-
-  # Matches the id of the program associated with this observation.
-  programId: WhereOrderProgramId
 
   # Matches the subtitle of the observation.
   subtitle: WhereOptionString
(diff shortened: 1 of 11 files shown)


modules/service/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -2469,6 +2469,10 @@ type UpdateDatasetsResult {
 
 # Observation selection and update description.  Use `SET` to specify the changes, `WHERE` to select the observations to update, and `LIMIT` to control the size of the return value.
 input UpdateObservationsInput {
+
+  # The program whose observations will be updated.
+  programId: ProgramId!
```
swalker2m, line 2474: Moved here from `WhereObservation`.  I'm unsure about this.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/MutationMapping.scala
```diff
@@ -202,14 +204,48 @@ trait MutationMapping[F[_]: MonadCancelThrow]
...
+                // non-empty list (eventually) so we need to catch the Nil
+                // case.
+
+                // Produces "Unable to map query":
+                // case Nil => Filter(False, child)
+
+                // Work around using a zero limit
+                case Nil => Limit(0, child)
+
+                case ids => Filter(ObservationPredicates.inObservationIds(ids), child)
```
swalker2m, line 228: `inObservationIds(ids)` is just `In(UniquePath(List("id")), oids)`.   It allows you to construct this with an empty list, but will fail when evaluated.   It seems like `In` should work with empty lists?  If not, `Filter(False, child)` should work in that it should match nothing.

I thought of `Limit(0, child)` afterwords and maybe that's fine anyway.

modules/service/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -1763,7 +1763,10 @@ type Mutation {
   updateAsterisms(
     # Bulk update asterism
     input: UpdateAsterismsInput!
-  ): UpdateAsterismsResult!
+  ): [Observation!]!
+# TODO: UpdateAsterismsResult
+#  ): UpdateAsterismsResult!
```
swalker2m, line 1768: Any news on when we'll be able to do this?

tpolecat (approved):
From the point of view of an Explore user this seems fine. We might want to circle back later and add an admin API for bulk updates that allows limited cross-program updates. :+1:

PR #90, opened 2022-08-30
Visualization Time

swalker2m:
Adds `visualizationTime` to observation creation and editing.  Times will be stored with the `timestamp` type in the database, which is limited to microsecond precision. The schema defined a scalar called `Instant` which implies a `java.time.Instant` which in turn implies nanosecond precision.  This PR introduces a corresponding `Timestamp` Scala type that can safely roundtrip to the database.  This type is then reflected in the GraphQL schema (replacing the `Instant` scalar  with a `Timestamp` scalar).


tpolecat:
Hm, munit-discipline should be available. I will double-check.

tpolecat (approved):
Looks great! :+1:

PR #92, opened 2022-08-31
wip: proposals

tpolecat:
This adds the proposal model and associated changes to program editing.


modules/service/src/main/resources/db/migration/V0090__proposals.sql
```diff
@@ -0,0 +1,84 @@
...
+  c_class             d_tag             not null  references t_proposal_class(c_tag),
+  c_min_percent       d_int_percentage  not null,
+  c_min_percent_total d_int_percentage  null      check((c_min_percent_total is not null) = (c_category = 'intensive' or c_category = 'large_program')),
+  c_totalTime         interval          null      check((c_totalTime is not null) = (c_category = 'intensive' or c_category = 'large_program'))
+);
+
+-- Partner Splits
+
+create table t_partner_split (
+  c_program_id        d_program_id      not null  references t_program(c_program_id) on delete cascade,
```
swalker2m, line 79: I guess I need to add cascading deletes.  I didn't initially because we don't plan on deleting programs, observations, or targets but of course we will at some point in an admin interface.

tpolecat: Yeah I happened to think of it but I don't think it's a big deal at this point.

modules/service/src/main/scala/lucuma/odb/service/ProgramService.scala
```diff
@@ -130,29 +163,54 @@ object ProgramService {
...
+                  .compile
+                  .toList
+                  .map(UpdateProgramResponse.Success(_))
+              }
+            }
+
+          // Update the proposal
+          val b: F[UpdateProgramResponse] =
+            SET.proposal.fold(UpdateProgramResponse.Success(Nil).pure[F]) {
+              proposalService.updateProposals(_, where, xa).map {
```
swalker2m, line 183: Will the `where` results differ here potentially from the `updatePrograms` case?

swalker2m (commented):
Looks great.  Lots of stuff for me to copy in there.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/NonNegDurationMapping.scala
```diff
@@ -25,13 +29,21 @@ trait NonNegDurationMapping[F[_]]
...
 
   private def valueAs[A: io.circe.Encoder](name: String)(f: Duration => A): CursorField[A] =
-    CursorField[A](name, c => c.fieldAs[Duration]("value").map(f), List("value"))
+    CursorField[A](name, c =>
+      c.fieldAs[Some[Duration]]("value") match {
+        case Ior.Right(Some(d)) => Result(f(d))
+        case _ => c.fieldAs[Duration]("value").map(f)
+      },
+      List("value")
+    )
```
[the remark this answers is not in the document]
tpolecat: It lets you implement a field directly by taking the cursor (which is zippering around in the result structure) and pulling values out of it and presenting them, typically transformed. This is what we're doing with `MappingExtras` and `FieldRef` ... I think this code is old, I'll change it to use `FieldRef`.

modules/service/src/main/scala/lucuma/odb/service/PartnerSplitsService.scala
```diff
@@ -0,0 +1,103 @@
...
+
+    def updateSplits(splits: Map[Tag, IntPercent], whichProgramIds: AppliedFragment, xa: Transaction[F]): F[List[Program.Id]]
+
+}
+
+object PartnerSplitsService {
+
+  /**
+   * Construct a `PartnerSplitsService` using the specified `Session`. This service is intended for
+  * indirect use by `ProgramService`, which is in charge of access control (which is unchecked here).
```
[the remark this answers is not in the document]
tpolecat: That comment is actually wrong. The checking happens in the "where" fragment ... it doesn't select the ids of programs you can't see. I'll update.

modules/service/src/main/scala/lucuma/odb/service/ProgramService.scala
tpolecat: Yes, there may not be any updates to program properties at al, in which case nothing is done by `a` above, which is also the case here. If both the program fields and the proposal fields are updated then yes, they should yield the same set of program ids.

modules/service/src/main/scala/lucuma/odb/service/extensions.scala
```diff
@@ -0,0 +1,11 @@
...
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.odb.service
+
+import cats.effect.MonadCancelThrow
+import skunk._
+import skunk.data.Completion
+
+extension [F[_]: MonadCancelThrow](s: Session[F]) def executeCommand(af: AppliedFragment): F[Completion] =
+  s.prepare(af.fragment.command).use(_.execute(af.argument))
```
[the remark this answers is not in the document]
tpolecat: I think I'm actually not using that (I will check) but yeah, something like that seems useful.

modules/service/src/main/scala/lucuma/odb/service/ProgramService.scala
swalker2m: i guess what i mean is, when the `where` fragment does its select it may end up selecting _different_ programs than those selected in the `a` case.  For example, if you select ones titled "Biff" and update them to "Buff" it will no longer find any "Biff"s.  Does that make sense?  Or am I reading it wrong?

tpolecat: Actually the way to do this is to select them into a temporary table and join to it. That way we're not sending the ids all over the place.

tpolecat: Ok I changed it to write a temp table containing the program ids selected by `where`, plus an indication of whether there is an existing proposal or not. Then I join with it to do all the follow-on updates. Still adding tests so I'll keep this unresolved for now.

modules/service/src/main/scala/lucuma/odb/service/ProgramService.scala
```diff
@@ -205,12 +222,14 @@ object ProgramService {
...
-        void"UPDATE t_program " |+|
-        void"SET " |+| us.intercalate(void", ") |+| void" " |+|
-        void"WHERE t_program.c_program_id IN (" |+| which |+| void") " |+|
-        void"RETURNING t_program.c_program_id"
+        void"""
+          UPDATE t_program
+          SET """ |+| us.intercalate(void", ") |+| void"""
+          FROM t_program_update WHERE t_program.c_program_id = t_program_update.c_program_id
+          RETURNING t_program.c_program_id
+        """
```
swalker2m, line 232: This seems good.  I'll do it in similar situations.  The only part that gives me pause is that elsewhere you have to assume that `t_program_update` exists already but nothing guarantees it?  This seems better than passing the list around though.

tpolecat:
Ok this is ready, pending CI. I addressed all the comments above.

swalker2m (approved):
Looks great!

PR #98, opened 2022-09-06
sort codecs

swalker2m:
Simply sorts the codecs in `Codecs`.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/util/Codecs.scala
+++ b/modules/service/src/main/scala/lucuma/odb/util/Codecs.scala
@@ -43,18 +43,8 @@ import skunk.data.Type
 // Codecs for some atomic types.
 trait Codecs {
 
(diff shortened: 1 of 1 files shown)


modules/service/src/main/scala/lucuma/odb/util/Codecs.scala
```diff
@@ -65,123 +55,137 @@ trait Codecs {
...
   val obs_active_status: Codec[ObsActiveStatus] =
     enumerated(Type("e_obs_active_status"))
 
-  // TEMPORARY; the tag will be correct soon
-  val too_activation: Codec[ToOActivation] =
-    `enum`(
-      e => e.label.toLowerCase(),
-      s => Enumerated[ToOActivation].all.find(_.label.toLowerCase == s),
-      Type("e_too_activation")
-    )
```
swalker2m, line 109: This should work now since the tag is now correct in the `lucuma-core` we're using.

PR #99, opened 2022-09-07
WIP on constraintSetGroups

tpolecat:
This is currently blocked on https://github.com/gemini-hlsw/gsp-graphql/issues/266 

This adds `Query.constraintSetGroups`, which is driven by a new view that collects unique configurations per program. It also introduces `SwitchMapping`, which is like `PrefixedMapping` but instead selects the underlying mapping on type + fieldname.


build.sbt
```diff
@@ -23,6 +23,7 @@ val postgresVersion            = "42.5.1"
 val skunkVersion               = "0.3.2"
 val lucumaSsoVersion           = "0.4.4"
 val testcontainersScalaVersion = "0.40.12"
+val paigesVersion              = "0.4.2"
```
tpolecat, line 26: This is for the GraphQL query pretty-printer, which we should upstream to Grackle once we're sure we like it. I added code to the main mapping to print the query out on each request.

modules/service/src/main/resources/db/migration/V0003__functions.sql
```diff
@@ -0,0 +1,3 @@
+-- https://stackoverflow.com/questions/54372666/create-an-immutable-clone-of-concat-ws/54384767#54384767
+CREATE OR REPLACE FUNCTION immutable_concat_ws(text, VARIADIC text[])
+RETURNS text AS 'text_concat_ws' LANGUAGE internal IMMUTABLE PARALLEL SAFE;
```
tpolecat, line 3: I'm using a generated column below, which is defined as a pure expression based on other fields in the row. I need to use the `text_concat_ws` function but it's not declared immutable for some reason . This basically just redeclares it with the right modifier.

modules/service/src/main/resources/db/migration/V0030__observations.sql
```diff
@@ -171,6 +171,22 @@ create table t_observation (
...
+      c_image_quality,
+      c_sky_background,
+      c_water_vapor,
+      coalesce(c_air_mass_min::text, 'null'),
+      coalesce(c_air_mass_max::text, 'null'),
+      coalesce(c_hour_angle_min::text, 'null'),
+      coalesce(c_hour_angle_max::text, 'null')
+    )
+   ) stored,
+
```
tpolecat, line 189: To group observations by conditions I concatenate all the relevant fields together and use that as the key. I need to do this so the constraint set group has a key in the obs table that it can join with.

modules/service/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -4699,6 +4695,9 @@ type Query {
 
   # Selects the first `LIMIT` matching observations based on the provided `WHERE` parameter, if any.
   observations(
+
+    programId: ProgramId
```
tpolecat, line 4699: At some point we decided that top-level queries for observations needed to specify the program id. I think we should get rid of these anyway and make the user traverse down to the program.

modules/service/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -5500,8 +5499,6 @@ type TargetEnvironmentGroupSelectResult {
 }
 
 type TargetGroup {
-  # IDs of observations that use the same constraints
-  observationIds: [ObservationId!]!
```
tpolecat, line 5504: This wasn't being used in the client and it's redundant with observations/matches/id so I removed it (same with other group types).

modules/service/src/main/scala/lucuma/odb/graphql/GraphQLRoutes.scala
```diff
@@ -52,6 +54,9 @@ object GraphQLRoutes {
                   {
                     for {
                       user <- OptionT(client.get(a))
+                      // If the user has never hit the ODB using http then there will be no user
+                      // entry in the database. So go ahead and [re]canonicalize here to be sure.
+                      _    <- OptionT.liftF(userSvc.canonicalizeUser(user))
```
tpolecat, line 59: I ran into this issue when we tested with Explore, which never hits the server via plain HTTP and thus never went through the normal user-canonicalizing code.

modules/service/src/main/scala/lucuma/odb/graphql/OdbMapping.scala
```diff
@@ -222,7 +230,11 @@ object OdbMapping {
               ).combineAll
             )
 
-          // Override `fetch` to log the query. This is optional.
+          // Override `defaultRootCursor` to log the GraphQL query. This is optional.
+          override def defaultRootCursor(query: Query, tpe: Type, env: Env): F[Result[(Query, Cursor)]] =
+            Logger[F].info("\n\n" + PrettyPrinter.query(query).render(100) + "\n") *> super.defaultRootCursor(query, tpe, env)
```
tpolecat, line 235: As with the override of `fetch` below this just prints stuff out and you can comment it out if it bothers you.

modules/service/src/main/scala/lucuma/odb/graphql/util/PrettyPrinter.scala
```diff
@@ -0,0 +1,128 @@
...
+import org.typelevel.paiges.Doc
+import org.typelevel.paiges.Doc.char
+import org.typelevel.paiges.Doc.intercalate
+import org.typelevel.paiges.Doc.line
+import org.typelevel.paiges.Doc.space
+import org.typelevel.paiges.Doc.str
+import org.typelevel.paiges.Doc.text
+
+/** Paiges combinators for Grackle queries. */
+object PrettyPrinter {
```
tpolecat, line 24: This has turned out to be helpful for debugging. We can upstream it once we're happy with it.

tpolecat:
Ok this is ready for review, at long last.

modules/service/src/main/scala/lucuma/odb/graphql/util/PrettyPrinter.scala
swalker2m: Wow, nice.

swalker2m (approved):
Excellent, I never would have figured all of that out.  :+1:

PR #102, opened 2022-09-08
Science Requirements

swalker2m:
Adds observation science requirements, making a few tweaks to the GraphQL schema along the way.

### DIFF
--- a/modules/service/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/service/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -2230,7 +2230,7 @@ input SpectroscopyScienceRequirementsInput {
   focalPlaneAngle: AngleInput
 
   # The capabilities field may be unset by assigning a null value, or ignored by skipping it altogether
-  capabilities: SpectroscopyCapabilities
+  capability: SpectroscopyCapabilities
 }
 
 # Sequence step
--- a/modules/service/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/service/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -1851,8 +1851,8 @@ input ObservationPropertiesInput {
   # The scienceRequirements defaults to spectroscopy if not specified on creation, and may be edited but not deleted
   scienceRequirements: ScienceRequirementsInput
 
-  # The scienceMode describes the chosen observing mode and instrument, is optional and may be deleted
-  scienceMode: ScienceModeInput
+  # The observingMode describes the chosen observing mode and instrument, is optional and may be deleted
+  observingMode: ObservingModeInput
 
   # Whether the observation is considered deleted (defaults to PRESENT) but may be edited
   existence: Existence
@@ -2120,7 +2120,7 @@ type Science implements StepConfig {
 }
 
 # Edit or create an observation's science mode
-input ScienceModeInput {
+input ObservingModeInput {
   # The gmosNorthLongSlit field must be either specified or skipped altogether.  It cannot be unset with a null value.
   gmosNorthLongSlit: GmosNorthLongSlitInput
 
@@ -4250,10 +4250,10 @@ type Observation {
   ): ScienceRequirements!
 
   # The science configuration
-  scienceMode(
+  observingMode(
     # Set to true to include deleted values
     includeDeleted: Boolean! = false
-  ): ScienceMode
+  ): ObservingMode
 
   # ITC execution results
   itc(
@@ -4664,7 +4664,7 @@ type Query {
   ): [Program!]! #ProgramSelectResult!
 
   # Observations grouped by commonly held science mode
-  scienceModeGroup(
+  observingModeGroup(
     # Program ID
     programId: ProgramId!
 
@@ -4676,7 +4676,7 @@ type Query {
 
     # Set to true to include deleted values
     includeDeleted: Boolean! = false
-  ): ScienceModeGroupSelectResult!
+  ): ObservingModeGroupSelectResult!
 
   # Observations grouped by commonly held science requirements
   scienceRequirementsGroup(
@@ -4783,7 +4783,7 @@ type RightAscension {
 }
 
 # Base science mode
-type ScienceMode {
+type ObservingMode {
   # Instrument
   instrument: InstrumentType!
 
@@ -4797,7 +4797,7 @@ type ScienceMode {
   gmosSouthLongSlit: GmosSouthLongSlit
 }
 
-type ScienceModeGroup {
+type ObservingModeGroup {
   # IDs of observations that use the same constraints
   observationIds: [ObservationId!]!
 
@@ -4814,13 +4814,13 @@ type ScienceModeGroup {
   ): ObservationSelectResult!
 
   # Commonly held value across the observations
-  scienceMode: ScienceMode
+  observingMode: ObservingMode
 }
 
-# The matching scienceModeGroup results, limited to a maximum of 1000 entries.
-type ScienceModeGroupSelectResult {
-  # Matching scienceModeGroups up to the return size limit of 1000
-  matches: [ScienceModeGroup!]!
+# The matching ObservingModeGroup results, limited to a maximum of 1000 entries.
+type ObservingModeGroupSelectResult {
+  # Matching ObservingModeGroups up to the return size limit of 1000
+  matches: [ObservingModeGroup!]!
 
   # `true` when there were additional matches that were not returned.
   hasMore: Boolean!
@@ -4836,7 +4836,7 @@ enum ScienceMode {
 }
 
 type ScienceRequirements {
-  # Science requirement mode
+  # Science mode
   mode: ScienceMode!
 
   # Spectroscopy requirements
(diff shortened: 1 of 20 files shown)


modules/service/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -429,13 +429,6 @@ input FluxDensityContinuumSurfaceInput {
   units: FluxDensityContinuumSurfaceUnits!
 }
 
-# Focal plane angle source angle in appropriate units
-input FocalPlaneAngleInput {
-  microarcseconds: Long
-  milliarcseconds: BigDecimal
-  arcseconds: BigDecimal
-}
-
```
swalker2m, line 438: I deleted this in favor of using generic `AngleInput` everywhere for consistency.

modules/service/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -1858,8 +1851,8 @@ input ObservationPropertiesInput {
   # The scienceRequirements defaults to spectroscopy if not specified on creation, and may be edited but not deleted
   scienceRequirements: ScienceRequirementsInput
 
-  # The scienceMode describes the chosen observing mode and instrument, is optional and may be deleted
-  scienceMode: ScienceModeInput
```
swalker2m, line 1862: `scienceMode` is also used for the choice between imaging and spectroscopy.  I wanted to remove the ambiguity so I renamed this one to "observing mode" which I think might be more appropriate anyway.

modules/service/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -2234,10 +2227,10 @@ input SpectroscopyScienceRequirementsInput {
   focalPlane: FocalPlane
 
   # The focalPlaneAngle field may be unset by assigning a null value, or ignored by skipping it altogether
-  focalPlaneAngle: FocalPlaneAngleInput
+  focalPlaneAngle: AngleInput
 
   # The capabilities field may be unset by assigning a null value, or ignored by skipping it altogether
-  capabilities: SpectroscopyCapabilities
+  capability: SpectroscopyCapabilities
```
swalker2m, line 2233: We only allow one capability so I renamed it to the singular `capability`.

modules/service/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -3413,17 +3406,6 @@ enum FocalPlane {
...
-type FocalPlaneAngle {
-  # Focal plane angle in µas
-  microarcseconds: Long!
-
-  # Focal plane angle in mas
-  milliarcseconds: BigDecimal!
-
-  # Focal plane angle in arcsec
-  arcseconds: BigDecimal!
-}
```
swalker2m, line 3425: Again, this goes away in favor of just using `Angle`.

modules/service/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -4262,16 +4244,10 @@ type Observation {
   constraintSet: ConstraintSet!
 
   # The top level science requirements
-  scienceRequirements(
-    # Set to true to include deleted values
-    includeDeleted: Boolean! = false
-  ): ScienceRequirements!
```
swalker2m, line 4268: This and the old `scienceMode` below had a spurious `includeDeleted` flag that wasn't used even in the tmp-api.

modules/service/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -4832,30 +4808,30 @@ type ScienceModeGroup {
...
-enum ScienceRequirementMode {
-  # ScienceRequirementMode Imaging
+enum ScienceMode {
+  # ScienceMode Imaging
   IMAGING
 
-  # ScienceRequirementMode Spectroscopy
+  # ScienceMode Spectroscopy
   SPECTROSCOPY
 }
```
swalker2m, line 4830: Now this enum matches what we have in `lucuma-core`.

modules/service/src/main/scala/lucuma/odb/graphql/input/AngleInput.scala
```diff
@@ -40,20 +40,6 @@ object AngleInput {
...
-    case (v, "MICROSECONDS")    => Result(getMicroseconds(v.toLong))
-    case (v, "MILLIARCSECONDS") => Result(getMilliarcseconds(v.toInt))
-    case (v, "MILLISECONDS")    => Result(getMilliseconds(v.toInt))
-    case (v, "ARCSECONDS")      => Result(getArcSeconds(v.toDouble))
-    case (v, "SECONDS")         => Result(getSeconds(v.toInt))
-    case (v, "ARCMINUTES")      => Result(getArcMinutes(v.toInt))
-    case (v, "MINUTES")         => Result(getMinutes(v.toInt))
-    case (v, "DEGREES")         => Result(getDegrees(v.toDouble))
-    case (v, "HOURS")           => Result(getHours(v.toDouble))
-  }
```
swalker2m, line 55: This was left over from the days when the API had "decimal" and "integer" inputs but was no longer being used.

modules/service/src/main/scala/lucuma/odb/graphql/table/ObservationView.scala
```diff
@@ -53,6 +53,21 @@ trait ObservationView[F[_]] { self: SkunkMapping[F] =>
           }
         }
       }
+
+      object ScienceRequirements {
+        val Mode: ColumnRef = col("c_science_mode", science_mode.embedded)
+
+        object Spectroscopy {
+          val Wavelength         = col("c_spec_wavelength",          wavelength_pm.embedded)
+          val Resolution         = col("c_spec_resolution",          pos_int.opt)
```
swalker2m, line 62: I'm not sure what `embedded` does or why it is necessary, to be honest.  @tpolecat ? I ended up switching `embedded` for `opt` when required by test case failures and it seems to depend upon whether the item is a simple field or one with field mappings for different units.

modules/service/src/test/scala/lucuma/odb/graphql/mutation/createObservation.scala
```diff
@@ -27,6 +32,19 @@ import java.time.LocalDateTime
...
+      p.toList match {
+        case Nil    => ac
+        case h :: t => ac.downField(h).downPath(t*)
+      }
+
+    def liftIO[A: Decoder]: IO[A] =
+      ac.as[A].leftMap(f => new RuntimeException(f.message)).liftTo[IO]
+
+    def downIO[A: Decoder](p: String*): IO[A] =
+      downPath(p*).liftIO[A]
```
swalker2m, line 46: Something like this might be generally useful in the test cases?

modules/service/src/test/scala/lucuma/odb/graphql/mutation/updateObservations.scala
```diff
@@ -700,10 +700,197 @@ class updateObservations extends OdbSuite
...
+
+  [ERROR] lucuma-odb-test - Error computing GraphQL response.
+  java.lang.AssertionError: assertion failed
+    at get @ skunk.util.Pool$.free$1(Pool.scala:148)
+    at get @ skunk.util.Pool$.free$1(Pool.scala:148)
+    at flatMap @ fs2.Compiler$Target.flatMap(Compiler.scala:163)
+    at flatMap @ fs2.Compiler$Target.flatMap(Compiler.scala:163)
+    at flatMap @ fs2.Pull$.fs2$Pull$$$_$interruptGuard$1(Pull.scala:938)
+    at unsafeRunSync @ munit.CatsEffectFixturesPlatform$$anon$1.beforeAll(CatsEffectFixturesPlatform.scala:41)
+    at *> @ skunk.net.SSLNegotiation$.negotiateSSL(SSLNegotiation.scala:49)
```
swalker2m, line 826: Here is the mysterious resource leak we (@tpolecat) discussed.

modules/service/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
cquiroz: Seems good

modules/service/src/main/scala/lucuma/odb/graphql/table/ObservationView.scala
tpolecat: `embedded` is a hack that's necessary when you have a column that represents a non-nullable field for a nullable embedded object. So it means "this column has to be null in the database schema but it's never null from the point of view of the GraphQL type it ends up in".

modules/service/src/test/scala/lucuma/odb/graphql/mutation/createObservation.scala
tpolecat: Yeah maybe we could move this up into `OdbSuite` or something.

tpolecat (approved):
:+1: A few comments but nothing important. Looks great!

PR #105, opened 2022-09-09
schema updates

swalker2m:
A WIP update for discussing the changes to observing mode.


modules/service/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -894,18 +894,19 @@ input GmosNorthGratingConfigInput {
 }
 
 # Edit or create GMOS North Long Slit advanced configuration
-input GmosNorthLongSlitAdvancedConfigInput {
-  # The overrideWavelength field may be unset by assigning a null value, or ignored by skipping it altogether
-  overrideWavelength: WavelengthInput
+input GmosNorthLongSlitInput {
```
swalker2m, line 897: Collapses advanced and basic into one input for GMOS North long slit and one for GMOS South long slit.  The input allows you to

* change the grating, filter and/or FPU
* override the central wavelength that would otherwise be used by default
* override the exposure time mode from the science requirements
* optionally set explicit values for x-binning, etc. that are otherwise defaulted

modules/service/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -2921,13 +2883,15 @@ enum CloudExtinction {
   THREE_POINT_ZERO
 }
 
-# ConfigurationMode
-enum ConfigurationModeType {
-  # ConfigurationModeType GmosNorthLongSlit
+# ObservingMode
+enum ObservingModeType {
```
swalker2m, line 2887: The old "science mode" and "configuration mode type" would become "observing mode" and "observing mode type".

modules/service/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -3607,64 +3571,143 @@ enum GmosNorthGrating {
...
-# GMOS North Long Slit mode
-type GmosNorthLongSlit {
-  # GMOS North Long Slit basic configuration
-  basic: GmosNorthLongSlitBasic!
-
-  # GMOS North Long Slit advanced configuration
-  advanced: GmosNorthLongSlitAdvanced
-}
+# GMOS Long Slit Observing Mode common fields
+interface GmosLongSlitObservingMode {
```
swalker2m, line 3575: An interface with the common GMOS long slit observing mode values. For any property, such as `xBin`, there are three fields:

1. the property value that will be used (e.g., `xBin`)
2. the default value of that property if not explicitly set (e.g., `defaultXBin`)
3. the (optional) explicit value chosen for this property by the user (e.g., `explicitXBin`)

Only the `explicit___` value appears in the input.  The others are calculated or set from constants.

modules/service/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -3607,64 +3571,143 @@ enum GmosNorthGrating {
...
+
   # GMOS North Grating
   grating: GmosNorthGrating!
 
   # GMOS North Filter
   filter: GmosNorthFilter
 
   # GMOS North FPU
-  fpu: GmosNorthBuiltinFpu!
+  fpu: GmosNorthBuiltinFpu
```
swalker2m, line 3678: `grating`, `filter` and `fpu` are set when the observing mode is selected from the table but can be subsequently updated.

modules/service/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
swalker2m: If you only care about getting the value that will be used, the main property value should be consulted.  For x-binning it would be just `xBin`.  If you want to know what the default value we use will be (which for x-binning depends on the grating, target and image quality) use `defaultXBin`.  A UI could compare the `xBin` to `defaultXBin` to determine whether the property has been changed from its nominal value.  It could offer a way to abandon customizations and go back to the defaults.

modules/service/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
toddburnside: Could we just update the exposure time mode in the science requirements directly? And not have an override? Might be something for Andy to answer.

swalker2m: That sounds good to me.  I'm going to assume we can do that for now.  I think it implies that in the requirements section we'll have to allow them the option to pick fixed exposure time/count.

toddburnside: Right. In best case we remove it from the old advanced configuration tab and only have it in the requirements.

modules/service/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -3607,64 +3565,168 @@ enum GmosNorthGrating {
...
+  defaultWavelengthDithersNm: [BigDecimal!]!
+  explicitWavelengthDithersNm: [BigDecimal!]
+
+  spatialOffsets: [q!]!
+  defaultSpatialOffsets: [q!]!
+  explicitSpatialOffsets: [q!]
+
+  # The grating as it was initially selected.  See the `grating` field for the
+  # grating that will be used in the observation.
+  initialGrating: GmosNorthGrating!
```
swalker2m, line 3721: Andy expressed a desire to go back to the table of options in the UI and have the value you originally selected (before any edits via the `GmosNorthLongslitInput`) highlighted.  Perhaps it is possible to do this purely in the UI if it is limited to a single session?  Otherwise we need to keep up with the initial grating, filter and FPU.  Note that these are read-only so there's never any confusion about how to change the grating, filter, or FPU.

PR #123, opened 2022-09-30
predicates refactor

tpolecat:
This ended up leading to some changes to the way `Path` works in Grackle, which turned this into a larger refactor than I anticipated. 

The gist of the Grackle change is that paths now have a starting GraphQL type. This allows the path to consult the schema and figure out whether it's a simple value or a list of values (the old `ListPath`/`UniquePath` distinction), which ended up being really hard to keep track of when trying to build up predicates in a composable way. So now you would construct the path from a program to its observations' ids like this

```scala
ProgramType / "observations" / "id"
```

which ends up clarifying things a lot. But it made `Path` construction reliant on a `Schema`, which required some refactoring that ended up causing circularity issues in our messy cake. So I straightened that out.

image

The new `Predicates` object (inherited by all mappings) provides entry points for constructing predicates, starting at a root type (the result type for the query, typically). For example:

```scala
Predicates.target.id.gtEql(x) // GtEql(TargetType / "target" / "id", Const(x))
```

The logic in the `lucuma.odb.graphql.predicate` package is composable and very simple. I have added some higher-level constructors for things like `isVisibleTo` (on `Program`) and `includeDeleted` (on `Existence`). You can see the effect of these changes in mappings that define query elaborators (`QueryMapping`, `MutationMapping`, etc.).


swalker2m (approved):
Thanks for doing all of this.  A tremendous improvement, making predicates much easier to reuse. :+1:

PR #133, opened 2022-10-13
Update to Grackle 0.8.x, add SelectResult mappings

tpolecat:
- [x] Query.programs
- [x] Query.targets
- [x] Query.observations
- [x] Mutation.updateAsterisms
- [x] Mutation.updateObservations
- [x] Mutation.updatePrograms
- [x] Program.observations


swalker2m (commented):
This is looking great.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/MutationMapping.scala
```diff
@@ -81,12 +81,22 @@ trait MutationMapping[F[_]: MonadCancelThrow] extends Predicates[F] {
   // Convenience for constructing a SqlRoot and corresponding 1-arg elaborator.
   private trait MutationField {
     def Elaborator: PartialFunction[Select, Result[Query]]
-    def FieldMapping: SqlRoot
+    def FieldMapping: RootEffect
```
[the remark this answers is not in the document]
tpolecat: Prior to 0.8 you would define a mutation on a root mapping in order to perform some effectful computation and then produce a follow-up query (insert a value and then query it back, for instance). This was also used for subscriptions, which just compute a stream of queries. But in 0.8 all the root mappings go away (they're just normal fields on `type Query` or whatever) so this is how you do effectful computations now.

tpolecat:
Ok this is ready for review. It's not as generalized as it probably could be, but I think I have gone as far enough as makes sense for now since we haven't used this facility much yet.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/MutationMapping.scala
```diff
@@ -81,18 +84,58 @@ trait MutationMapping[F[_]: MonadCancelThrow] extends Predicates[F] {
...
   }
   private object MutationField {
     def apply[I: ClassTag: TypeName](fieldName: String, inputBinding: Matcher[I])(f: (I, Query) => F[Result[Query]]) =
       new MutationField {
-        val FieldMapping = SqlRoot(fieldName, mutation = Mutation.simple((child, env) => env.getR[I]("input").flatTraverse(f(_, child))))
+        val FieldMapping =
+          RootEffect.computeQuery(fieldName) { (query, tpe, env) =>
```
tpolecat, line 93: `RootEffect` replaces all existing roots with mutations.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/MutationMapping.scala
```diff
@@ -208,12 +247,12 @@ trait MutationMapping[F[_]: MonadCancelThrow] extends Predicates[F] {
       val selectObservations: F[Result[(List[Observation.Id], Query)]] =
         idSelect.traverse { which =>
           observationService.use { svc =>
-            svc
-              .selectObservations(which)
-              .fproduct {
-                case Nil => Limit(0, child)
-                case ids => Filter(Predicates.observation.id.in(ids), child)
```
tpolecat, line 215: This dance isn't necessary anymore. `In(Nil)` now correctly evaluates to `false`.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/MutationMapping.scala
```diff
@@ -81,18 +84,58 @@ trait MutationMapping[F[_]: MonadCancelThrow] extends Predicates[F] {
...
+            }
+          }
         val Elaborator =
           case Select(`fieldName`, List(inputBinding("input", rInput)), child) =>
             rInput.map(input => Environment(Env("input" -> input), Select(fieldName, Nil, child)))
       }
   }
 
+  def mutationResultSubquery[A: Order](predicate: Predicate, order: OrderSelection[A], limit: Option[NonNegInt], collectionField: String, child: Query): Result[Query] =
+    val limitʹ = limit.foldLeft(1000)(_ min _.value)
```
swalker2m, line 110: I guess we should stick the limit constant somewhere at some point.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/MutationMapping.scala
swalker2m: Cool.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/MutationMapping.scala
```diff
@@ -81,18 +84,58 @@ trait MutationMapping[F[_]: MonadCancelThrow] extends Predicates[F] {
...
+  def observationResultSubquery(oids: List[Observation.Id], limit: Option[NonNegInt], child: Query) =
+    mutationResultSubquery(
+      predicate = Predicates.observation.id.in(oids),
+      order = OrderSelection[Observation.Id](ObservationType / "id"),
+      limit = limit,
+      collectionField = "observations",
+      child          
+    )
+
+  def programResultSubquery(oids: List[Program.Id], limit: Option[NonNegInt], child: Query) =
```
swalker2m, line 130: `pids` I suppose

modules/service/src/main/scala/lucuma/odb/graphql/mapping/ResultMapping.scala
```diff
@@ -0,0 +1,109 @@
...
+      size  <- items.listSize
+    yield limit < 0 || size > limit
+
+
+  /** A cursor transformation that takes `n` elements from a list (if it's a list). */
+  object Take {
+    def apply(n: Int)(c: Cursor): Result[Cursor] =
+      (c.listSize, c.asList(Seq)).mapN { (size, elems) =>
+        if size <= n then c
+        else ListTransformCursor(c, size - 1, elems.init)
```
swalker2m, line 39: This is where the extra item for `hasMore` is taken out?

modules/service/src/main/scala/lucuma/odb/graphql/util/MappingExtras.scala
```diff
@@ -19,5 +23,41 @@ trait MappingExtras[F[_]] extends Mapping[F] {
         CursorField(field, _.field(underlyingField, None).flatMap(_.as[A].map(f)), List(underlyingField))
     }
   }
+      
+  /**
+    * A dispatching `TypeMapping` that selects a different underlying mapping based on the GraphQL
+    * `Type` and field name.
+    */
```
swalker2m, line 30: This will be used for angles and wavelengths etc. that appear in multiple places?

swalker2m (approved):
Looks great to me!  :+1:

modules/service/src/main/scala/lucuma/odb/graphql/mapping/MutationMapping.scala
tpolecat: :+1:

modules/service/src/main/scala/lucuma/odb/graphql/mapping/ResultMapping.scala
tpolecat: Correct.

modules/service/src/main/scala/lucuma/odb/graphql/util/MappingExtras.scala
tpolecat: Yes, I think it's easier to deal with than `PrefixedMapping`. So we can go back and change those.

PR #135, opened 2022-10-18
GMOS North Long Slit Observing Mode Creation

swalker2m:
This is a WIP on adding observing mode.  At this point it supports creating (most of) a GMOS North Long Slit observing mode upon observation creation.  Missing:

* correct `defaultXBin` calculation
* GMOS South Long Slit observing mode
* editing


tpolecat:
Wow, this is a lot. I feel shame. Looks good!

modules/service/src/main/scala/lucuma/odb/graphql/mapping/GmosNorthLongSlitMapping.scala
```diff
@@ -0,0 +1,91 @@
...
+        SqlField("grating", GmosNorthLongSlitTable.Grating),
+        SqlField("filter",  GmosNorthLongSlitTable.Filter),
+        SqlField("fpu",     GmosNorthLongSlitTable.Fpu),
+        SqlObject("centralWavelength"),
+    
+        // xBin
+//        FieldRef[Option[GmosYBinning]]("explicitXBin").as("yBin", _.getOrElse(defaultXBin)),
+
+        // defaultXBin
+        // TODO: a function of FPU, target SourceProfile, and ImageQuality
```
swalker2m, line 59: The default x-binning calculation involves looking up the FPU (which I think I can get with a simple `FieldRef`) along with the target's `SourceProfile` and `ImageQuality`, which are used to get the effective slit width.  I don't know how to handle the latter two.

tpolecat: I talked to Miles and I have an idea of how to do this. If you want to punt for now and leave this field unmapped (or mapped to a constant) I can come back and fill it in.

swalker2m: That sounds great, thanks.  I'll map it to a constant for now and add the offsets.  Then I should probably stop and merge this checkpoint before it gets any bigger!

modules/service/src/main/scala/lucuma/odb/graphql/mapping/GmosNorthLongSlitMapping.scala
```diff
@@ -0,0 +1,105 @@
...
+
+        // TODO: a function of FPU, target SourceProfile, and ImageQuality
+        FieldRef[GmosNorthFpu]("fpu")
+          .as(
+            "defaultXBin",
+            fpu => GmosLongSlitMath.xbinNorth(
+              fpu,
+              /*placeholder*/ SourceProfile.Point(SpectralDefinition.BandNormalized(UnnormalizedSED.StellarLibrary(StellarLibrarySpectrum.O5V), SortedMap.empty)),
+              /*placeholder*/ ImageQuality.TwoPointZero,  // so the FPU size is the effective slit width for now
+              PosDouble.unsafeFrom(2.0)
```
swalker2m, line 71: Now updated with placeholder `SourceProfile` and `ImageQuality` values.  It's a shame that we need to pass the entire `SourceProfile`.  The `SpectralDefinition` is irrelevant for the purpose of calculating the x-binning.

swalker2m (commented):
I think this is a pretty good chunk of the problem so I'll mark it ready for review

PR #142, opened 2022-10-27
GMOS South Long Slit

swalker2m:
Generalizes a bit the GMOS long slit code and adds GMOS South.


modules/service/src/main/scala/lucuma/odb/graphql/input/util.scala
```diff
@@ -0,0 +1,13 @@
...
+package lucuma.odb.graphql.input
+
+import cats.syntax.foldable.*
+import edu.gemini.grackle.Result
+
+def oneOrFail[A](options: (Option[A], String)*): Result[A] =
+  options.toList.flatMap(_._1.toList) match {
+    case List(a) => Result(a)
+    case _       => Result.failure(s"Expected exactly one of ${options.map(_._2).intercalate(", ")}")
+  }
```
swalker2m, line 13: We used this logic in several places so I pulled it up to the package level.

tpolecat: :+1:

tpolecat (approved):
This looks great!

PR #144, opened 2022-10-28
Switch to SwitchMapping

swalker2m:
Changes the uses of `PrefixedMapping` to `SwitchMapping`.  (Relies on changes to `goos-south-long-slit` branch/PR.) Did I do this correctly?


tpolecat (approved):
:+1: perfect! For the cases where it works (i.e., you don't need a long path to disambiguate mappings, as is always the case so far) I think the behavior of `SwitchMapping` is easier to understand than `PrefixedMapping`.

PR #145, opened 2022-10-30
Observation Events

swalker2m:
More or less a copy of what @tpolecat did for program events.  There is an opportunity for a bit more sharing across all the event types I think.  I added tests but I'm worried about how slow they run so I stopped at a handful.


tpolecat (approved):
Looks great! :+1:

PR #158, opened 2022-11-16
WIP: streaming event bug

tpolecat:
This adds a testcase to demonstrate a Grackle bug. It's a degenerate case so I'm not going to try to chase it down right now.


tpolecat:
This is actually fixed (with a test) in main, so closing.

PR #172, opened 2022-12-05
SourceProfileCodec

swalker2m:
Converts `SourceProfileEncoder` into `SourceProfileCodec` (and adds the encoder for emission lines that was missing).  We store the `SourceProfile` as a JSON blob in the database, so the encoder is used for that.  We will need to pass the `SourceProfile` to the ITC though, so I added a decoder.

This prompts a few questions:

* The codec has to exactly match the `OdbSchema.graphql` or else queries will fail.  Even base types like `Angle` and `Wavelength` have to be encoded in all of their possible units, not just the base microarcseconds or picometers.  With that in mind, is `lucuma-odb` the right place for this or should it move to core or schemas?  I'm hesitant to do that since a change to the JSON codec elsewhere would break the ODB.
* The ITC is called via graphql as well and the `SourceProfile` is just passed as part of the input.  In theory we could lift the JSON blob and pass it directly to the ITC without first decoding and then re-encoding it.   That obviously means that again the JSON codec understood by the two (ODB and ITC) would have to match exactly and could fail at runtime if only one or the other is updated.
* Perhaps all of this is pointing to fully mapping `SourceProfile` to tables and doing a service to read/write it?


cquiroz:
Don't we have some decoders already in lucuma-schemas?
https://github.com/gemini-hlsw/lucuma-schemas/tree/main/lucuma-schemas/src/main/scala/lucuma/schemas/decoders

tpolecat (approved):

cquiroz:
we may want to consider the time based magnitudes at this stage rather than having to convert them later

tpolecat:
I think I might prefer to punt on time-varying magnitudes and use it as an exercise for schema evolution once we have the basic stuff working.

PR #182, opened 2022-12-14
WIP: Query/targetGroup

tpolecat:
This upturned a Grackle issue that I wasted a lot of time trying to track down, but was finally able to work around it by adding a `program` field to `TargetGroup`. Still need to add a test.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/graphql/OdbMapping.scala
+++ b/modules/service/src/main/scala/lucuma/odb/graphql/OdbMapping.scala
@@ -127,6 +127,8 @@ object OdbMapping {
           with SubscriptionMapping[F]
           with TargetEnvironmentMapping[F]
           with TargetMapping[F]
+          with TargetGroupMapping[F]
+          with TargetGroupSelectResultMapping[F]
           with TargetSelectResultMapping[F]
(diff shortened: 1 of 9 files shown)


modules/service/src/main/scala/lucuma/odb/graphql/OdbMapping.scala
```diff
@@ -226,13 +230,15 @@ object OdbMapping {
...
                 TargetEnvironmentElaborator,
+                TargetGroupElaborator,
                 QueryElaborator,
               ).combineAll
             )
 
           // Override `defaultRootCursor` to log the GraphQL query. This is optional.
           override def defaultRootCursor(query: Query, tpe: Type, env: Env): F[Result[(Query, Cursor)]] =
-            Logger[F].info("\n\n" + PrettyPrinter.query(query).render(100) + "\n") *> super.defaultRootCursor(query, tpe, env)
+            Logger[F].info("\n\n" + PrettyPrinter.query(query).render(100) + "\n") >>
```
tpolecat, line 240: I was having a worrying issue with evaluation order with `*>` here so I switched it to `>>` and will investigate further. Unrelated to anything else happening here.

swalker2m (approved):
👍

PR #183, opened 2022-12-15
ITC

swalker2m:
This PR is a bit of a monster, but about 1/3 is test code.  It implements the `itc` query for an observation.  I had to move it to the root in order to make it work with Grackle.   One guide to reviewing this work would be to look first at the schema updates.  Then there is code to produce corresponding data that can be thus encoded.  In particular:

* a new database view, `v_itc_params`, collecting ITC parameter data from disparate tables,
* an `ItcInputService` which just queries the database to collect the parameters needed to call the ITC,
* an `Itc` client which calls the aforementioned input service and then passes the parameters along to the actual `lucuma.itc.client.ItcClient`, collecting the data into a more easily digestible format, and
* an `ItcResultEncoder` to create schema-compatible JSON for the results.

Unfortunately the view cannot also include the observing mode parameters since there will eventually be scores of these each in their own instrument specific tables.

N.B. the `itc` query itself in the schema may go away, but all the code to implement it will be needed when the sequence is generated.

### DIFF
--- a/modules/service/src/main/resources/db/migration/V0110_itc_params.sql
+++ b/modules/service/src/main/resources/db/migration/V0110_itc_params.sql
@@ -1,0 +1,30 @@
+CREATE VIEW v_itc_params AS
+SELECT
+  o.c_observation_id,
+  o.c_image_quality,
+  o.c_cloud_extinction,
+  o.c_sky_background,
+  o.c_water_vapor,
+  o.c_air_mass_min,
+  o.c_air_mass_max,
+  o.c_hour_angle_min,
+  o.c_hour_angle_max,
+  o.c_spec_signal_to_noise,
+  o.c_spec_signal_to_noise_at,
+  o.c_observing_mode_type,
+  t.c_target_id,
+  t.c_sid_rv,
+  t.c_source_profile
+FROM
+  t_observation o
+INNER JOIN t_asterism_target a
+  ON o.c_observation_id = a.c_observation_id
+INNER JOIN t_target t
+  ON a.c_target_id      = t.c_target_id
+WHERE
+      o.c_spec_signal_to_noise IS NOT NULL
+  AND o.c_observing_mode_type  IS NOT NULL
+  AND t.c_sid_rv               IS NOT NULL
+ORDER BY
+  o.c_observation_id,
+  t.c_target_id
--- a/modules/service/src/main/resources/db/migration/V0110_itc_params.sql
+++ b/modules/service/src/main/resources/db/migration/V0110_itc_params.sql
@@ -17,14 +17,10 @@ SELECT
   t.c_source_profile
 FROM
   t_observation o
-INNER JOIN t_asterism_target a
+LEFT JOIN t_asterism_target a
   ON o.c_observation_id = a.c_observation_id
-INNER JOIN t_target t
+LEFT JOIN t_target t
   ON a.c_target_id      = t.c_target_id
-WHERE
-      o.c_spec_signal_to_noise IS NOT NULL
-  AND o.c_observing_mode_type  IS NOT NULL
-  AND t.c_sid_rv               IS NOT NULL
 ORDER BY
   o.c_observation_id,
   t.c_target_id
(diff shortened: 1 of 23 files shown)


modules/service/src/main/resources/db/migration/V0110_itc_params.sql
```diff
@@ -0,0 +1,30 @@
...
+FROM
+  t_observation o
+INNER JOIN t_asterism_target a
+  ON o.c_observation_id = a.c_observation_id
+INNER JOIN t_target t
+  ON a.c_target_id      = t.c_target_id
+WHERE
+      o.c_spec_signal_to_noise IS NOT NULL
+  AND o.c_observing_mode_type  IS NOT NULL
+  AND t.c_sid_rv               IS NOT NULL
```
swalker2m, line 27: These are all required to call the ITC, but optional in the schema.  I'll need to stop filtering on these in the view and instead handle it in code if I want the output to be more informative.

modules/service/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -4687,6 +4681,16 @@ type Query {
...
 
+  # ITC execution results
+  itc(
+    programId:     ProgramId!
+
+    observationId: ObservationId!
+
+    # Whether to use cached results (true) or ignore the cache and make a remote ITC call (false).
+    useCache: Boolean! = true
+  ): ItcSuccess
```
swalker2m, line 4692: This had to move to the root since executing an effect at a lower level is apparently problematic.

modules/service/src/main/scala/lucuma/odb/graphql/OdbMapping.scala
```diff
@@ -152,14 +163,54 @@ object OdbMapping {
...
+
+            def toJson(res: List[ItcResult]): F[Json] =
+              res.partitionMap {
+                case ItcResult.Error(msg)           => msg.asLeft[ItcResult.Success]
+                case s @ ItcResult.Success(_, _, _) => s.asRight[String]
+              }.bimap(_.headOption, _.maxByOption(_.exposureTime.value)) match {
+                case (_, Some(success)) => success.asJson.pure[F]
+                case (Some(error), _)   => new RuntimeException(s"Could not obtain ITC results: $error").raiseError[F, Json]
+                case _                  => Json.Null.pure[F]
+              }
```
swalker2m, line 198: Here we're taking a decision about how to handle multiple ITC results (picking the longest exposure time successful one) for an observation when there is more than one target in its asterism.  I think it might be good to offer results per target instead.

modules/service/src/main/scala/lucuma/odb/Config.scala
```diff
@@ -18,38 +19,44 @@ import natchez.http4s.NatchezMiddleware
...
-  hcDataset:    String,          // Honeycomb dataset
-  database:     Config.Database, // Database config
-  domain:       String,          // Domain, for CORS headers
+  port:       Port,             // Our port, nothing fancy.
+  itcRoot:    Uri,              // ITC service URI
+  sso:        Config.Sso,       // SSO config
+  serviceJwt: String,           // Only service users can exchange API keys, so we need a service user JWT.
+  honeycomb:  Config.Honeycomb, // Honeycomb config
+  database:   Config.Database,  // Database config
+  domain:     String,           // Domain, for CORS headers
```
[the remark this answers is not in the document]
tpolecat: 👍🏻

modules/service/src/main/scala/lucuma/odb/graphql/client/Itc.scala
```diff
@@ -0,0 +1,257 @@
...
+    final case class Missing(
+      targetId: Option[Target.Id],
+      params:   NonEmptySet[Param]
+    ) extends Result
+
+    def missing(
+      targetId: Option[Target.Id],
+      params:   NonEmptySet[Param]
+    ): Result =
+      Missing(targetId, params)
```
tpolecat, line 108: If you use Scala 3 `enum` you don't need these widened constructors (but there may be a reason this data type won't work with `enum`).

tpolecat (approved):
Wow this looks great.  👍🏻 👍🏻 

Once we get it deployed as a service we'll need to swap out the in-memory `ItcCache` with one backed by Redis.

modules/service/src/main/scala/lucuma/odb/graphql/client/Itc.scala
swalker2m: Thanks, that hadn't occurred to me but it makes it better.

PR #188, opened 2022-12-22
Query/asterismGroup

tpolecat:
This adds asterism groups via a mechanism that makes me feel shame but it's the best I could come up with.

- I added a `c_asterism_group` column on `t_observation` that contains a sorted list (as `jsonb`) of the targets associated with the observation, thus giving us a key we can use for efficient grouping. The field is automatically recomputed when a relevant row is added, updated, or deleted in `t_asterism_target`.

- There is also now a `v_asterism_group` view that contains unique program+asterism pairs, plus the key of one observation that uses the asterism, through which we can join to get the asterism's targets. 

The rest of the mapping follows trivially, as they say.

### DIFF
--- a/modules/service/src/main/resources/db/migration/V0030__observations.sql
+++ b/modules/service/src/main/resources/db/migration/V0030__observations.sql
@@ -225,7 +225,8 @@ create table t_observation (
   -- observing mode
   c_observing_mode_type e_observing_mode_type null default null,
 
-  --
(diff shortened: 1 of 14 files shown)


swalker2m (commented):
Looks great!  :+1:  The `c_asterism_group` column idea seems cool.

Is this not ready for final review?

tpolecat:
Yep, it's ready. I clicked wrong.

PR #189, opened 2022-12-23
mode views

swalker2m:
This PR is intended to be another step on the way to sequence generation.  In particular, it adds a view per observing mode that combines the raw view parameters in the corresponding table with other fields required for sequence generation.  For example, for GMOS North Long Slit, there is the `t_gmos_north_long_slit` table and now the `v_gmos_north_long_slit` view which selects all the table's fields and adds the observation image quality and the target source profile (if any).

Finally, since the x-binning calculation requires exactly these two extra fields I decided I could go back to the `GmosLongSlitMapping` and correct the `defaultXBin` field.


modules/service/src/main/scala/lucuma/odb/graphql/mapping/GmosLongSlitMapping.scala
```diff
@@ -144,17 +153,20 @@ trait GmosLongSlitMapping[F[_]]
...
+            for {
+              fpu <- cursor.fieldAs[GmosNorthFpu]("fpu")
+              iq  <- cursor.fieldAs[ImageQuality]("imageQuality")
+              j   <- cursor.fieldAs[Option[Json]]("sourceProfile")
+              sp  <- j.traverse(json => Result.fromEither(json.as[SourceProfile].leftMap(_.message)))
+            } yield
+              sp.fold(GmosXBinning.Two) { sourceProfile =>  // TODO: What should the real default be if there is no target
+                GmosLongSlitMath.xbinNorth(fpu, sourceProfile, iq, PosDouble.unsafeFrom(2.0))
+              },
+          List("fpu", "imageQuality", "sourceProfile")
```
swalker2m, line 168: I wanted to get this out before the Christmas break, but now I'll go back and see if I can generalize this across north/south.

tpolecat (approved):
This all looks great. :+1:

PR #202, opened 2023-01-10
GMOS Long Slit Science Sequence Generation Preliminary

swalker2m:
I'm soliciting early feedback on sequence generation in the form of this PR.  It adds GMOS Long Slit science sequence generation, without regard to executed datasets and without actually hooking it up to the AP yetI.  In particular it relies upon a lot of model code, and even introduces some of its own, so I'm nervous about adding it to the database.  We haven't been heavily dependent on `lucuma-core` before now.


tpolecat (approved):
This looks very good 👍🏻. As we discussed on Slack, I think it's ok for this to depend on the model a bit more than the SQL/GraphQL code does because it's not constraining the schemas (which is what we want to avoid). It might make senes to pull this out into an independent module that `service` depends on, just to enforce the separation of concerns.

swalker2m:
Created a separate module (called `sequence`) as discussed.

PR #204, opened 2023-01-10
generalize switchmapping

tpolecat:
This generalize `SwitchMapping` to take a list of `(Path, ObjectMapping)` pairs (I have run into a case where type + fieldname isn't good enough).


swalker2m:
This seems like a good improvement.  :+1:

PR #205, opened 2023-01-10
Observation/plannedTime

tpolecat:
This adds the `plannedTime` field to `type Observation`.

### DIFF
--- a/modules/service/src/main/resources/db/migration/V0030__observations.sql
+++ b/modules/service/src/main/resources/db/migration/V0030__observations.sql
@@ -150,6 +150,11 @@ create table t_observation (
   c_active_status      e_obs_active_status not null    default 'active',
   c_visualization_time timestamp           null        default null,
 
+  -- planned time summary
+  c_pts_pi        interval not null default '0h',
+  c_pts_uncharged interval not null default '0h',
+  c_pts_execution interval not null default '0h',
+  
   -- position angle constraint
   c_pac_mode           e_pac_mode          not null    default 'unbounded',
   c_pac_angle          d_angle_µas         not null    default 0,
--- a/modules/service/src/main/scala/lucuma/odb/graphql/mapping/NonNegDurationMapping.scala
+++ b/modules/service/src/main/scala/lucuma/odb/graphql/mapping/NonNegDurationMapping.scala
@@ -9,24 +9,28 @@ import edu.gemini.grackle.skunk.SkunkMapping
 import io.circe.Encoder
 import lucuma.core.model.Program
 import lucuma.odb.graphql.table.AllocationTable
+import lucuma.odb.graphql.table.ObservationView
 import lucuma.odb.graphql.table.ProgramTable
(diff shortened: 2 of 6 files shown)


modules/service/src/main/scala/lucuma/odb/graphql/mapping/NonNegDurationMapping.scala
```diff
@@ -9,24 +9,28 @@ import edu.gemini.grackle.skunk.SkunkMapping
...
-        LargeProgramType / "totalTime"       -> nonNegDurationMapping(ProposalTable.TotalTime)(ProposalTable.ProgramId),
-        AllocationType / "duration"          -> nonNegDurationMapping(AllocationTable.Duration)(AllocationTable.ProgramId, AllocationTable.Partner),
+        ProgramType / "plannedTime" / "pi"            -> nonNegDurationMapping(ProgramTable.PlannedTime.Pi)(ProgramTable.Id),
+        ProgramType / "plannedTime" / "uncharged"     -> nonNegDurationMapping(ProgramTable.PlannedTime.Uncharged)(ProgramTable.Id),
+        ProgramType / "plannedTime" / "execution"     -> nonNegDurationMapping(ProgramTable.PlannedTime.Execution)(ProgramTable.Id),
+        ObservationType / "plannedTime" / "pi"        -> nonNegDurationMapping(ObservationView.PlannedTime.Pi)(ObservationView.Id),
+        ObservationType / "plannedTime" / "uncharged" -> nonNegDurationMapping(ObservationView.PlannedTime.Uncharged)(ObservationView.Id),
+        ObservationType / "plannedTime" / "execution" -> nonNegDurationMapping(ObservationView.PlannedTime.Execution)(ObservationView.Id),
```
tpolecat, line 30: This is where I needed the more general `SwitchMapping` from #204

modules/service/src/main/resources/db/migration/V0030__observations.sql
```diff
@@ -150,6 +150,11 @@ create table t_observation (
   c_active_status      e_obs_active_status not null    default 'active',
   c_visualization_time timestamp           null        default null,
 
+  -- planned time summary
+  c_pts_pi        interval not null default '0h',
+  c_pts_uncharged interval not null default '0h',
+  c_pts_execution interval not null default '0h',
```
swalker2m, line 156: This seems fine for now.  How might it eventually be implemented?  It'll have to be a fairly complicated calculation that we won't be able to directly execute in a stored procedure.  Many different kinds of updates to the observation would have to also update these values.

swalker2m (approved):
Looks good.  :+1:

modules/service/src/main/resources/db/migration/V0030__observations.sql
tpolecat: Unclear. We may need to recompute it when we do various updates, or just do it on-demand as a top-level computation.

PR #207, opened 2023-01-11
schemas module?

swalker2m:
This PR reshuffles a few things and adds a `schema` module.  The issue is that sequence generation will need to create a JSON representation of the sequence, which will not only be returned from the queries but also used to generate step and atom ids.  So we need a place to put JSON codecs where the sequence generator and the service both can use them.  The JSON codecs have to be tied to the graphql schema, of course, so it seemed to make sense to put them all together into a new `schema` module.

Does this make sense?  Is there a better option?  If we keep this, I'll need to add the encoders (at least) for GMOS steps in `schema`.


cquiroz:
Maybe some of these should go into lucuma-schemas?

swalker2m:
> Maybe some of these should go into lucuma-schemas?

Perhaps you're right.  @tpolecat what do you think?  Is `lucuma-schemas` just for the ODB?  I see that's the only one there, so it seems like it could be a module in `lucuma-odb`?  Each service with a GraphQL API could have a `schema` module I suppose.

I don't have a good sense for `lucuma-schemas`.  It has some `model` stuff in there like `Visit` and `ExecutionEvent` which seem like they might rather be in `lucuma-core`.  The ODB schema is there in something called `templates`.  Why "templates"?

tpolecat (approved):
I think this looks good. I would like to keep the ODB as self-contained as possible and not have the schema living elsewhere. I see `lucuma-schemas` as a helper for client code that has a snapshot of the old schema at some point in time, but it's not the source of truth.

cquiroz:
Maybe odb should take over Lucuma schemas and be the source of truth. It maybe problematic to have these two places where schemas and encoders are made available

swalker2m:
> Maybe odb should take over Lucuma schemas and be the source of truth. It maybe problematic to have these two places where schemas and encoders are made available

Is it just GraphQL schemas and JSON codecs?  If so, that makes sense to me.  Or does it contain other client stuff derived from the schema?

One thing to keep in mind though is that the encoder has to write, for example, all possible unit options for types like `Angle` and `Wavelength` in order for Grackle to fulfill a query in any of those units.  So it produces microarcseconds, milliarcseconds, arcseconds, etc. That would not be optimal, say, as an input that a client sends to the database and may in fact be rejected altogether.

rpiaggio:
I'm late to the party, but thought I'd aswer some of the questions raised here and add a couple of comments regarding `lucuma-schemas`.

`lucuma-schemas` is meant to provide client code for our GraphQL endpoints, including, but not limited to, the ODB.

> Is it just GraphQL schemas and JSON codecs?

Yes, that's what it's meant to be, with caveats:

The schema itself is irrelevant for `clue`, instead relying on (mostly generated) code for the schema:
- `case class`es for `input` types.
- Encoders for such classes (also `Eq` instances).

> Why "templates"?

`templates` contains the template for `clue` to generate the schema code for clients. A template basically maps enum and model types to existing classes. It also provides the necessary encoders and decoders for clients (unless already generated). These are not necessarily the same as required for the server: A good example is the `Angle` codec that @swalker2m mentioned. Also, clients don't need `input` type encoders.

> It has some model stuff in there like Visit and ExecutionEvent which seem like they might rather be in lucuma-core.

I agree, and in fact there's a closed PR for that: https://github.com/gemini-hlsw/lucuma-core/pull/566. 

IIRC, I think someone objected to having client-only model classes in `lucuma-core` and I ended up putting those in `lucuma-schemas` instead for lack of a better place.

----

All that said, I like the approach of having ODB (and each GQL server project) publishing their own `schema` module. This would result in a schema upgrade publishing the corresponding schema project for clients, which is nice. Necessary changes for clients must be implemented before publishing the upgraded schema, lest test fail.

cquiroz:
I'm ok with each service having its own schema but I think we want to solve the problem of common schema

For example many services would need the concept of Target and I think we would like to have each service use the same target definition 

As a minimum it would be good for api user to have a uniform model for these "entities"

I'm looking for a way to compose schemas such that I can "import" the odb schema (or the common bits) and reuse them in the services (iTC,n-gage, etc ...)

PR #209, opened 2023-01-11
add some synthetic ids

tpolecat:
Adding synthetic ids to some nullable embedded objects that were mapped as if they were non-null (selecting fields when it's null provokes an AssertionError).

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/graphql/table/ObservationView.scala
+++ b/modules/service/src/main/scala/lucuma/odb/graphql/table/ObservationView.scala
@@ -78,9 +78,13 @@ trait ObservationView[F[_]] extends BaseMapping[F] {
             val Value: ColumnRef       = col("c_spec_signal_to_noise_at",    wavelength_pm.embedded)
           }
 
+          object WavelengthCoverage {
+            val SyntheticId: ColumnRef = col("c_spec_wavelength_coverage_id", observation_id.embedded)
+            val Value: ColumnRef       = col("c_spec_wavelength_coverage",    wavelength_pm.embedded)
+          }
+
           val Resolution: ColumnRef         = col("c_spec_resolution",          pos_int.opt)          
           val SignalToNoise: ColumnRef      = col("c_spec_signal_to_noise",     signal_to_noise.opt)        
-          val WavelengthCoverage: ColumnRef = col("c_spec_wavelength_coverage", wavelength_pm.embedded)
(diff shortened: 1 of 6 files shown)


modules/service/src/main/scala/lucuma/odb/graphql/table/ObservationView.scala
```diff
@@ -67,13 +67,30 @@ trait ObservationView[F[_]] extends BaseMapping[F] {
...
+            val Value: ColumnRef       = col("c_spec_wavelength_coverage",    wavelength_pm.embedded)
+          }
+          
+          object FocalPlaneAngle {
+            val SyntheticId: ColumnRef = col("c_spec_focal_plane_angle_id", observation_id.embedded)
+            val Value: ColumnRef       = col("c_spec_focal_plane_angle",    angle_µas.embedded)
+          }
+
+          val Resolution: ColumnRef         = col("c_spec_resolution",          pos_int.opt)          
+          val SignalToNoise: ColumnRef      = col("c_spec_signal_to_noise",     signal_to_noise.opt)
```
swalker2m, line 92: Why is `SignalToNoise` done this way as optional but `SignalToNoiseAt` a different way with `embedded` and a synthetic id? 

They are both optional in the GraphQL schema:

```
  # Requested signal to noise ratio
  signalToNoise: PosBigDecimal

  # Requested wavelength for the requested signal to noise
  signalToNoiseAt: Wavelength
```

The are both nullable columns:

```
  c_spec_signal_to_noise     numeric(5,2)    null default null, 
  c_spec_signal_to_noise_at  d_wavelength_pm null default null,
```

I probably did the `opt` vs `embedded` thing originally to make something work without really understanding the difference.

swalker2m (approved):
Thanks, I clearly still struggle with Grackle and the requirements it puts on your tables and views to make it work.  :+1:

modules/service/src/main/scala/lucuma/odb/graphql/table/ObservationView.scala
tpolecat: The difference is that `signalToNoise` is an atomic value and `signalToNoiseAt` is a GraphQL object (a `Wavelength`). Objects need keys (and a null key for an embedded object is how we indicate that the object itself is null).

I agree that Grackle is hard to use. It's getting better bit it's still frustrating.

swalker2m: Thanks, that makes sense.

PR #210, opened 2023-01-13
update core

tpolecat:


tpolecat:
what is happening? why are we pulling in sso-frontend-client and odb-schema?

```
	* edu.gemini:lucuma-core_3:0.62.0 (early-semver) is selected over {0.61.0, 0.61.1}
	    +- edu.gemini:lucuma-core-testkit_3:0.62.0            (depends on 0.62.0)
	    +- edu.gemini:lucuma-odb-schema_3:20230113-           (depends on 0.62.0)
	    +- edu.gemini:lucuma-sso-frontend-client_3:0.4.6      (depends on 0.61.1)
	    +- edu.gemini:lucuma-itc-client_3:0.4                 (depends on 0.61.0)
```

swalker2m:
> why are we pulling in sso-frontend-client and odb-schema?

`odb-schema` is a new module in `lucuma-odb` itself so it isn't really being pulled in per se.  I don't know about `sso-frontend-client` though

tpolecat:
Ohhh ok good, thanks. That makes sense.

swalker2m:
And it looks like `sso-backend-client`, which we need of course,  is setup to depend on `sso-frontend-client`.

PR #215, opened 2023-01-17
query vs transport

swalker2m:
This PR is a bit of an experiment.  The idea is to reconcile a few incompatible goals:

1. It would be nice for the ODB `schema` module to be the one source of truth for the ODB GraphQL schema and corresponding json codecs.
2. For servicing queries with json results, we have to produce json containing fully expanded objects.  For example, wavelengths in pm, nm, etc and angles in microarcseconds, milliarcseconds, etc.  Grackle can then trim the actual result down to match just what the user asked for.
3. For normal encoding though, only the canonical unit should be used (pm for wavelength, microarcseconds for angle, etc.)

To make this possible, a single shared `Decoder` is fine for all types but two different `Encoder`s are required depending on context.  I called these "query" and "transport" for lack of a better idea.  You get the appropriate codec depending on which alternative that you import.  

My heart is in the right place, I think, but I'm open to other names, to pointing out flaws in the implementation, or to being steered off this course altogether if it doesn't make sense.

### DIFF
--- a/modules/schema/src/main/scala/lucuma/odb/json/time.scala
+++ b/modules/schema/src/main/scala/lucuma/odb/json/time.scala
@@ -1,0 +1,105 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.odb.json
+
+import cats.syntax.bifunctor.*
+import cats.syntax.either.*
+import io.circe.Codec
+import io.circe.Decoder
+import io.circe.DecodingFailure
+import io.circe.Encoder
+import io.circe.HCursor
+import io.circe.Json
+import io.circe.refined._
+import io.circe.syntax._
+
+import java.math.RoundingMode.DOWN
+import java.time.Duration
+import java.time.temporal.ChronoUnit.MICROS
+import scala.util.Try
+
+object time {
+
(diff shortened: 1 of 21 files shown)


modules/schema/src/main/scala/lucuma/odb/json/time.scala
```diff
@@ -0,0 +1,105 @@
...
+import java.time.Duration
+import java.time.temporal.ChronoUnit.MICROS
+import scala.util.Try
+
+object time {
+
+  // N.B. We cannot represent any random `Duration` with the existing GraphQL
+  // schema.  Only a Duration expressed as a `Long` in milliseconds can be
+  // represented.  We need to either change the schema or make a specialized
+  // duration type.
```
swalker2m, line 27: Regardless of the fate of this PR, I'd like to revisit our use of `Duration` in another pass.  We're storing exposure times as an `interval` which doesn't align with `java.time.Duration` (and which need a non-negative constraint as well).  We also encode `Duration` as a time in microseconds as a `Long` value, but that also doesn't align.  This caused me pain and seems like it will continue to produce pain.  I'm going to try an `Interval` type along the lines of our `Timestamp`, I think.

cquiroz:
I haven't looked at this in detail but as a concept this seems good

I'd like to aim for a single source of truth on the schema and/or encoders so we don't have a separate one in lucuma-schemas.

tpolecat (approved):
👍🏻  I think this makes sense. Codecs that need to be consistent with the schema should live here.

PR #219, opened 2023-01-18
fix CatalogName mapping

tpolecat:
Quick PR to fix the CatalogName mapping.


modules/service/src/main/scala/lucuma/odb/graphql/table/TargetView.scala
```diff
@@ -44,7 +44,7 @@ trait TargetView[F[_]] extends BaseMapping[F] {
         val SyntheticId = col("c_sid_catalog_info_id", target_id.embedded)
         val Name        = col("c_sid_catalog_name", catalog_name.embedded)
         val Id          = col("c_sid_catalog_id", varchar.embedded)
-        val ObjectType  = col("c_sid_catalog_object_type", varchar.embedded)
+        val ObjectType  = col("c_sid_catalog_object_type", varchar.opt)
```
tpolecat, line 47: This field is nullable so it needs `opt` rather than `embedded` (which is for non-nullable fields of nullable embedded objects).

swalker2m (approved):
Looks great. :+1:

PR #225, opened 2023-01-24
TimeSpan - lucuma-core 0.63

swalker2m:
Updates to use the `lucuma.core.util.TimeSpan` type instead of `NonNegDuration`.  The `TimeSpan` maps directly to the GraphQL schema and is consistent with the way angles and wavelengths work.  The underlying database datatype is left as a Postgres `interval` which can hold longer time intervals.  This could be a potential issue if updated directly in the database instead of using the API.


tpolecat (approved):
This looks great :+1: thanks!

PR #231, opened 2023-01-25
bugfix

swalker2m:
Inadvertently changed an input field from an `input` object to a `type`, breaking introspection among other things.


tpolecat:
If we add a test that does an introspection query would it catch this kind of thing?

toddburnside (approved):
LGTM

swalker2m:
> If we add a test that does an introspection query would it catch this kind of thing?

I think so, an introspection attempt generated:

```
{
  "error": "Introspection must provide input type for arguments, but received: TimeSpan!."
}
```

Though this input is ultimately still unused, which is why the existing test cases pass I suppose.

tpolecat:
In any case 👍🏻 for this.

PR #233, opened 2023-01-26
wip: target edit

tpolecat:
Ok this adds target editing, along with sanity checks for some tricky cases with tracking and source profile (editing vs. replacing with a different type). It will probably end up needing more work but I'd like to get it merged so Todd can start testing again.


swalker2m (approved):
Wow, that must have been a lot of work!  Hopefully targets will be the most difficult case. Looks great.  :+1:

PR #238, opened 2023-01-31
Subscription/targetEdit

tpolecat:
This adds a mapping for the `targetEdit` subscription, which completes the mapping for `type Subscription`. The code for all of these is very similar so I'm going to look at abstracting it out in a followup PR.


swalker2m (approved):
Looks great.  :+1:

PR #243, opened 2023-02-06
fix proposal class discriminator

tpolecat:
There is a bug in Grackle that causes failed interface discrimination to succeed (!) which revealed a few bugs here. I need to make all of this more robust but this fixes it for now so Todd can be unblocked.


swalker2m (approved):
:+1:

toddburnside (approved):
LGTM

PR #244, opened 2023-02-07
clean up tests

tpolecat:
This factors out a bunch of duplicated code. It's still not great but it's better I think.


swalker2m (approved):
Nice. :+1:

PR #245, opened 2023-02-07
WIP: Mutation/cloneTarget

tpolecat:
Ok this adds Mutation/cloneTarget with two shameful hacks:

1. There is no obvious way in Grackle to return two values (the old and new targets in this case) so I added a view with every pair of target ids and am using that as the root of the result, which makes everything straightforward. `EXPLAIN ANALYZE` says this is fine and not slow, so I don't feel _that_ bad about it.

2. I found that during testing the change event streams for targets and observations were dying with non-existent portal errors, which I can't explain. I think maybe some changes in Skunk or fs2 are trying to be clever about when to close portals and are doing it too eagerly. In any case I replaced it with a straight interpolated string query and it works now. I will keep investigating.


swalker2m (approved):
This looks great!  I have nothing to add or subtract.  :+1:

PR #268, opened 2023-02-14
Initial Sequence Generation

swalker2m:
Adds initial sequence generation, without regard to any already executed datasets.  Adds a `sequence` field to `query` which generates the sequence and produces a JSON representation of it for Grackle to service the request.  Much of the PR involves adding JSON codecs for the sequence model from core.

The query extracts `GeneratorParams` from the database (which are also now used for ITC lookups since there is a lot of overlap).  Assuming there is enough data to call the ITC, the sequence can be generated.  By the way, as part of reworking the ITC call I simplified the `itc` query results into success or a service error (as opposed to the potential for a successful result containing an error).

The `sequence` module now has a `lucuma.odb.sequence.Generator` which accepts the `GeneratorParams` and creates a matching sequence assuming the ITC returns satisfactory results.  Individual observing modes are implemented in the `sequence` module under `lucuma.odb.sequence.<instrument>.<mode>.Generator`.


modules/schema/src/main/scala/lucuma/odb/json/sequence.scala
```diff
@@ -0,0 +1,224 @@
...
+      Json.obj(
+        "configChange" -> a.configChange.asJson,
+        "exposure"     -> a.exposure.asJson,
+        "readout"      -> a.readout.asJson,
+        "write"        -> a.write.asJson,
+        "total"        -> a.total.asJson
+      )
+    }
+
+  given given_Decoder_Step_GmosNorth: Decoder[Step.GmosNorth] =
```
tpolecat, line 56: Are you using these instances by name anywhere?

modules/sequence/src/main/scala/lucuma/odb/sequence/util/SequenceIds.scala
```diff
@@ -0,0 +1,71 @@
...
+import java.util.UUID
+import scala.util.Using
+
+
+object SequenceIds {
+
+  def namespace(
+    observationId: Observation.Id,
+    params:        GeneratorParams
+  ): UUID =
```
swalker2m, line 24: A potential point of discussion here.  The sequence IDs are generated using "type 3" UUIDs (i.e., a hash of input bytes).  I'd originally thought to generate the sequence without IDs and pass those bytes into `UUID.nameUUIDFromBytes`.  It occurs to me though that the sequence generation is deterministic from the generator params and observation id so in theory that is (almost) enough.  We'd need to take into account changes to the generator code itself by incorporating a GIT hash I think and in the ITC using the version information it returns with results.

An added benefit is that we can LRU cache the generated sequence by the "namespace" UUID.

modules/schema/src/main/scala/lucuma/odb/json/sequence.scala
swalker2m: No, but the automatically generated names clash.

modules/sequence/src/main/scala/lucuma/odb/sequence/util/SequenceIds.scala
tpolecat: :+1: based on Slack discussion this makes sense. You can't get to git info in the Heroku build but you can get the commit via (dyno-metadata)[https://devcenter.heroku.com/articles/dyno-metadata], which is a blunt object but it should work.

swalker2m: Added a `CommitHash` which is taken from the metadata as discussed.

swalker2m:
This is a bit lightly tested but it works at some level and @cquiroz is working on an offset position visualizer which needs the sequence.

modules/sequence/src/main/scala/lucuma/odb/sequence/util/CommitHash.scala
```diff
@@ -0,0 +1,67 @@
...
+
+package lucuma.odb.sequence.util
+
+import cats.Order
+import cats.Show
+import cats.syntax.eq.*
+import cats.syntax.option.*
+import lucuma.core.optics.Format
+
+import java.util.HexFormat
```
tpolecat, line 12: Need to update `system.properties` to JDK 17 for this class.

tpolecat (approved):
Wow, this looks great. Other than the JDK issue :+1:

PR #272, opened 2023-02-17
Brightness Edit Bug

swalker2m:
Adds a test case illustrating a problem reported by @rpiaggio.  In particular, when editing only the brightness information (ignoring SED), we are being fooled into thinking an entire point source definition is being provided because SED is optional. The result is that we swap out the existing point source for the new one without the SED instead of just updating the brightness information.

**RNC: EDIT:** ok I think this is fixed. I added code that favors editing where possible. Added analogous testcases for uniform and gaussian sources.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/graphql/input/sourceprofile/SourceProfileInput.scala
+++ b/modules/service/src/main/scala/lucuma/odb/graphql/input/sourceprofile/SourceProfileInput.scala
@@ -31,7 +31,7 @@ object SourceProfileInput {
           case (Some(point), None, None)     => Result(Point(point))
           case (None, Some(uniform), None)   => Result(Uniform(uniform))
           case (None, None, Some(gaussian))  => Result(gaussian)
-          case _                             => Result.failure("Expected exactly one of point, uniform, or guassian.")
+          case _                             => Result.failure("Expected exactly one of point, uniform, or gaussian.")
         }
     }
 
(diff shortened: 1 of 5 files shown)


modules/service/src/main/scala/lucuma/odb/graphql/input/sourceprofile/SourceProfileInput.scala
```diff
@@ -44,7 +44,7 @@ object SourceProfileInput {
       ) =>
         (rPoint, rUniform, rGaussian).parTupled.flatMap {
 
-          // If the user provides a full definitiom we can change the source profile
+          // If the user provides a full definition we can change the source profile
           case (Some(Left(p)), None, None) => Result(_ => Result(Point(p)))
```
swalker2m, line 48: This case is matched by the new test case instead of the one we want, down at line 53.

swalker2m:
```
values are not the same
=> Obtained
{
  "updateTargets" : {
    "targets" : [
      {
        "sourceProfile" : {
          "point" : {
            "bandNormalized" : {
              "sed" : null,
              "brightnesses" : [
                {
                  "band" : "R",
                  "value" : "15.0",
                  "units" : "VEGA_MAGNITUDE"
                }
              ]
            }
          }
        }
      }
    ]
  }
}
             "bandNormalized" : {
-              "sed" : null,
+              "sed" : {
+                "stellarLibrary" : "B5_III"
+              },
               "brightnesses" : [
```

tpolecat:
The problem is that we can't distinguish between replacing the `bandNormalized` with a new one that has no `sed` from editing the existing `bandNormalized` and ignoring the `sed`. We'll change it to favor the "editing" interpretation.

swalker2m (commented):
It seems I'm not allowed to approve this since I opened the PR originally :-/  Perhaps you can approve it though?  Regardless, thanks this looks great. `or` will come in handy elsewhere I'm sure.  :+1:

swalker2m:
I wasn't allowed to approve this, since I was the "author".  Regardless Rob did all the work and it looked good to me so I just merged it.

PR #275, opened 2023-02-21
"fix" issue with binning calculation for observations with more than one source profile

tpolecat:
"Fix" for issue discussed on Slack [here](https://gemini-software.slack.com/archives/CCDGZ8KNV/p1676999490923509). The fix is temporary; we need to talk to the astronomers and figure out how to select default binning for the observing more when there's more than one applicable source profile.


toddburnside (approved):
LGTM

swalker2m (approved):
Thanks!  :+1:

PR #280, opened 2023-02-23
Mutation/cloneObservation

tpolecat:
This adds `Mutation/cloneObservation` via
- `ObservationService::cloneObservation`, which delegates to
  - `AsterismService::cloneAsterism` and
  - `ObservingModeServices::cloneFunction`, which delegates to
    - `GmosLongSlitService::cloneNorth` and `cloneSouth`

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/ObservationService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/ObservationService.scala
@@ -60,6 +60,7 @@ import lucuma.odb.data.Tag
 import lucuma.odb.data.Timestamp
 import lucuma.odb.graphql.input.AirMassRangeInput
 import lucuma.odb.graphql.input.ConstraintSetInput
+import lucuma.odb.graphql.input.CloneObservationInput
 import lucuma.odb.graphql.input.ElevationRangeInput
 import lucuma.odb.graphql.input.HourAngleRangeInput
 import lucuma.odb.graphql.input.ObservationPropertiesInput
@@ -94,6 +95,10 @@ sealed trait ObservationService[F[_]] {
     SET:   ObservationPropertiesInput.Edit,
     which: AppliedFragment
   ): F[Result[List[Observation.Id]]]
+
+  def cloneObservation(
+    input: CloneObservationInput
+  ): F[Result[Observation.Id]]
 
 }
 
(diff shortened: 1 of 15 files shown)


modules/service/src/main/scala/lucuma/odb/service/ObservationService.scala
```diff
@@ -279,6 +284,65 @@ object ObservationService {
...
+                case Some(oid2) =>
+
+                  // The remaining clone operations yield no results and can't fail (logically, anyway)
+                  val cloneTheRest = 
+                    List(
+                    Statements.cloneAsterism(input.observationId, oid2),
+                    Statements.cloneGmosNorthStatic(input.observationId, oid2),
+                    Statements.cloneGmosNorthDynamic(input.observationId, oid2),
+                    Statements.cloneGmosNorthLongSlit(input.observationId, oid2),
+                  ).traverse_(af => session.prepareR(af.fragment.command).use(_.execute(af.argument)))
```
swalker2m, line 321: How do we know it is GMOS North?  Where is GMOS South?

modules/service/src/main/scala/lucuma/odb/service/ObservationService.scala
```diff
@@ -645,6 +709,214 @@ object ObservationService {
...
+          c_mos_pre_imaging,
+          c_nod_and_shuffle,
+          c_stage_mode
+        FROM t_gmos_north_static
+        WHERE c_observation_id = $observation_id
+      """.apply(newOid ~ oldOid)
+
+    def cloneGmosNorthDynamic(oldOid: Observation.Id, newOid: Observation.Id): AppliedFragment =
+      sql"""
+        INSERT INTO t_gmos_north_dynamic (
```
swalker2m, line 833: There is an `ObservingModeServices`, which I'd thought could be used to manipulate the specific table for the observing mode via the matching service.  For example, `GmosLongSlitService` for now with others added as new modes are introduced.  It's used elsewhere in `ObservationService` like: https://github.com/gemini-hlsw/lucuma-odb/blob/main/modules/service/src/main/scala/lucuma/odb/service/ObservationService.scala#L213-L256.

swalker2m (commented):
This is looking good, but I'm confused about observing modes.

modules/service/src/main/scala/lucuma/odb/service/ObservationService.scala
tpolecat: I'm just copying rows without regard to what the actual mode is, but it would probably be more efficient to read in the mode and do something less dumb. And yes I supposed it should also support GS.

modules/service/src/main/scala/lucuma/odb/service/ObservationService.scala
tpolecat: Ok thanks, I didn't know about this. :+1:

swalker2m (approved):
Looks great!  :+1:

PR #282, opened 2023-02-23
comment out unused query fields

tpolecat:
These query fields aren't needed for XT and may not be needed at all, so for the sake of shortening our checklist I commented them out.

### DIFF
--- a/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -4626,13 +4626,13 @@ enum QuasarSpectrum {
 
 type Query {
   # All science targets (if any) for the given observation (or environment). Set `includeDeleted` to true to include any deleted targets in the asterism.
-  asterism(
-    # Observation ID
-    observationId: ObservationId!
+  # asterism(
+  #   # Observation ID
+  #   observationId: ObservationId!
 
-    # Set to true to include deleted values
-    includeDeleted: Boolean! = false
-  ): [Target!]!
+  #   # Set to true to include deleted values
+  #   includeDeleted: Boolean! = false
+  # ): [Target!]!
 
   # Observations grouped by commonly held science asterisms
   asterismGroup(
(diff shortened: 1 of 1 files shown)


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -4626,13 +4626,13 @@ enum QuasarSpectrum {
...
-  asterism(
-    # Observation ID
-    observationId: ObservationId!
+  # asterism(
+  #   # Observation ID
+  #   observationId: ObservationId!
 
-    # Set to true to include deleted values
-    includeDeleted: Boolean! = false
-  ): [Target!]!
```
swalker2m, line 4635: I suppose you can ask for an observation's targets via `asterismGroup` with the observation set to equal a particular id but it seems a bit awkward.  I would think we should have this one eventually.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -4762,34 +4762,34 @@ type Query {
...
-    LIMIT: NonNegInt
+  #   # Limits the result to at most this number of matches (but never more than 1000).
+  #   LIMIT: NonNegInt
 
-    # Set to true to include deleted values
-    includeDeleted: Boolean! = false
-  ): ObservingModeGroupSelectResult!
+  #   # Set to true to include deleted values
+  #   includeDeleted: Boolean! = false
+  # ): ObservingModeGroupSelectResult!
```
swalker2m, line 4777: This will be needed if we're going to have a "Configurations" tab.  This isn't needed for XT?

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -4798,25 +4798,25 @@ type Query {
...
 
   # Target environment for the given observation (or environment id)
-  targetEnvironment(
-    # Observation ID
-    observationId: ObservationId!
-  ): TargetEnvironment
+  # targetEnvironment(
+  #   # Observation ID
+  #   observationId: ObservationId!
+  # ): TargetEnvironment
```
swalker2m, line 4804: Target environment will eventually have guide stars, blind offsets etc.  I suppose we could just get the asterism from there instead of a dedicated `asterism` query if we're trying minimize.

swalker2m (approved):
:+1: We can bring them back when/if needed.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
tpolecat: I think it was determined that the configurations tab isn't happening for XT. At least for the moment.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
tpolecat: Yes, I was thinking you could just query the observation and pull out the target environment that way. As far as I know there isn't a requirement to provide convenience methods for API end-users in XT.

PR #291, opened 2023-02-28
fix obs name computation

### LINKED ISSUE #250: Deleted targets still appear in the Observation Title
If the asterism for an observation contains a deleted target, that target name still appears in the observation title. However, the deleted target (correctly) does not appear in the asterism list for the observation.

tpolecat:
resolves #250


swalker2m (approved):
Nice!  :+1:  I think I would have tried to compute the title every time it is requested but this has got to be faster.

PR #292, opened 2023-02-28
recompute asterism groups when targets are deleted

### LINKED ISSUE #249: Deleted targets/observations being returned by `asterismGroup` query.
They are being returned even if the `includeDeleted` parameter is excluded or set to false.

tpolecat:
Resolves #249, leaves the remainder in #193


swalker2m (approved):
I see it is still marked as a Draft but it seems good to me.  :+1:

PR #303, opened 2023-03-03
Execution config hack

swalker2m:
`ExecutionConfig` is an interface, with two existing implementations `GmosNorthExecutionConfig` and `GmosSouthExecutionConfig`.  In theory there will be one per instrument.  The structure of these types is the same; only leaf values like filters, FPU options, etc. differ.

As I understand it, when a pure JSON result is calculated the structure of the object is used to determine the type.  Since it cannot distinguish the two types based on structure it ends up picking the first matching one.

It's suboptimal because it clutters the API, but this can be "solved" by adding a bogus field that differs for each type.


tpolecat (approved):
Looks "good" :+1:

PR #308, opened 2023-03-08
respect mah includeDeleted

### LINKED ISSUE #307: targetGroup query is returning deleted targets
I returns them even if the `includeDeleted` flag is explicitly set to `false`.

To reproduce, create a target, then set it's existence value to `DELETED`.

Run this query with the appropriate program id:
``` graphQL
query {
  targetGroup(programId: "p-102", includeDeleted: false) {
    matches {
      target {
        id
        existence
      }
    }
  }
}
```

The deleted target will be included in the results.

tpolecat:
image

resolves #307


swalker2m (approved):
Ok, then!  :+1:

PR #319, opened 2023-03-20
Pre-Smart Gcal Prep

swalker2m:
A small update to match `lucuma-core` 0.71.0, which is required for smart gcal.


modules/service/src/main/resources/db/migration/V0220__enum_updates.sql
```diff
@@ -0,0 +1,22 @@
...
+
+INSERT INTO t_gmos_north_fpu values ('Ifu2Slits', 'IFU-2', 'IFU 2 Slits', null, 33500000);
+INSERT INTO t_gmos_north_fpu values ('IfuBlue', 'IFU-B', 'IFU Left Slit (blue)', null, 31750000);
+INSERT INTO t_gmos_north_fpu values ('IfuRed', 'IFU-R', 'IFU Right Slit (red)', null, 35250000);
+
+UPDATE t_gmos_north_long_slit SET c_fpu = 'Ifu2Slits' WHERE c_fpu = 'Ifu1';
+UPDATE t_gmos_north_long_slit SET c_fpu = 'IfuBlue' WHERE c_fpu = 'Ifu2';
+UPDATE t_gmos_north_long_slit SET c_fpu = 'IfuRed' WHERE c_fpu = 'Ifu3';
+
+DELETE FROM t_gmos_north_fpu WHERE c_tag LIKE 'Ifu_';
```
swalker2m, line 17: Ordinarily we wouldn't be updating primary keys this way I think, but these will be lightly used if used at all.  `Ifu1` for 2 slits and `Ifu2` for something else is a bit confusing and didn't match the enum tags so I wanted to straighten it out now.  In general I wonder if we should define the foreign key references like the ones in `t_gmos_north_long_slit` to cascade on update?

tpolecat (approved):
:+1:

PR #322, opened 2023-03-22
Add Attachments endpoints

toddburnside:
I have not yet written any tests for this. I wanted to get feedback before undertaking that.

Also, I probably need to add more tracing or logging so we have visibility into any AWS errors.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/Config.scala
+++ b/modules/service/src/main/scala/lucuma/odb/Config.scala
@@ -7,7 +7,13 @@ import cats.Show
 import cats.effect._
 import cats.syntax.all._
 import ciris._
+import ciris.refined._
 import com.comcast.ip4s.Port
+import eu.timepit.refined.cats._
+import eu.timepit.refined.types.string.NonEmptyString
+import fs2.aws.s3.models.Models.BucketName
+import fs2.aws.s3.models.Models.FileKey
+import lucuma.core.model.Program
 import lucuma.core.model.User
 import lucuma.itc.client.ItcClient
 import lucuma.odb.sequence.util.CommitHash
@@ -33,6 +39,7 @@ case class Config(
   serviceJwt: String,           // Only service users can exchange API keys, so we need a service user JWT.
   honeycomb:  Config.Honeycomb, // Honeycomb config
   database:   Config.Database,  // Database config
+  aws:        Config.Aws,       // AWS config
   domain:     String,           // Domain, for CORS headers
   commitHash: CommitHash        // From Heroku Dyno Metadata
 ) {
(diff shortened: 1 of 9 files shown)


modules/service/src/main/scala/lucuma/odb/Config.scala
```diff
@@ -141,6 +148,49 @@ object Config {
...
+          NonEmptyString.from(b).toOption
+          .map(BucketName.apply) 
+        } 
+      }
+
+    lazy val fromCiris: ConfigValue[Effect, Aws] = (
+      envOrProp("CLOUDCUBE_ACCESS_KEY_ID").as[NonEmptyString],
+      envOrProp("CLOUDCUBE_SECRET_ACCESS_KEY").as[NonEmptyString].redacted,
+      envOrProp("CLOUDCUBE_URL").as[Uri].as[Uri.Path].as[NonEmptyString],
+      envOrProp("CLOUDCUBE_URL").as[Uri].as[BucketName],
```
tpolecat, line 189: CloudCube is the Heroku service we're using, I guess?

modules/service/src/main/scala/lucuma/odb/Main.scala
```diff
@@ -112,25 +114,28 @@ object Main extends IOApp {
...
+      itcClient        <- itcClientResource
+      ssoClient        <- ssoClientResource
+      userSvc          <- pool.map(UserService.fromSession(_))
+      middleware       <- Resource.eval(ServerMiddleware(domain, ssoClient, userSvc))
+      graphQLRoutes    <- GraphQLRoutes(itcClient, commitHash, ssoClient, pool, SkunkMonitor.noopMonitor[F], GraphQLServiceTTL, userSvc)
+      attachmentSvc    <- pool.flatMap(ses => AttachmentService.fromConfigAndSession(awsConfig, ses))
+      attachmentRoutes =  AttachmentRoutes.apply[F](attachmentSvc, ssoClient, awsConfig.fileUploadMaxMb)
     } yield { wsb =>
-      middleware(routes(wsb))
+      middleware(attachmentRoutes <+> graphQLRoutes(wsb))
```
tpolecat, line 138: Might as well swap these since most requests will go to `/graphql`.

modules/service/src/main/scala/lucuma/odb/graphql/AttachmentRoutes.scala
```diff
@@ -0,0 +1,92 @@
...
+    object DescriptionMatcher extends OptionalQueryParamDecoderMatcher[String]("description")
+
+    val pathBase = "attachment"
+
+    val routes = HttpRoutes.of[F] {
+      case req @ GET -> Root / pathBase / ProgramId(programId) / AttachmentId(attachmentId) =>
+        ssoClient.require(req) { user =>
+          attachmentService.getAttachment(user, programId, attachmentId).flatMap {
+            case Left(exc)     => exc.toResponse
+            case Right(stream) => Async[F].delay(Response(Status.Ok, body = stream))
```
tpolecat, line 63: The `Response` constructor is pure so you should just be able to say `.pure[F]` here.

modules/service/src/main/scala/lucuma/odb/service/AttachmentService.scala
```diff
@@ -0,0 +1,358 @@
...
+          session.prepareR(stmt).use( pg =>
+            pg.option(af.argument)
+              .flatMap{
+                case None    => Async[F].raiseError(AttachmentException.FileNotFound)
+                case Some(s) => 
+                  // should be non empty because of a database check
+                  NonEmptyString.from(s)
+                    .toOption
+                    .fold(Async[F].raiseError(AttachmentException.InvalidName("File name is missing")))(
+                      Async[F].delay)
```
tpolecat, line 217: `.pure[F]` here

modules/service/src/main/scala/lucuma/odb/service/AttachmentService.scala
```diff
@@ -0,0 +1,358 @@
...
+          f.onError { case e => Trace[F].attachError(e, ("error", true)) }
+        }
+
+      // TODO: eventually will probably want to check for write access for uploading/deleting files.
+      def withAccess[A](user: User, programId: Program.Id)(f: F[A]): F[A] = user match {
+        // guest users not allowed to upload files - at least for now.
+        case GuestUser(_) => Async[F].raiseError(AttachmentException.Forbidden)
+        case _            =>
+          ProgramService.fromSessionAndUser(session, user)
+            .userHasAccess(programId)
```
tpolecat, line 162: This is going to end up running in a different transaction on the session, so you probably want to internalize the access check and run your operations within a transaction.

tpolecat (commented):
This is looking great. A few comments.

modules/service/src/main/scala/lucuma/odb/Config.scala
toddburnside: Yep. And it sets these variables for us.

modules/service/src/main/scala/lucuma/odb/Main.scala
toddburnside: Good point.

modules/service/src/main/scala/lucuma/odb/service/AttachmentService.scala
tpolecat: :+1: Yeah the second case we probably don't want to hold the transaction open. If access is revoked during the upload they'll get a freebie but I think that's a case we don't need to worry about.

swalker2m: We could always check again after the upload, if so inclined.  I agree not worth worrying about.

tpolecat (approved):
This looks great! :+1: :+1:

PR #323, opened 2023-03-23
GMOS Long Slit Smart GCal

swalker2m:
Smart Gcal defines the process of taking an instrument configuration and finding the most appropriate GCal unit configuration for a flat or arc.  So-called "smart" steps in a generated sequence will be replaced with fully configured Gcal steps that match the instrument configuration at that point.

This PR is concerned with defining and populating the necessary tables.  In particular, there is a `t_gcal` table for the gcal configurations that will be shared by all instruments and there will be a per-instrument `t_smart_{inst_name}` table that references it.  Roughly speaking, the instrument table will define the search key using the instrument's filters, dispersers, etc. and the gcal table defines the corresponding value.

The PR adds:

* A smartgcal module that is mostly concerned with parsing the existing Smart GCal configuration files.
* Database schema definitions
* A _programmatic_ _repeatable_ database migration that reads the configuration file, parses it, loads it into a temporary table, and then inserts the values into the corresponding smart gcal tables.
* A `SmartGcalService` that facilitates lookups.
* Expansion logic to replace `SmartGcal` steps with appropriately configured `Gcal` steps.

Unfortunately to make this reasonably fast, I need to forgo `skunk` and stick with simplistic SQL strings and JDBC.  The problem is that I could find no way in `skunk` to bulk insert data (a la the SQL `COPY FROM STDIN` command).  Nevertheless, I am using `skunk` encoders to format the data for the copy which is a bit weird.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/Main.scala
+++ b/modules/service/src/main/scala/lucuma/odb/Main.scala
@@ -159,13 +159,13 @@ object Main extends IOApp {
         password = config.password.some
       )
 
-    val drop   = sql"""DROP DATABASE "#${config.database}""""(Void)
-    val create = sql"""CREATE DATABASE "#${config.database}""""(Void)
+    val drop   = sql"""DROP DATABASE "#${config.database}"""".command
+    val create = sql"""CREATE DATABASE "#${config.database}"""".command
 
     session.use { s =>
       for {
-        _ <- s.prepare(drop.fragment.command).flatMap(_.execute(drop.argument)).void
-        _ <- s.prepare(create.fragment.command).flatMap(_.execute(create.argument)).void
+        _ <- s.execute(drop).void
+        _ <- s.execute(create).void
       } yield()
     }
 


modules/service/src/main/scala/db/migration/R__SmartGmosNorth.scala
```diff
@@ -0,0 +1,109 @@
...
+      r.key  ~
+      r.value
+    }
+
+  val (tmp, inst) = SmartGcalTable.forInstrument(
+    GmosNorth,
+    Col("c_step_order", "int8"),
+    KeyColumns,
+    NonEmptyList.one(Col("c_exposure_time", "interval"))
+  )
```
swalker2m, line 100: I'll just pinpoint one of the lamest parts of the loader.  It uses a `skunk` encoder to write the data for the `COPY FROM STDIN`, the order of the encoding has to match the table column order but there's no guarantee of this anywhere.  I'm guessing there could be a way to group the encoder with the column definition and address this.

modules/service/src/main/scala/db/migration/R__SmartGmosNorth.scala
```diff
@@ -0,0 +1,109 @@
...
+import java.io.InputStream
+
+/**
+ * Repeatable Smart GCal configuration loader for GMOS North.  Located,
+ * instantiated and executed by flyway.
+ */
+class R__SmartGmosNorth extends SmartGcalMigration[TableRow]("GMOS North") {
+
+  val definitionFiles: NonEmptyList[(String, IO[InputStream])] =
+    gcalFilesFromClasspath("GMOS-N_ARC", "GMOS-N_FLAT")
```
swalker2m, line 32: These are taken from the classpath at the moment, but we could consider reading them from, say, AWS at some point and allow them to be uploaded, checked and applied like they are in `ocs`.

modules/service/src/main/scala/db/migration/SmartGcalMigration.scala
```diff
@@ -0,0 +1,59 @@
...
+      definitionFiles
+        .map { case (_, is) => fs2.io.readInputStream(is, ByteChunkSize, closeAfterUse = true) }
+        .reduce
+        .through(fs2.compression.checksum.crc32)
+        .compile
+        .to(Array)
+        .unsafeRunSync()
+
+    BigInt(bytes).intValue
+  }
```
swalker2m, line 50: GMOS North alone is taking between 5 and 6 seconds to load, but this will guarantee that is is only loaded when actually updated.

modules/service/src/main/scala/db/migration/R__SmartGmosNorth.scala
tpolecat: Yeah maybe they could be in another GH repo. The same commit hash logic would work but we would have some insulation in case they want to roll back a change or whatever. What you have now is fine of course.

modules/service/src/main/scala/db/migration/R__SmartGmosNorth.scala
tpolecat: Yeah this is a general problem with multi-column codecs. It looks like the constraints will probably catch most errors ... ?

modules/service/src/main/scala/db/migration/SmartGcalMigration.scala
tpolecat: Oh the checksum takes that long? Wow. I guess we have to do it though.

tpolecat (commented):
This all looks good so far. Very clean.

modules/service/src/main/scala/db/migration/SmartGcalMigration.scala
swalker2m: I didn't write that very clearly.  The checksum runs in milliseconds and avoids a 5 to 6 second load when the data file has been updated.

modules/service/src/main/scala/db/migration/R__SmartGmosNorth.scala
swalker2m: If it doesn't line up then the `COPY` will usually fail at runtime when the migration is attempted.  Flyway does each migration in a transaction so it all gets rolled back.  I suppose you could reverse, say, two columns of the same type and import the data incorrectly though.

modules/service/src/test/scala/lucuma/odb/graphql/query/smartgcal.scala
```diff
@@ -0,0 +1,618 @@
...
+      _ <- define(3, low  = 600_000, expTimeSec = 3)
+
+      // multi steps
+      _ <- define(4, stepOrder = 10, disperser = GmosNorthGrating.B600_G5303, expTimeSec = 4)
+      _ <- define(5, stepOrder =  9, disperser = GmosNorthGrating.B600_G5303, expTimeSec = 5)
+
+      // step count
+      _ <- define(6, disperser = GmosNorthGrating.B600_G5307, count = 2, expTimeSec = 6)
+
+    } yield ()
```
swalker2m, line 131: Defines various smart gcal lookups for the test cases.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -492,8 +492,11 @@ enum GcalContinuum {
   # GcalContinuum IrGreyBodyHigh
   IR_GREY_BODY_HIGH
 
-  # GcalContinuum QuartzHalogen
-  QUARTZ_HALOGEN
+  # GcalContinuum QuartzHalogen 5W
+  QUARTZ_HALOGEN5
+
+  # GcalContinuum QuartzHalogen 100W
+  QUARTZ_HALOGEN100
```
swalker2m, line 499: Will this update break Explore?  There are two Quartz Halogen lamps so we need to distinguish them.

swalker2m (commented):
Okay, GMOS North smart gcal seems to be working.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
toddburnside: I think we're OK since the enum value has been added to lucuma-core. If there is a problem lucuma-schemas and explore probably need to be updated to the new lucuma-core.

modules/service/src/main/scala/lucuma/odb/logic/Generator.scala
```diff
@@ -0,0 +1,281 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.odb.logic
```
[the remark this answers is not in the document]
tpolecat: Yeah this seems fine. I suppose these things could be services themselves (it was not my intent that services only be for db interaction) but it's fine like this.

modules/service/src/test/scala/lucuma/odb/graphql/OdbSuite.scala
```diff
@@ -232,6 +233,27 @@ abstract class OdbSuite(debug: Boolean = false) extends CatsEffectSuite with Tes
...
+    super.beforeAll()
+
+    dbInitialization.foreach { init =>
+      Main
+        .databasePoolResource[IO](databaseConfig)
+        .flatten
+        .use(init)
+        .unsafeRunSync()
+    }
+  }
```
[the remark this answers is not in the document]
tpolecat: This seems fine.

modules/smartgcal/src/main/scala/lucuma/odb/smartgcal/parsers/util.scala
```diff
@@ -0,0 +1,98 @@
...
+    def matchingRegex(pat: String): Either[String, NonEmptyList[A]] =
+      catching(classOf[PatternSyntaxException])
+        .either(new Regex(pat))
+        .leftMap(_ => s"Invalid regex pattern $pat")
+        .flatMap { r =>
+          m.toList.collect { case (key, value) if r.matches(key) => value } match {
+            case Nil     => formatNoMatch(s"Pattern '$pat' matched nothing").asLeft
+            case a :: as => NonEmptyList(a, as).asRight
+          }
+        }
```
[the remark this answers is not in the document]
tpolecat: yikes

tpolecat (approved):
Wow this was a lot of work. It looks great and is really well organized. I have a few important comments but :+1: merge at will.

modules/service/src/main/scala/lucuma/odb/Main.scala
```diff
@@ -144,37 +144,109 @@ object Main extends IOApp {
...
+    server(reset, skipMigration).use(_ => Concurrent[F].never[ExitCode])
 
   /** Our actual entry point. */
   def run(args: List[String]): IO[ExitCode] = {
     implicit val log: SelfAwareStructuredLogger[IO] =
       Slf4jLogger.getLoggerFromName[IO]("lucuma-odb")
 
-    val skipMigration = args.contains("-skip-migration")
+    val reset         = Arg.ResetDatabase(args)
+    val skipMigration = Arg.SkipMigration(args)
```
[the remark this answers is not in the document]
swalker2m: This is a good idea, I'll circle back to it once I finish some release tasks.

PR #335, opened 2023-04-04
Use Decline for args

swalker2m:
Switches the argument parsing to use `Decline`.  I'm aware I may be "doing it wrong" so don't hesitate to say so.  Without a command name / header to use for the parser, it yields help messages like:

```
service Usage:  [--reset] [--skip-migration]
service 
service 
service 
service Options and flags:
service     --help
service         Display this help text.
service     --reset
service         Drop and recreate the database before starting.
service     --skip-migration
service         Skip database migration on startup.
service ... finished with exit code 0
```

which seems okay to me, I think.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,1 @@
-web: ./modules/service/target/universal/stage/bin/lucuma-odb-service -Dfile.encoding=UTF-8 -Dcats.effect.stackTracingMode=DISABLED -Dcats.effect.tracing.mode=none
+web: ./modules/service/target/universal/stage/bin/lucuma-odb-service -Dfile.encoding=UTF-8 -Dcats.effect.stackTracingMode=DISABLED -Dcats.effect.tracing.mode=none serve
--- a/README.md
+++ b/README.md
@@ -45,10 +45,10 @@ Alternatively, you can run the app from within SBT with `service/reStart`
 running `docker-compose` `down` and then `up` as described above.  You can
 supply optional arguments to simplify development though:
 
-* `-reset` - Drops then creates the database for you. Do this after cycling
+* `--reset` - Drops then creates the database for you. Do this after cycling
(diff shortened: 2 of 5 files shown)


toddburnside (approved):
Seems fine to me.

tpolecat:
I don't want to make this any more complicated than it needs to be, but it seems like maybe we should use `CommandIOApp` and have a `serve` command like SSO does. This seems more natural. https://github.com/gemini-hlsw/lucuma-sso/blob/master/modules/service/src/main/scala/Main.scala#L45

swalker2m:
> I don't want to make this any more complicated than it needs to be, but it seems like maybe we should use CommandIOApp and have a serve command like SSO does

Okay thanks I wasn't sure what to use for the name and header, but I'll copy what SSO does.

swalker2m:
I updated this to use `CommandIOApp` following the SSO model.  It makes use with revolver a bit awkward perhaps because of the superfluous `serve` but it's fine with me if it doesn't bother anyone else:

```
service/reStart serve --reset
```

for example.  I could make `reset` a command on par with `serve` but then you'd need two `reStart`s to start up with a clean database.

Procfile
```diff
@@ -1 +1 @@
-web: ./modules/service/target/universal/stage/bin/lucuma-odb-service -Dfile.encoding=UTF-8 -Dcats.effect.stackTracingMode=DISABLED -Dcats.effect.tracing.mode=none
+web: ./modules/service/target/universal/stage/bin/lucuma-odb-service -Dfile.encoding=UTF-8 -Dcats.effect.stackTracingMode=DISABLED -Dcats.effect.tracing.mode=none serve
```
swalker2m, line 1: I think we'll need `serve` here?

tpolecat: Yes, correct.

tpolecat (approved):
:+1:

PR #339, opened 2023-04-06
Bump core and itc

rpiaggio:
This should fix the NPE crash


swalker2m (approved):
Thanks for handling this. 👍

swalker2m:
D'oh, looks like SSO has to be brought up to the latest `lucuma-core` as well.

PR #341, opened 2023-04-06
Add updateAttachments mutation

toddburnside:


modules/service/src/main/scala/lucuma/odb/graphql/input/ObsAttachmentPropertiesInput.scala
```diff
@@ -0,0 +1,29 @@
...
+import lucuma.odb.data.Nullable
+import lucuma.odb.data.Tag
+import lucuma.odb.graphql.binding._
+
+object ObsAttachmentPropertiesInput {
+
+  case class Edit(
+    description: Nullable[NonEmptyString],
+    checked: Option[Boolean]
+  )
```
swalker2m, line 19: There's only `Edit` because creation is handled by the REST API, correct?

swalker2m (approved):
This looks excellent to me.  :+1:

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -2696,9 +2725,17 @@ type Attachment {
...
-# Metadata for `enum AttachmentType`
-type AttachmentTypeMeta {
-  tag:        AttachmentType!
+input ObsAttachmentPropertiesInput {
+  # The description field may be unset by assigning a null value, or ignored by skipping it altogether
+  description: NonEmptyString
+
+  # The checked status can be set, or ignored by skipping it altogether
+  checked: Boolean
+}
```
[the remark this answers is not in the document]
toddburnside: No. Although I suppose we could allow changing the file name since it is no longer used as part of the key for S3. 

I'm not allowing changing the attachment type because if the attachment is, for example, already assigned to one or more observation as a finder chart, we'd have to figure out what to do if they changed the type to MOS mask. Or maybe we'd have to check and only allow changing the type if it wasn't assigned. This seemed more complicated and I'm hoping we can get away with it.

modules/service/src/main/scala/lucuma/odb/graphql/input/ObsAttachmentPropertiesInput.scala
toddburnside: Correct. The REST API is the only place that knows how to deal with S3. So creation and delete only happen there.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -5707,6 +5744,54 @@ type WavelengthDither {
...
+
+  # A nested attachment filter that must not match in order for the NOT itself to match.
+  NOT: WhereObsAttachment
+
+  # Matches the attachment ID.
+  id: WhereOrderObsAttachmentId
+
+  # Matches the attachment file name.
+  fileName: WhereOptionString
+}
```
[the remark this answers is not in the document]
swalker2m: Yes, we haven't added all the properties to the observation `WHERE` yet either.  File size might not be as useful as the type and checked.  I guess if it is easy though, why not.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -2681,13 +2710,13 @@ type AsterismGroupSelectResult {
...
+scalar ObsAttachmentId
 
-# Program Attachment
-type Attachment {
-  id:             AttachmentId!
-  attachmentType: AttachmentType!
+# Program Observation Attachment
+type ObsAttachment {
+  id:             ObsAttachmentId!
+  attachmentType: ObsAttachmentType!
```
cquiroz, line 2719: Can we deduce the mime type from the attachment type?

PR #358, opened 2023-04-15
Add missing spaces in asterism deletion SQL statements

rpiaggio:


swalker2m (approved):
Thanks for fixing these!  :+1:

PR #362, opened 2023-04-17
Planned time enums

swalker2m:
This is part of the work required to add planned time estimates.  It introduces an `Enums` type used for working with enums defined in the database.  In particular `TimeEstimate`s are named values corresponding to something for which there is an associated time cost.  For example, moving the science fold requires 15 seconds.  We want

* time estimate values defined in the database, and
* values that we can reference by name (e.g., `TimeEstimate.ScienceFold`)

For this reason there is a `TimeEstimateMeta` class with the data itself _and_ a Scala `TimeEstimate` `enum` that references the individual values.  This association between `enum` value and metadata needs to be done upon once upon startup and needs to fail if the linking entries cannot be found.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/graphql/OdbMapping.scala
+++ b/modules/service/src/main/scala/lucuma/odb/graphql/OdbMapping.scala
@@ -38,9 +38,7 @@ import lucuma.itc.client.ItcResult
 import lucuma.itc.client.SpectroscopyModeInput
 import lucuma.itc.client.SpectroscopyResult
 import lucuma.odb.graphql._
-import lucuma.odb.graphql.enums.AttachmentTypeEnumType
-import lucuma.odb.graphql.enums.FilterTypeEnumType
-import lucuma.odb.graphql.enums.PartnerEnumType
+import lucuma.odb.graphql.enums.Enums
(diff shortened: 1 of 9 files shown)


modules/service/src/main/scala/lucuma/odb/graphql/OdbMapping.scala
```diff
@@ -102,11 +100,11 @@ object OdbMapping {
...
     itcClient:  ItcClient[F],
-    commitHash: CommitHash
-  ):  F[Mapping[F]] =
-    Trace[F].span(s"Creating mapping for ${user0.displayName} (${user0.id}, ${user0.role})") {
-      database.use(enumSchema(_)).map { enums =>
-        new SkunkMapping[F](database, monitor) with BaseMapping[F]
+    commitHash: CommitHash,
+    enums:      Enums
+  ):  Mapping[F] =
+        new SkunkMapping[F](database, monitor)
```
swalker2m, line 106: Creating the `OdbMapping` no longer requires a trip to the database but I left the indentation for now to avoid obscuring what really changes.

tpolecat (approved):
This looks great :+1:

PR #364, opened 2023-04-18
trusted

swalker2m:
Adds a bit of configuration from the normal database session pool to the single-use session that I employ for reading enumerations.

There is a startup issue in development, which is presumably related to #362.  Checking the difference in configuration, this seems like the only applicable discrepancy so I'm trying it here.

```
2023-04-18T12:33:14.928251+00:00 app[web.2]: [[34mINFO [0;39m] [36mlucuma-odb[0;39m -  
2023-04-18T12:33:14.928291+00:00 app[web.2]: [[34mINFO [0;39m] [36mlucuma-odb[0;39m - CommitHash.: f96f9fc96b736f141b112db128792cce7a90c8a0 
2023-04-18T12:33:14.928338+00:00 app[web.2]: [[34mINFO [0;39m] [36mlucuma-odb[0;39m - CORS domain: lucuma.xyz 
2023-04-18T12:33:14.928370+00:00 app[web.2]: [[34mINFO [0;39m] [36mlucuma-odb[0;39m - ITC Root...: https://itc-development.herokuapp.com/itc 
2023-04-18T12:33:14.928405+00:00 app[web.2]: [[34mINFO [0;39m] [36mlucuma-odb[0;39m - Port.......: 35049 
2023-04-18T12:33:14.928434+00:00 app[web.2]: [[34mINFO [0;39m] [36mlucuma-odb[0;39m -  
2023-04-18T12:33:15.154808+00:00 app[web.2]: skunk.exception.StartupException: 
2023-04-18T12:33:15.154810+00:00 app[web.2]: 🔥  Startup negotiation failed.
2023-04-18T12:33:15.154810+00:00 app[web.2]: 🔥  
2023-04-18T12:33:15.154811+00:00 app[web.2]: 🔥  
2023-04-18T12:33:15.154813+00:00 app[web.2]: 🔥  Postgres FATAL 28000 raised in ClientAuthentication (auth.c:536)
2023-04-18T12:33:15.154814+00:00 app[web.2]: 🔥  
2023-04-18T12:33:15.154814+00:00 app[web.2]: 🔥    Problem: [36mNo pg_hba.conf entry for host "54.175.235.216", user[0m
2023-04-18T12:33:15.154815+00:00 app[web.2]: 🔥  [36m           "lrwzaxxcyugqer", database "d7koa1u6u30mh2", no encryption.[0m
2023-04-18T12:33:15.154816+00:00 app[web.2]: 🔥  
[...]

PR #365, opened 2023-04-18
shuffle enum load

swalker2m:
Continued debugging (see #362, #364).  This will shuffle the enum loading until a bit later in order to use the normal session pool.


tpolecat (approved):
:+1: when it compiles

PR #371, opened 2023-04-20
Delete lucuma.odb.data.Timestamp

swalker2m:
The `Timestamp` Scala class started in `lucuma.odb.data.Timestamp` but was eventually copied to `lucuma.core`.  Once the ODB was updated to a matching version of core, I should have deleted its own `Timestamp` but I forgot to do so.  This PR rectifies that oversight.


rpiaggio (approved):
LGTM

PR #379, opened 2023-04-26
Updated to use the new itc api

cquiroz:
This compiles but I'm not sure how to test it


cquiroz:
Besides the new API, the new ITC server brings a much faster algorithm when a s/n@wavelength parameter is specified

flake.nix
```diff
@@ -0,0 +1,34 @@
...
+          overlays = [ typelevel-nix.overlay scala-cli-overlay ];
+        };
+      in {
+        devShell = pkgs.devshell.mkShell {
+          imports = [ typelevel-nix.typelevelShell ];
+          packages = [
+            pkgs.nodePackages.graphqurl
+            pkgs.docker
+            pkgs.docker-compose
+            pkgs.postgresql
```
tpolecat, line 24: Seems like you shouldn't need postgres.

tpolecat (approved):
Seems fine. 👍🏻

flake.nix
cquiroz: This is to get the `psql` client

PR #385, opened 2023-04-27
Add proposal attachments

toddburnside:
Because the proposal attachments are one per type per program, I opted not to give the table a serial id, but rather used a primary key of `program_id, attachment_type`. Critiques of this approach are welcome. Because of this, though, I was required to figure out a few things. See the comments in `ProposalAttachmentMapping.scala` and `ProposalAttachmentMetadataService.scala` for a specific example.


modules/service/src/main/scala/lucuma/odb/graphql/mapping/ProposalAttachmentMapping.scala
```diff
@@ -0,0 +1,28 @@
...
+import table.ProposalAttachmentTable
+
+trait ProposalAttachmentMapping[F[_]] extends ProposalAttachmentTable[F] with ProgramTable[F] {
+
+  lazy val ProposalAttachmentMapping =
+    ObjectMapping(
+      tpe = ProposalAttachmentType,
+      fieldMappings = List(
+        // SqlField("program_id", ProposalAttachmentTable.ProgramId, key = true, hidden = true),
+        SqlField("program_id", ProposalAttachmentTable.ProgramId, hidden = true),
```
toddburnside, line 18: The primary key in the table is `program_id` + `attachment_type`, but if I made both of those key fields here, in required my where clause in `ProposalAtttachmentMetadataService` to use both of these keys. I'm not sure if there are other consequences of not having `program_id` be a marked as a key here or which version is preferred.

modules/service/src/main/scala/lucuma/odb/service/ProposalAttachmentMetadataService.scala
```diff
@@ -0,0 +1,81 @@
...
+      programId: Program.Id,
+      SET: ProposalAttachmentPropertiesInput.Edit,
+      which: AppliedFragment
+    ): Option[AppliedFragment] = 
+      updates(SET).map {us =>
+        void"UPDATE t_proposal_attachment "  |+|
+        void"SET " |+| us.intercalate(void", ") |+| void" " |+|
+        sql"WHERE t_proposal_attachment.c_program_id = $program_id ".apply(programId) |+|
+        void"AND t_proposal_attachment.c_attachment_type IN (" |+| which |+| void") " |+|
+        // void"AND (t_proposal_attachment.c_attachment_type, t_proposal_attachment.c_program_id) IN (" |+| which |+| void") " |+|
```
toddburnside, line 77: If I mark `program_id` as a key in the `ProposalAttachmentMapping`, I would need to use this IN clause with both the attachment_type and program_id. I'm not sure why the order is attachment_type first.

swalker2m (commented):
This was a lot of work and it looks very clean to me.  I'm :+1: but I would like Rob's input on the primary key issue you raised.  It seems like the mapping should have the full primary key.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/ProposalAttachmentMapping.scala
tpolecat: Yeah you need both key columns for sure, otherwise Grackle may not be able to tell them apart if you're querying across multiple programs. If this is a hassle you can create a view that constructs the composite key.

modules/service/src/main/scala/lucuma/odb/service/ProposalAttachmentMetadataService.scala
toddburnside: The `which` AppliedFragment is created in `MutationMapping.scala` starting at line 150. Grackle somehow uses the filterPredicate to build the AppliedFragment using the keys defined in the table mapping.

PR #392, opened 2023-05-01
groups

tpolecat:
First of all, I'm sorry. This is both late and big.

This adds groups as follows:

- Every group has an ordered list of elements, which are observations and other groups.
- Groups have metadata that give it meaning from a scheduling point of view. This information can be specified and edited but we're not using it anywhere yet.
- Each program has an implied top-level group whose id is null. Any observation or group within the null group appears at the top level. This means an element's parent group can be null its index cannot be, as the top-level elements are ordered.
- Elements' parent group and index are modeled as fields, and you specify an element's location by editing these fields. If you specify a location that is already in use, conflicting elements will be shifted out of the way as needed. Similarly, if you move an element then any hole left behind will be closed.
- When creating or editing an element, if no group is specified then the top-level null group is assumed. If no index is specified then the element will be placed last in the destination group.

Follow-on work:

- [ ] add a `Group` class to core (?)
- [ ] finalize the API once @hugo-vrijswijk has had a chance to try it
- [ ] add access control checks
- [ ] add an event stream for group modification


build.sbt
```diff
@@ -59,7 +59,6 @@ lazy val sequence = project
   .settings(
     name := "lucuma-odb-sequence",
     libraryDependencies ++= Seq(
-      "edu.gemini"     %% "lucuma-itc-core"                 % lucumaItcVersion,
```
tpolecat, line 62: This seems to have disappeared.

modules/sequence/src/test/scala/lucuma/odb/sequence/TestItcClient.scala
```diff
@@ -44,6 +44,9 @@ object TestItcClient {
       ): F[SpectroscopyResult] =
         SpectroscopyResult(Version, result.some).pure[F]
 
+      def optimizedSpectroscopyGraph(input: lucuma.itc.client.OptimizedSpectroscopyGraphInput, useCache: Boolean): F[lucuma.itc.client.OptimizedSpectroscopyGraphResult] =
```
tpolecat, line 47: This method is needed now but luckily we're not calling it, so I just stubbed it out.

swalker2m (approved):
Wow this was a lot of work!  It looks very clean though keeping up with order is not trivial.  I'd like to merge this before the "todo" items are complete if possible to get the library updates.

PR #393, opened 2023-05-01
groups channel

tpolecat:
This adds a subscription for group modifications, triggered when you edit or move a group. It follows the same pattern as the other subscriptions.


swalker2m (approved):
Impeccable. :+1:

PR #397, opened 2023-05-02
migration test

tpolecat:
This adds a lame script that lets you do a sanity check for migrations. I don't know how to write shell scripts so if anyone wants to push more commits to make this better, please do.


swalker2m (approved):
Looks good.  Does it take a long time to fetch the dump?  I guess it will depend on how big the db gets.

PR #405, opened 2023-05-08
Update to the latest lucuma-itc client

cquiroz:


modules/service/src/main/scala/lucuma/odb/logic/Itc.scala
```diff
@@ -189,10 +182,8 @@ object Itc {
       ): F[EitherNel[Error, ResultSet]] =
         targets.traverse { case (tid, si) =>
           client.spectroscopy(si, useCache).map {
-            case SpectroscopyResult(_, None)                                 =>
-              NoResult(tid).leftNel
-            case SpectroscopyResult(_, Some(s @ IntegrationTime(_, _, _))) =>
-              Success(tid, si, s).rightNel
+            case IntegrationTimeResult(_, results) =>
+              Success(tid, si, results.head).rightNel
```
swalker2m, line 186: Do we need to take the max here instead of just the first item or is it always just one item?

swalker2m (approved):
Looks great. :+1:

modules/service/src/main/scala/lucuma/odb/logic/Itc.scala
cquiroz: In spectroscopy only one is returned

PR #408, opened 2023-05-09
remove event ids

tpolecat:
This deprecates event ids in the API, stubs them out in the mappings (all ids are now const zero) and removes them from the database and back-end channel processing.


swalker2m (approved):
Pre-approved. Looks good. :+1:

Issue #300, opened 2023-03-03
CI Check for Migration Version


swalker2m:
If two or more developers independently create a database migration with the same version number, it can work in their local environments but the second and subsequent merges will fail in production.  This is easy to do inadvertently and results in failure at runtime.  Ideally we could add a check to ensure that any new migrations have a version that is strictly greater than previous ones.

PR #417, opened 2023-05-10
Verify migrations in CI

armanbilge:
Closes https://github.com/gemini-hlsw/lucuma-odb/issues/300.

Uses the Scala.js action developed here: https://github.com/gemini-hlsw/migration-validator-action

On that repo I setup a bunch of branches testing the action in different scenarios. So fairly confident it works 😅 
https://github.com/gemini-hlsw/migration-validator-action/branches/all


tpolecat (approved):
I don't understand how any of this works, but let's give it a go :+1:

PR #431, opened 2023-05-15
lib updates and logging

tpolecat:
This updates lucuma-graphql-routes (which transitively updates Grackle and Skunk) and **gets rid of all stacktraces and stdio output in tests**, so they run clean now. It's a big diff but the changes are mostly mechanical.

Executive summary for **Grackle**:
- `Result` is no longer an alias for `IorNec`, but is instead an ADT with four constructors. Three correspond to the `Ior` cases and the fourth is for internal errors that should not be shown to the user.

Executive summary for **Skunk**:
- Twiddle syntax is being replaced with Scala 3 tuple syntax.

| Thing | Old Style | New Style | 
|------|----|----|
| codecs    | `int2 ~ bool ~ text` | `int2 *: bool *: text` |
| types     | `A ~ B ~ C`          | `(A, B, C)` |
| values    | `a ~ b ~ c`          | `(a, b, c)`  |
| arguments | `.execute(a ~ b)`    | `execute(a, b)` |


swalker2m (approved):
Excellent!  I'll miss the `~` but definitely not all the stack traces.  Thanks for getting everything up to date.  :+1:

PR #434, opened 2023-05-16
Limit sequence length

swalker2m:
Places a limit on sequence generation at 1,000 steps.  This is an attempt to prevent OOM errors for unexpectedly long sequences.  While this should be fine for GMOS Longslit, it is understood that IR instruments require lengthy sequences with short exposure times.  A follow-up PR will be needed with sequence streaming in the server and, perhaps, paging in the API.


tpolecat (approved):
👍🏻

PR #444, opened 2023-05-21
Planned time migration

swalker2m:
This PR contains the migration needed for adding planned time calculation to the sequence generation.  It is standalone but not particularly useful without the corresponding calculation.  I'm introducing it now though in order to make the size of the next PR somewhat less depressing.


modules/service/src/main/resources/db/migration/V0342__planned_time_updates.sql
```diff
@@ -0,0 +1,36 @@
+ALTER TABLE t_time_estimate
+       DROP c_category;
+
+DROP TABLE t_planned_time_category;
```
swalker2m, line 4: I didn't end up needing the planned time category enum.

tpolecat (approved):

PR #451, opened 2023-05-24
Notifiy observation subscription on timing windows change

rpiaggio:


rpiaggio:
> This looks good to me. @cquiroz recently pointed out that we need to do something similar for observing mode parameters. My question, maybe for @tpolecat, is whether this will result in getting _two_ edit events in the case that you change a timing window in this case along with some other observation property.

We recently discussed this with Rob, since timing windows are deleted and reinserted on each update. He pointed out that:

> If the same channel name is signaled multiple times with identical payload strings within the same transaction, only one instance of the notification event is delivered to listeners.

This is from Postgres NOTIFY doc: https://www.postgresql.org/docs/current/sql-notify.html

swalker2m (approved):
Excellent, thanks for the clarification on `NOTIFY` / `pg_notify`.

rpiaggio:
> if you quickly update e.g. offsets and timing windows, do you get a notification for each?

Define "quickly".  If you do both in the same transaction and the emitted payload is the same (ie: `observation.id` + `program.id` + `UPDATE`), then you should only get one notification.

BTW, I'm always sending `UPDATE` here. As mentioned, timing windows are always `DELETE`d and re`INSERT`ed, and I see no (easy) way of knowing if this corresponds to a new observation or an observation update. If it's a new observation with timing windows included, then there will be 2 notifications: one `INSERT` and one `UPDATE`. Unless anyone has a better way of handling this case, I think we can live with that.

PR #454, opened 2023-05-24
Add pre-signed URLs for S3 and other attachment changes

toddburnside:
- Adds endpoints for getting presigned S3 URLs for attachment files so that the files can be downloaded directly from Amazon
- Adds the filename to the end of the file path on S3 so that the links will download the file with that name (instead of just a UUID for the name).
- Change the information we store in the DB related to the s3 file keys. Instead of just storing the UUID we were using to generate the file key, we now store the full path (minus the "base path" needed by our Cloud Cube). This will allow us to change our file naming in the future if we need to, without having to worry about losing access to existing attachments.
- Provides an ordering for the attachments in graphQL.


swalker2m (approved):
This looks great to me.  It seems like these are good improvements.  After discussing it, I'm afraid I don't have any insightful feedback.  :+1:

PR #462, opened 2023-05-26
transaction reorg

### LINKED ISSUE #448: Internal error on CloneObservationInput
a query to add an obs to a target end with an internal error messagge

the query is:
```graphql

        mutation ($input: CloneObservationInput!){
          cloneObservation(input: $input) {
            newObservation 
        {
          id
          title
          subtitle
          status
          activeStatus
          visualizationTime
          posAngleConstraint 
        {
          mode
          angle 
        {
          microarcseconds
        }
      
        }
      
          plannedTime {
            execution 
        {
          microseconds
        }
      
          }
          targetEnvironment {
            asterism {
              id
            }
          }
          constraintSet 
        {
          cloudExtinction
          imageQuality
          skyBackground
          waterVapor
          elevationRange {
            airMass {
              min
              max
            }
            hourAngle {
              minHours
              maxHours
            }
          }
        }
      
          timingWindows 
        {
          inclusion
          startUtc
          end {
            ... on TimingWindowEndAt {
              atUtc
            }
            ... on TimingWindowEndAfter {
              after {
                milliseconds
              }         
              repeat {
                period {
                  milliseconds
                }
                times
              }
            }
          }
        }
      
          scienceRequirements {
            spectroscopy {
              wavelength 
        {
          picometers
        }
      
              resolution
              signalToNoise
              signalToNoiseAt 
        {
          picometers
        }
      
              wavelengthCoverage 
        {
          picometers
        }
      
              focalPlane
              focalPlaneAngle 
        {
          microarcseconds
        }
      
              capability
            }
          }
          observingMode 
        {
          gmosNorthLongSlit {
            initialGrating
            initialFilter
            initialFpu
            initialCentralWavelength 
        {
          picometers
        }
      
            grating
            filter
            fpu
            centralWavelength 
        {
          picometers
        }
      
            defaultXBin
            explicitXBin
            defaultYBin
            explicitYBin
            defaultAmpReadMode
            explicitAmpReadMode
            defaultAmpGain
            explicitAmpGain
            defaultRoi
            explicitRoi
            defaultWavelengthDithers 
        {
          picometers
        }
      
            explicitWavelengthDithers 
        {
          picometers
        }
      
            defaultSpatialOffsets 
        {
          microarcseconds
        }
      
            explicitSpatialOffsets 
        {
          microarcseconds
        }
      
          }
          gmosSouthLongSlit {
            initialGrating
            initialFilter
            initialFpu
            initialCentralWavelength 
        {
          picometers
        }
      
            grating
            filter
            fpu
            centralWavelength 
        {
          picometers
        }
      
            defaultXBin
            explicitXBin
            defaultYBin
            explicitYBin
            defaultAmpReadMode
            explicitAmpReadMode
            defaultAmpGain
            explicitAmpGain
            defaultRoi
            explicitRoi
            defaultWavelengthDithers 
        {
          picometers
        }
      
            explicitWavelengthDithers 
        {
          picometers
        }
      
            defaultSpatialOffsets 
        {
          microarcseconds
        }
      
            explicitSpatialOffsets 
        {
          microarcseconds
        }
      
          }
        }
      
        }
      
          }
        }
      
```
and variables:
```json
{
  "input": {
    "observationId": "o-129",
    "SET": {
      "targetEnvironment": {
        "asterism": [
          "t-4f0"
        ]
      }
    }
  }
}
```

tpolecat:
This fixes #448 as part of a larger cleanup-up, the goals of which are:

- Every interaction with the database should run in a single transaction, which means all services involved must be bound to the same `Session`, and the transaction must normally be handled externally (i.e., not in the service implementations themselves) in order to allow services to call one another freely without worrying about transaction boundaries. 
- All services should have easy access to other services, without requiring that they be constructed explicitly (which could lead again to multiple sessions).
- In order to support table auditing for the Chronicle, the database should always know who the current `User`. This requires setting a transaction-local variable when the transaction is initiated, which means all transactions should be initiated the same way. This will be implemented in a follow-up.

So the big change is to deal with services as a bundle. Rather than constructing individual service resources we have a single `Services` resource that constructs instances of individual services on demand and ensures they're all bound to the same `Session`. We use this interface _implicitly_, allowing uniform access to all services when it's in scope.

- Look at `Services.scala` to get an idea of what the deal is.

The practical effects on the code are:

- **Service factory methods** (formerly variations of `fromUserAndSession`) are all named `instantiate`, and most only take an _implicit_ `Services`. Because the `Services` is in implicit scope in the body of the factory method, service code has access to the session, user, and other services without requiring them as explicit constructor arguments and without risk that they're bound to different sessions.

- Most **Service methods** that return values in `F` now have a `using Transaction[F]` clause in order to prevent them being called without a transaction, and to reinforce the idea that the operations are now sub-transactional and can be composed freely. The attachment services need to manage transactions internally, so they demand a `NoTransaction[F]`.
 
- In **REST** routes the `Services` instance is constructed explicitly at each endpoint, after the `SsoClient` retrieves the user. This is a minor complication but it's isolated to the attachment endpoints. The test code for these endpoints constructs a service instance directly so there's a bit of hacking to allow construction both ways.

Other minor changes:

- Database services never need more than `Concurrent` (if they're using streams) or `MonadThrowCatch` (if they're not). So I weakened them when possible to make the constraint on `Services` less aggressive.
- No need to use `Sync.delay` to generate UUIDs, there's now a `UUIDGen[F]` typeclass provided by cats-effect.
- I moved the ITC and sequence generation methods into `QueryMapping`.


toddburnside (approved):
LGTM

modules/service/src/main/scala/lucuma/odb/service/GeneratorParamsService.scala
```diff
@@ -78,24 +78,25 @@ object GeneratorParamsService {
...
+  def instantiate[F[_]: Concurrent](using Services[F]): GeneratorParamsService[F] =
     new GeneratorParamsService[F] {
 
       import lucuma.odb.sequence.gmos
 
+      override def select(
+        programId: Program.Id,
+        which:     Observation.Id
+      )(using Transaction[F]): F[Option[EitherNel[MissingData, GeneratorParams]]] =
+        selectAll(programId, List(which)).map(_.get(which))
```
swalker2m, line 90: Was in necessary to move this down or is it just better esthetically?  Either way :+1: for sure, but am curious.

modules/service/src/main/scala/lucuma/odb/service/ProposalAttachmentFileService.scala
```diff
@@ -218,51 +220,43 @@ object ProposalAttachmentFileService extends AttachmentFileService {
...
-                .flatMap( _ =>
-                  for {
-                    uuid   <- Async[F].delay(UUID.randomUUID())
-                    path    = filePath(programId, uuid, fn.value)
-                    size   <- s3FileSvc.upload(path, data)
-                    result <- insertAttachmentInDB(user, programId, attachmentType, fn, description, size, path)
-                  } yield result
-                )
+              for {
+                _      <- services.transactionally {
```
swalker2m, line 249: I suppose this one is different (`using NoTransaction[F]` and then `services.transactionally`) to not hold a transaction open while it does a file upload?  What would happen in the theoretical case of another service method that is `using Transaction[F]` calling `insertAttachment`?  I guess it doesn't compile?

modules/service/src/main/scala/lucuma/odb/service/Services.scala
```diff
@@ -0,0 +1,202 @@
...
+import lucuma.itc.client.ItcClient
+import lucuma.odb.logic.Generator
+import lucuma.odb.logic.Itc
+import lucuma.odb.sequence.util.CommitHash
+import natchez.Trace
+import skunk.Session
+import skunk.Transaction
+
+/** Witnesses that there is no transaction in context. */
+type NoTransaction[F[_]] = NotGiven[Transaction[F]]
```
swalker2m, line 21: Hmmm, why is there an `fs2.compat.NotGiven` instead of just `scala.util.NotGiven`?

modules/service/src/main/scala/lucuma/odb/service/Services.scala
```diff
@@ -0,0 +1,202 @@
...
+  def smartGcalService: SmartGcalService[F]
+  
+  /** The `TargetService`. */
+  def targetService: TargetService[F]
+  
+  /** The `TimingWindowService`. */
+  def timingWindowService: TimingWindowService[F]
+
+  /** Construct an `Itc` service, given an `ItcClient`.*/
+  def itc(itcClient: ItcClient[F]): Itc[F]
```
swalker2m, line 110: Right now we just use the one client, `ItcClient` but I wonder if we'll need something to manage clients or if they should be included as generally available like services.

swalker2m (approved):
This is excellent.  It really seems to simplify transaction handling and service juggling.  :+1:

modules/service/src/main/scala/lucuma/odb/service/Services.scala
tpolecat: Ah auto-import picked the wrong thing. Will fix.

modules/service/src/main/scala/lucuma/odb/service/ProposalAttachmentFileService.scala
tpolecat: Correct. If your transaction status is part of the type it will prevent you from composing things the wrong way. It's a weak guarantee because nothing is forcing you to put those things on the types, but it seems like a reasonable pattern.

modules/service/src/main/scala/lucuma/odb/service/Services.scala
tpolecat: Yeah, I guess it depends on how it works on the back end. I think you only generally need a single http client in an app.

PR #470, opened 2023-05-30
Observation c_instrument column update

swalker2m:
`t_observation` has a `c_instrument` column, but it wasn't being updated when the observing mode changes.  This PR fixes that oversight. The `(c_observation_id, c_instrument)` combination is going to be used as a foreign key into the `t_observation` table in various places to ensure that we never mix instrument types.  For example, the `recordGmosNorthVisit` mutation should only work if the observation is in fact a GMOS North observation.

The two existing mode tables, `t_gmos_north_long_slit` and `t_gmos_south_long_slit`, had an fkey constraint on observation's `(c_observation_id, c_observing_mode_type)`.  That unfortunately isn't enough to prevent an inconsistency between the observing mode and the observation instrument.  Accordingly, I changed the constraint to include all three columns: `(c_observation_id, c_instrument, c_observing_mode_type)`.

I added an `instrument` field to the `Observation` GraphQL type as well.  This may be overkill because you can get it from the mode, but it made it easier to test and is a slight convenience.

### DIFF
--- a/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -4645,6 +4645,11 @@ type Observation {
   # The science configuration
   observingMode: ObservingMode
 
+  """
+  The instrument in use for this observation, if the observing mode is set.
+  """
+  instrument: Instrument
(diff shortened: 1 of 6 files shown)


tpolecat (approved):
:+1:

PR #475, opened 2023-05-31
Fix publishing?

swalker2m:
I think `NoPublishPlugin` is a little too aggressive:

```
object NoPublishPlugin extends AutoPlugin {

  override lazy val projectSettings = Seq(
    skip in publish := true,
    publish := (()),
    publishLocal := (()),
    publishArtifact := false,
    publishTo := None
  )
}
```

I'd like to try dialing it back a bit and see if we can accomplish a published `lucuma-odb-schema` and a non-published app that actually has all of its dependencies.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -64,9 +64,10 @@ lazy val schema = project
 lazy val sequence = project
   .in(file("modules/sequence"))
   .dependsOn(schema % "compile->test")
-  .enablePlugins(NoPublishPlugin)
   .settings(
     name := "lucuma-odb-sequence",
+    publish / skip := true,
+    publishTo := None,
     libraryDependencies ++= Seq(
       "edu.gemini"     %% "lucuma-itc-client"               % lucumaItcVersion,
       "edu.gemini"     %% "lucuma-itc-testkit"              % lucumaItcVersion          % Test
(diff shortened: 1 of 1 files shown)


tpolecat (approved):
Looks fine, let's merge. 

@armanbilge can you have a look and maybe suggest improvements for a follow-on? Is this something we could fix in our plugin?

armanbilge:
> I think `NoPublishPlugin` is a little too aggressive:

Sorry, would you mind expanding on this? In what way is it too aggressive, is there an sbt task that is not working? We might definitely want to tweak the `NoPublishPlugin` if it's not working with certain tasks.

swalker2m:
> Sorry, would you mind expanding on this? In what way is it too aggressive, is there an sbt task that is not working? We might definitely want to tweak the NoPublishPlugin if it's not working with certain tasks.

There are four modules in the build: `schema`, `sequence`, `smartgcal`, and `service`.  What I'd like to accomplish is to:

1. Publish `lucuma-odb-schema` in all of its glory
2. Not publish the other 3 to any public repository at any rate
3. Have the ODB app that is package include its dependencies as it did before.

Currently the `sequence` and `smartgcal` dependencies go missing for some reason.

tpolecat:
The issue is that sbt-native-packager doesn't include modules if they're not published, even if they're in the same project. It's dumb and we don't know how to fix it or work around it.

https://github.com/sbt/sbt-native-packager/issues/1221

armanbilge:
Aha, thank you both! Let me dig into this.

PR #476, opened 2023-05-31
Initial Chronicle table auditing

tpolecat:
Ok this adds initial support for automatic table auditing for Chronicle. It's built out enough to test
but nothing is set in stone yet. **Nothing here is user-facing.** The idea is as follows:

- All transactions are associated with a user id via a transaction-local variable `lucuma.user`.
- Changes to tables are automatically recorded in companion tables (just two examples for now):

    | Data Table | Chronicle Table |
    |--|--|
   | `t_program` | `t_chron_program_update` |
   |`t_asterism_target`  | `t_chron_asterism_target_update` |

- Each chronicle table contains the following columns:

    | Column | Nullable? | Meaning |
    |--|--:|--|
    |  `c_chron_id` | `not null` | Monotonically increasing unique id. |            
    |  `c_timestamp` | `not null` | Transaction time. |
    |  `c_transaction_id` | `not null` | Transaction id. |
    |  `c_user` | `null` | Current user. If `null` then the update happened outside of `Services`. |                
    |  `c_operation` | `not null` | Postgres operation, one of `INSERT`, `UPDATE`, `DELETE`, `TRUNCATE`. |           
    |  `c_program_id` | `not null` | Associated program id. May need to be nullable. |          

- Note that the transaction time and id will be the same for all updates that happen on the same transaction.

- For each underlying "real" data table (i.e., tables containing model data, not just associations) the corresponding Chronicle table contains two columns for each data column: one indicating that the column was updated and the other containing the new value.

    | Data Column | Chronicle Column | Meaning |
    |--|--|--|
    | `c_foo   sometype` | `c_mod_foo   bool not null` | Was `c_foo` modified? |
    |         | `c_new_foo   sometype null` | What's the new value? `null` if `c_mod_foo` is false. |

- For join tables like `t_asterism_target` that simply represent the presence or absence of an association there is no real "update" semantics; it's only insert and delete. So the Chronicle table contains only key columns, with the same names as in the underlying table, rather than `mod` and `new` columns.

- There will likely be situations for other tables where neither of these strategies works and we'll need to come up with a solution.

- Note that there is **NO API** for accessing any of this for now. The only way to access Chronicle information is directly
  via SQL.


modules/service/src/main/resources/db/migration/V0361__chron_program_updates.sql
```diff
@@ -0,0 +1,144 @@
...
+    c_mod_pts_uncharged = chron.c_mod_pts_uncharged OR mod_pts_uncharged,
+    c_mod_pts_execution = chron.c_mod_pts_execution OR mod_pts_execution,
+    c_new_program_id    = CASE WHEN chron.c_mod_program_id    OR mod_program_id    THEN NEW.c_program_id    END,
+    c_new_existence     = CASE WHEN chron.c_mod_existence     OR mod_existence     THEN NEW.c_existence     END,
+    c_new_pi_user_id    = CASE WHEN chron.c_mod_pi_user_id    OR mod_pi_user_id    THEN NEW.c_pi_user_id    END,
+    c_new_pi_user_type  = CASE WHEN chron.c_mod_pi_user_type  OR mod_pi_user_type  THEN NEW.c_pi_user_type  END,
+    c_new_name          = CASE WHEN chron.c_mod_name          OR mod_name          THEN NEW.c_name          END,
+    c_new_pts_pi        = CASE WHEN chron.c_mod_pts_pi        OR mod_pts_pi        THEN NEW.c_pts_pi        END,
+    c_new_pts_uncharged = CASE WHEN chron.c_mod_pts_uncharged OR mod_pts_uncharged THEN NEW.c_pts_uncharged END,
+    c_new_pts_execution = CASE WHEN chron.c_mod_pts_execution OR mod_pts_execution THEN NEW.c_pts_execution END;
```
swalker2m, line 130: This is likely going to change soon, which raises the point that the `chron` tables will also have to be migrated whenever the associated table itself is migrated.

swalker2m (approved):
The code itself is impeccable and easy to follow.  I remain concerned about the complexity of adding and maintaining, essentially, a time dimension in an already sprawling data model. The program table is straightforward but other stuff (say observations) may not be easy.  I guess I don't have any other ideas though :-/  Maybe at least hold off on adding more until we get further along?

modules/service/src/main/resources/db/migration/V0361__chron_program_updates.sql
tpolecat: Yes, the Chronicle tables will need to be updated as the model changes. I don't have a sense for how hard that will be.

tpolecat:
So, yes, it's unclear whether this is the right way to do it, but at least we can have it in place and see how it does with the handful of things it's tracking. Maybe we can figure out how to simplify ... a lot of this will come down to what the actual Chronicle app wants to see.

PR #477, opened 2023-05-31
Cross publish schema codecs

rpiaggio:


swalker2m (approved):
Thanks, sorry about that I never think of JS.  :+1:

PR #478, opened 2023-05-31
Add validation for attachment file extensions

toddburnside:


modules/service/src/main/scala/lucuma/odb/service/AttachmentFileService.scala
```diff
@@ -39,8 +39,26 @@ trait AttachmentFileService {
...
+    if (isOK) MonadThrow[F].unit
+    else {
+      val msg = allowedExtensions match {
+        case head :: Nil => s"Must be a ${head.value.toUpperCase} file."
+        case _           =>
+          val valids = allowedExtensions.map(_.value.toUpperCase).sorted.mkString(", ")
+          s"Must be one of: $valids"
+      }
+      MonadThrow[F].raiseError(InvalidRequest(s"Invalid file. $msg"))
+    }
```
tpolecat, line 61: Elsewhere we're using structured return types rather than exceptions to handle user errors, so at some point it might be good to refactor the attachments code to work the same way. This is fine for now though.

tpolecat (approved):
Looks good!

modules/service/src/main/scala/lucuma/odb/service/AttachmentFileService.scala
toddburnside: OK. I guess I could make a structured type similar to `grackle.Result` that I could map to the http responses I need. I'll look into that sometime soon.

PR #480, opened 2023-06-01
Record Visit

swalker2m:
This is the first of a series of updates to support the Observe / ODB API.  Every time an observation is at least partially observed requires that Observe start by recording a visit, which consists of the instrument static config.  In response, it gets back a visit ID that will be needed for future interactions.

Notes:

1. Right now I've left this API call open for convenience but it will eventually need to be restricted to services.  @tpolecat this may be a case for serving a distinct API based on the user.
2. I needed to update to Postgres 15 for the `UNIQUE NULLS NOT DISTINCT` feature.


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -1004,19 +1019,30 @@ type GmosNorthStatic {
...
+"""
+GMOS North static configuration input parameters
+"""
 input GmosNorthStaticInput {
-  # GMOS North Detector option
-  detector: GmosNorthDetector! = HAMAMATSU
+  """
+  GMOS North Stage Mode (default to FOLLOW_XY)
+  """
+  stageMode: GmosNorthStageMode
```
swalker2m, line 1029: I wanted to make this:

```
stageMode: GmosNorthStageMode = FOLLOW_XY
```

but Grackle can't seem to handle supplying the default value so I settled for supplying it in code and documenting it here.  :-/

modules/service/src/main/scala/lucuma/odb/graphql/input/GmosNorthStaticInput.scala
```diff
@@ -0,0 +1,35 @@
...
+      case List(
+        GmosNorthStageModeBinding.Option("stageMode", rStageMode),
+        GmosNorthDetectorBinding.Option("detector", rDetector),
+        MosPreImagingBinding.Option("mosPreImaging", rPreImaging),
+        GmosNodAndShuffleInput.Binding.Option("nodAndShuffle", rNodAndShuffle)
+      ) =>
+        (rStageMode, rDetector, rPreImaging, rNodAndShuffle)
+          .parMapN { (stageMode, detector, preImage, ns) =>
+            GmosNorth(
+              stageMode.getOrElse(GmosNorthStageMode.FollowXy),
```
swalker2m, line 27: Here's where I do what I'd hoped Grackle would do for me with the default value in the schema itself.

swalker2m:
This is taken from the old Sangria / tmp-api days but I think it still applies.  It describes the Observe interactions in more detail: https://github.com/gemini-hlsw/lucuma-tmp-api/blob/master/docs/ObserveOdbInteraction.md#starting-a-visit

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -6092,7 +6170,7 @@ type TargetGroup {
 
   # Link back to program, temporary, to avoid some mapping issues.
   # Please don't use this in API calls.
-  program: Program! @deprecated 
+  program: Program! @deprecated
```
cquiroz, line 6173: I didn't know you could add annotations to the schema, nice

modules/service/src/main/scala/lucuma/odb/graphql/input/GmosNorthStaticInput.scala
cquiroz: I don't think this is exposed in explore, right?

modules/service/src/main/scala/lucuma/odb/graphql/input/GmosNorthStaticInput.scala
```diff
@@ -0,0 +1,35 @@
...
+              detector.getOrElse(GmosNorthDetector.Hamamatsu),
```
cquiroz, line 28: Do we still need to consider non `Hamamatsu` detectors?

swalker2m: Good question.  I think we eventually need to import old programs in some capacity.

cquiroz (approved):
Looks great, Are we already storing when images are being taken?

swalker2m:
> Are we already storing when images are being taken?

Not yet, I plan on rolling out the rest of the [Observe / ODB API](https://github.com/gemini-hlsw/lucuma-tmp-api/blob/master/docs/ObserveOdbInteraction.md#starting-a-visit) over the coming weeks.

swalker2m:
@rpiaggio @toddburnside @cquiroz  So, `dev` and `production` are on Postgres 15.3 so I could merge this now.  Unfortunately `staging` was set up in a time of Postgres 14 and accordingly is using 14.3.  So if I merge, when we promote to staging, we'd need to upgrade Postgres.  That apparently requires a bit of effort and even downtime.  So, is there any urgency to promote dev to staging in the near future?

cquiroz:
We probably need it on dev for a while to get the explore fixes ready anyway. I'd say we wait for that to promote to staging

PR #481, opened 2023-06-01
Fix Publishing? Take 2

swalker2m:
@armanbilge This allows the internal project dependencies to be published so that the app packaging will include them.  This is a temporary work around, I hope.  Ideally `smartgcal` and `sequence` would not be published.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -65,6 +65,7 @@ lazy val schema =
 
 lazy val sequence = project
   .in(file("modules/sequence"))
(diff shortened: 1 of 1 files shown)


swalker2m:
See #475 which didn't end up working out.  I first turned off Mima checks for those two bundles when it complained and then ran into issues with header generation.

swalker2m:
Waiting for #482 to deploy and if it works, I'll close this.

PR #482, opened 2023-06-01
Fix service app packaging

armanbilge:
Alternative to https://github.com/gemini-hlsw/lucuma-odb/pull/481 and https://github.com/gemini-hlsw/lucuma-odb/pull/475, does this do what you want?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -98,6 +98,7 @@ lazy val service = project
   .enablePlugins(NoPublishPlugin, JavaAppPackaging)
   .settings(
     name := "lucuma-odb-service",
+    projectDependencyArtifacts := (Compile / dependencyClasspathAsJars).value,
     libraryDependencies ++= Seq(
       "ch.qos.logback" %  "logback-classic"                    % logbackVersion,
       "com.monovore"   %% "decline-effect"                     % declineVersion,


build.sbt
```diff
@@ -98,6 +98,7 @@ lazy val service = project
   .enablePlugins(NoPublishPlugin, JavaAppPackaging)
   .settings(
     name := "lucuma-odb-service",
+    projectDependencyArtifacts := (Compile / dependencyClasspathAsJars).value,
```
armanbilge, line 101: If this works I'll look into upstreaming it.

build.sbt
```diff
@@ -98,6 +98,7 @@ lazy val service = project
   .enablePlugins(NoPublishPlugin, JavaAppPackaging)
```
armanbilge, line 98: I can't comment on it but the line above is suspicious to me.
```scala
.dependsOn(sequence % "compile->test", smartgcal % "compile->test")
```
This says that that `Compile` scope of the `service` module depends on the `Test` scope of `sequence` and `smartgcal`. Is that really right? I.e. you want all your test artifacts on your application classpath?

swalker2m: I think you're right.  Some additional test dependencies would need to be spelled out but I can update this in a future PR.

swalker2m (approved):
Thanks, looks great! Let's give it a try.  :+1:

PR #483, opened 2023-06-01
Build improvement

swalker2m:
An improvement suggested by @armanbilge in #482.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -65,13 +65,16 @@ lazy val schema =
 
 lazy val sequence = project
   .in(file("modules/sequence"))
-  .dependsOn(schema.jvm % "compile->test")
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -65,13 +65,16 @@ lazy val schema =
 
 lazy val sequence = project
   .in(file("modules/sequence"))
-  .dependsOn(schema.jvm % "compile->test")
+  .dependsOn(schema.jvm)
```
armanbilge, line 68: Ah, now I understand what this was for :) I think it was supposed to be `"compile;test->test"` i.e. "compile depends on compile, test depends on test", vs `"compile->test"` i.e. "compile depends on test".

But this works too.

armanbilge (approved):

PR #485, opened 2023-06-02
Future length parameter

swalker2m:
* Makes it possible to control (within limits) the size of the `possibleFuture` atoms of a sequence.
* Fixes a typo in a field name (was `hashMore`, is now `hasMore`)


toddburnside (approved):
LGTM

PR #488, opened 2023-06-06
Sequence event

swalker2m:
Implements the `addSequenceEvent` mutation and updates the visit ID type to `Gid` (instead of `Uid`).


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -5509,38 +5514,23 @@ type SequenceDigest {
...
 
 # Sequence-level events.  As commands are issued to execute a sequence, corresponding events are generated.
-type SequenceEvent implements ExecutionEvent {
-  # Identifies the observation to which the event refers
-  location: SequenceEventLocation!
-
-  # Sequence event data
-  payload: SequenceEventPayload!
-
+type SequenceEvent { # implements ExecutionEvent {
```
swalker2m, line 5517: Grackle (?) or me (?) struggle with interfaces.  This would not work when implementing the interface.

modules/service/src/main/scala/lucuma/odb/service/VisitService.scala
```diff
@@ -2,131 +2,121 @@
...
-           $observation_id,
-           $instrument
-       """.command
-
-   }
- }
\ No newline at end of file
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.odb.service
```
swalker2m, line 7: Sorry for the huge change set here.  The file was inexplicably indented one space to the right.

modules/service/src/main/scala/lucuma/odb/graphql/binding/SequeneCommandBinding.scala
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
```
toddburnside, line 1: Type in file name.

toddburnside (approved):
LGTM, but don't have any constructive input on the `implements ExecutionEvent` issue.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
swalker2m: They aren't actually implemented yet. I will have to remove it from those too unless Rob has any ideas.

modules/sequence/src/test/scala/lucuma/odb/sequence/TestItcClient.scala
```diff
@@ -47,27 +49,29 @@ object TestItcClient {
...
-          graphResult._1,
-          graphResult._2,
-          SignalToNoise.Min,
-          None
-        ).pure[F]
+      override def optimizedSpectroscopyGraph(
+        input:    OptimizedSpectroscopyGraphInput,
+        useCache: Boolean = true
+      ): F[OptimizedSpectroscopyGraphResult] =
+        ???
```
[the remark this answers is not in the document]
swalker2m: This is just a stubbed out ITC client implementation for testing and the ODB doesn't need those methods.

PR #491, opened 2023-06-06
ObsAttachment assignment to Observations

toddburnside:
~~There is one outstanding issue regarding the graphql query for `obsAttachments.observations(includeDeleted: true)`. I'll make a comment about it in `OdbSchema.graphql`.~~

I eliminated the above problem by doing as Rob suggested and not returning the observations within the obsAttachments. Instead, I am using the cached observations in explore to build a map from attachments to observations.


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -2974,6 +2977,18 @@ type ObsAttachment {
...
+  observations(
+    # Set to true to include deleted values
+    includeDeleted: Boolean! = false
+
+    # Starts the result set at (or after if not existent) the given observation id.
+    OFFSET: ObservationId
+
+    # Limits the result to at most this number of matches (but never more than 1000).
+    LIMIT: NonNegInt
+  ): ObservationSelectResult!
```
toddburnside, line 2991: I took this directly from `type Program`, and copied the elaborator from `ProgramMapping` to `ObsAttachmentMapping`. However, there is a problem here that isn't present in `Program`. This works fine:
`observations(includeDeleted: false)`
but if `includeDeleted` is `true`, `OFFSET` also needs to be included or there is an `AssertionError`. 
I can't figure out what is different in the code. ☹️ 

There is a test where one line can be uncommented to produce the AssertionError. I'll make a comment there.

modules/service/src/test/scala/lucuma/odb/graphql/attachments/obsAttachmentsAssignments.scala
```diff
@@ -0,0 +1,342 @@
...
+      aid1 <- insertAttachment(pi, pid, file1).toAttachmentId
+      aid2 <- insertAttachment(pi, pid, file2).toAttachmentId
+      oid1 <- createObservation(pi, pid, (aid1, file1), (aid2, file2))
+      oid2 <- createObservation(pi, pid, (aid2, file2))
+      _    <- assertObservation(pi, pid, oid1, (aid1, file1), (aid2, file2))
+      _    <- assertObservation(pi, pid, oid2, (aid2, file2))
+      _    <- assertAttachmentsWithObs(pi, pid, (aid1, file1, List(oid1)), (aid2, file2, List(oid1, oid2)))
+      _    <- deleteObservation(pi, pid, oid1)
+      _    <- assertAttachmentsWithObs(pi, pid, (aid1, file1, List.empty), (aid2, file2, List(oid2)))
+      // _    <- assertAttachmentsWithObs(pi, pid, true, (aid1, file1, List(oid1)), (aid2, file2, List(oid1)))
```
toddburnside, line 339: Uncommenting this line will cause the AssertionError noted above

modules/service/src/main/scala/lucuma/odb/graphql/mapping/ObsAttachmentMapping.scala
```diff
@@ -22,7 +40,34 @@ trait ObsAttachmentMapping[F[_]] extends ObsAttachmentTable[F] with ProgramTable
...
-        SqlObject("program", Join(ObsAttachmentTable.ProgramId, ProgramTable.Id))
+        SqlObject("program", Join(ObsAttachmentTable.ProgramId, ProgramTable.Id)),
+        SqlObject("observations")
       )
     )
+
+  lazy val ObsAttachmentElaborator: Map[TypeRef, PartialFunction[Select, Result[Query]]] = 
+    Map(
+      ObsAttachmentType -> {
+        case Select("observations", List(
```
toddburnside, line 51: This was taken from `ProgramMapping`. I can't see why it works there and has a problem here.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
swalker2m: I wonder if making `includeDeleted` optional with a default makes any difference.  (I realize that's not what is done in `Program`, but it seems like it should be.)

swalker2m (approved):
I looked through this again and will re-up my prior :+1:

PR #496, opened 2023-06-07
Record Step Bindings

swalker2m:
Before Observe begins to execute a step, it records the complete configuration in the database.  There is an input per instrument to accomplish this.  For example, `RecordGmosNorthStepInput` and `RecordGmosSouthStepInput`.  This PR rejiggers the `input`s in the schema a bit and implements all the bindings.

The mutation implementation itself is pending, but this is already so long that it seemed appropriate to make a separate PR. I  may make further changes as the bindings are used.


toddburnside (approved):
LGTM

PR #503, opened 2023-06-12
Record Step (Part 2)

swalker2m:
The second installment in a series for implementing `recordStepGmosNorth` and `recordStepGmosSouth`.   Before the seqexec executes a step, it records the complete configuration in the database via the `recordStep` mutation for the corresponding instrument.  In response it receives the recorded step id which is needed for subsequent events.

The first PR (#496) included the input bindings.  This PR brings the migrations and database services.  The next will add the mappings for the step result and test cases. This PR is not yet tested but does at least pass an initial sanity check.

(Also adds a guide state to the `Science` step config and includes the `SmartGcal` step config.)


toddburnside (approved):
LGTM

PR #509, opened 2023-06-13
Zipper decoder with either index or selected field

rpiaggio:
This PR modifies the `Zipper` decoder so that it only requires one of `index` or `selected`.


rpiaggio:
> The code looks good. I'm wondering about the motivation for it. If you have two items in the zipper that are equal according to `Eq` then the decoded zipper might not be the same as the one that was encoded unless you select the same index (which of course is the default here I see).

The motivation is not having to query the `selected` field (and all its subfields) in order to decode into the model. It becomes similar to what we do with different units in `Angle`, etc.

swalker2m (approved):
It will preserve the index and roundtrip (so long as the encoder includes the index), so this seems fine to me.  :+1:

PR #512, opened 2023-06-14
Record Step (Part 3)

swalker2m:
Part 3 of recording GMOS steps before execution.  (Part 1 #496, Part 2 #503).  Here we add the instrument configuration to the step record.  So this is just the GMOS dynamic configuration part.  Unfortunately (?) there is overlap with the same information that appears in generated steps.  As a result, we need `SwitchMapping`s to limit the application of the database mapping to the context of step records.

Next up will be the step configuration itself (science, gcal, etc.)


modules/service/src/main/scala/lucuma/odb/graphql/mapping/GmosDynamicMapping.scala
```diff
@@ -0,0 +1,84 @@
...
+              r <- cursor.field("ampReadMode", None).flatMap(_.as[GmosAmpReadMode])
+            } yield Json.obj(
+              "xBin"        -> x.asJson,
+              "yBin"        -> y.asJson,
+              "ampCount"    -> c.asJson,
+              "ampGain"     -> g.asJson,
+              "ampReadMode" -> r.asJson
+            ),
+            List("xBin", "yBin", "ampCount", "ampGain", "ampReadMode")
+        ),
```
swalker2m, line 53: I resorted to a `CursorFieldJson` out of ignorance.  I wasn't sure how to make a single `GmosCcdMode` mapping that is the same for both GMOS North and South when two different tables would hold the fields.  Is there another way?

modules/service/src/main/scala/lucuma/odb/graphql/table/GmosDynamicTables.scala
```diff
@@ -0,0 +1,67 @@
...
+
+    val Dtax: ColumnRef          = col("c_dtax",              gmos_dtax)
+    val Roi: ColumnRef           = col("c_roi",               gmos_roi)
+
+    // Grating (if any)
+    object Grating {
+      val Disperser: ColumnRef  = col("c_grating_disperser",  grating.embedded)
+      val Order: ColumnRef      = col("c_grating_order",      gmos_grating_order.embedded)
+      val Wavelength: ColumnRef = col("c_grating_wavelength", wavelength_pm.embedded)
+    }
```
swalker2m, line 40: I flipped this from `opt` to `embedded` when I got an incomprehensible error. Yet the `Fpu` definition below seems to need `opt` and not `embedded`.  It seems to work but a critical, informed look would be appropriate.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/GmosDynamicMapping.scala
tpolecat: Elsewhere we use `SqlObject` and then have a `SwitchMapping` that chooses the backing columns. Does that not work here?

modules/service/src/main/scala/lucuma/odb/graphql/table/GmosDynamicTables.scala
tpolecat: If it works I would just merge and maybe open an issue to review it.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/GmosDynamicMapping.scala
swalker2m: Yes, I had some mental block about it but that works.  I tried the same trick for the `GmosCustomMask` (which is currently implemented with a `CursorFieldJson`) and got a dreaded `AssertionError` from Grackle so I reverted that part and will open an issue.  At least until we execute sequences with a custom mask it shouldn't be an issue.

PR #516, opened 2023-06-16
Record Step (Part 5):  StepConfig Mapping

swalker2m:
Adds the `StepConfig` (e.g., gcal, offset etc. -- details not directly involving a particular instrument).  This is part of the step record sent by Observe to the ODB to keep up with the configuration that was actually executed.

A complication seen throughout is the need for `SwitchMapping`s everywhere.  This is because the database representation of the step overlaps with generated sequence steps which are computed and returned directly as JSON.  The `SwitchMapping` prevents finding the database mapping when processing the JSON by explicitly being defined only for a path tied to something in the database.


tpolecat (approved):
Looks good 👍🏻

PR #526, opened 2023-06-20
AddStepEvent

swalker2m:
Adds the `addStepEvent` mutation, which is used by Observe as a step transitions from configure through observe to end.  The step event must refer to a previously recorded step (via, `recordStep`).


tpolecat (approved):
Looks great. :+1:

PR #535, opened 2023-06-22
itc `EffectField`

swalker2m:
Adds the `itc` query to the `observation`, leaving the top-level `itc` query for now but marking it `@deprecated`.  It will be removed when there has been sufficient time for Explore to adjust.

This was not a priority issue per se, but I wanted to apply the `EffectField` to a simple case because it will be needed elsewhere.  The `sequence` query will be moved to the observation in a future PR.


tpolecat (approved):
This is great. Simpler than I expected. :+1:

PR #538, opened 2023-06-23
Update scala and apply linting rules

cquiroz:


swalker2m (approved):
Seems we had one or two unused imports.  Looks fine assuming it passes CI.  Thanks for doing this! :+1:

PR #549, opened 2023-06-27
wip: observed conditions

tpolecat:
This adds logging of observed conditions, for the Chronicle. There is currently no way to _read_ such values, other than than when initially returned following insert. This was about 5x more work than I expected.


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -216,6 +216,107 @@ input ConstraintSetInput {
...
+scalar TransactionId
+
+interface ChronicleEntry {
+  id: ChronicleId!
+  transactionId: TransactionId
+  user: User
+  timestamp: Timestamp!
+}
+
+type ConditionsEntry { # implements ChronicleEntry
```
tpolecat, line 236: I think we can't do this kind of mapping yet. Need to ask Miles.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -216,6 +216,89 @@ input ConstraintSetInput {
...
+  seeing:     AngleInput
+  extinction: Extinction
+  wavelength: WavelengthInput
+  azimuth:    AngleInput
+  elevation:  AngleInput
+}
+
+type ConditionsIntuition {
+  expectation: ConditionsExpectation
+  seeingTrend: SeeingTrend
```
swalker2m, line 273: Is `SeeingTrend` defined here?  I don't see it.

tpolecat: The three enumerated types here seemed pretty wishy-washy so I defined them in tables.

swalker2m: It doesn't need some kind of definition in the schema though?

tpolecat: No, the enum type is added to the schema in `OdbMapping` when it smooshes the schema source with `enums.schema`

```scala
          val schema: Schema =
            unsafeLoadSchema("OdbSchema.graphql") |+| enums.schema
```

tpolecat: In a just world you couldn't construct a schema with missing types, but Grackle doesn't mind.

tpolecat:
Ok this is up to date and ready for review I guess.

modules/service/src/main/scala/lucuma/odb/graphql/predicate/Predicates.scala
```diff
@@ -13,32 +13,60 @@ trait Predicates[F[_]] extends BaseMapping[F] {
...
+    // val observationEdit         = ObservationEditPredicates(Path.from(ObservationEditType))
+    // val program                 = ProgramPredicates(Path.from(ProgramType))
+    // val programEdit             = ProgramEditPredicates(Path.from(ProgramEditType))
+    // val proposalAttachment      = ProposalAttachmentPredicates(Path.from(ProposalAttachmentType))
+    // val proposalClass           = ProposalClassPredicates(Path.from(ProposalClassType))
+    // val sequenceEvent           = ExecutionEventPredicates(Path.from(SequenceEventType))
+    // val setAllocationResult     = SetAllocationResultPredicates(Path.from(SetAllocationResultType))
+    // val target                  = TargetPredicates(Path.from(TargetType))
+    // val targetEdit              = TargetEditPredicates(Path.from(TargetEditType))
+    // val targetGroup             = TargetGroupPredicates(Path.from(TargetGroupType))
```
swalker2m, line 41: Let's remove the commented lines?

tpolecat:
@swalker2m do you think `Extinction` should move into core? I guess it probably should. I have a habit of creating data types locally because it's easier.

swalker2m:
> do you think Extinction should move into core?

Yeah most everything else like this goes into core so it seems like this should too, ultimately.  I could imagine AGS making use of it. I don't see any problem with merging as is though and moving it later.

swalker2m (approved):
This looks great to me. :+1:

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
swalker2m: There is a `SqlInterfaceMapping`, would that work?  Eg: https://github.com/gemini-hlsw/gsp-graphql/blob/main/modules/sql/shared/src/test/scala/SqlInterfacesMapping.scala

tpolecat: The thing is, I don't need a discriminator for the insert result type. It's returning the concrete subtype so I don't know why extending the interface makes any difference to anything.

PR #553, opened 2023-06-28
ITC `useCache` flag

swalker2m:
Adds the `useCache` flag to the effect field version of the `itc` query.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/graphql/mapping/ObservationMapping.scala
+++ b/modules/service/src/main/scala/lucuma/odb/graphql/mapping/ObservationMapping.scala
@@ -127,13 +127,17 @@ trait ObservationMapping[F[_]]
           } yield (p, o, c)
         }
 
-      private def callItc(pid: Program.Id, oid: Observation.Id, useCache: Boolean): F[Result[(Observation.Id, Itc.ResultSet)]] =
+      private def callItc(
+        pid:      Program.Id,
+        oid:      Observation.Id,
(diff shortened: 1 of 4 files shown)


modules/service/src/main/scala/lucuma/odb/graphql/mapping/ObservationMapping.scala
```diff
@@ -98,36 +103,51 @@ trait ObservationMapping[F[_]]
...
-            .lookup(pid, oid, useCache = true)
+            .lookup(pid, oid, useCache)
             .map {
               case Left(errors)     => Result.failure(errors.map(_.format).intercalate(", "))
-              case Right(resultSet) => (oid, resultSet).success
+              case Right(resultSet) => ((oid, useCache), resultSet).success
             }
         }
 
       def runEffects(queries: List[(Query, Cursor)]): F[Result[List[(Query, Cursor)]]] =
```
swalker2m, line 144: To be honest, I'm not sure under what circumstances the queries would be "batched" and when "batched" whether each cursor's environment will be the same.  I'm assuming possibly not, so the results for two such queries with the same observation id might disagree about whether to use the cache.  I wish this could be simplified though.

tpolecat: As I understand it the environments will differ in general, and no batching or anything is done for you. But the handler can do whatever it wants (including batching).

PR #555, opened 2023-06-29
Add `sequence` field to Observation

swalker2m:
Moves the `sequence` field from `Query` to `Observation` (though it is temporarily left on `Query` and marked `deprecated` to ease the transition).  The same was done for the `itc` field in #535.  This is just a generalization of the same code to apply to both `itc` and `sequence`.


cquiroz (approved):
LGTM

PR #556, opened 2023-06-29
Don't hold a transaction open while calling the ITC

swalker2m:
The ITC lookup (and therefore also the sequence generation) were requiring a transaction while calling the ITC.  This PR rejiggers the implementation to only require a transaction while gathering the parameters.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/logic/Itc.scala
+++ b/modules/service/src/main/scala/lucuma/odb/logic/Itc.scala
@@ -6,7 +6,6 @@ package lucuma.odb.logic
 import cats.MonadThrow
 import cats.Order
 import cats.data.EitherNel
-import cats.data.EitherT
 import cats.data.NonEmptyList
 import cats.syntax.applicativeError.*
 import cats.syntax.either.*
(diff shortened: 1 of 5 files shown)


tpolecat (approved):
:+1: good catch

modules/service/src/main/scala/lucuma/odb/logic/Itc.scala
```diff
@@ -27,22 +26,22 @@ import lucuma.itc.client.ItcClient
...
+    oid: Observation.Id
+  )(using Transaction[F]): F[EitherNel[Itc.Error, GeneratorParams]]
 
-  def spectroscopy(
-    targets:  NonEmptyList[(Target.Id, SpectroscopyIntegrationTimeInput)],
+  def callService(
+    params:   GeneratorParams,
     useCache: Boolean
-  ): F[EitherNel[Itc.Error, Itc.ResultSet]]
+  )(using NoTransaction[F]): F[EitherNel[Itc.Error, Itc.ResultSet]]
```
tpolecat, line 44: :+1:

PR #559, opened 2023-06-30
Remove deprecated instrument field from sequence codecs

rpiaggio:
The boolean fields indicating the instrument in the execution config seem to be deprecated.

This PR removes them from the canonical codecs accordingly, relying instead on the `instrument` field to verify the correct decoder is being used.

### DIFF
--- a/modules/schema/src/main/scala/lucuma/odb/json/sequence.scala
+++ b/modules/schema/src/main/scala/lucuma/odb/json/sequence.scala
@@ -172,19 +172,20 @@ trait SequenceCodec {
     rootDecoder(Instrument.GmosSouth)(InstrumentExecutionConfig.GmosSouth.apply)
 
   private def rootEncoder[R, S: Encoder, D: Encoder](using Encoder[Offset], Encoder[TimeSpan])(
+    name:       String,
     instrument: Instrument
   )(root: R => ExecutionConfig[S, D]): Encoder[R] =
     Encoder.instance { (a: R) =>
       root(a).asJson.mapObject { obj =>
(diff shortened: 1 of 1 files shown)


swalker2m (changes requested):
These fields are actually required in order for Grackle to distinguish between GMOS North and GMOS South `ExecutionConfig`s, given that they otherwise have the same structure.  The `@deprecated` flag is a bit misleading but was added to discourage clients or human API users from actually including it in their queries.

swalker2m:
See: https://github.com/gemini-hlsw/lucuma-odb/pull/303

rpiaggio:
Ah ok, I didn't know this was a Grackle requirement.

Anyway, referencing these fields in the standard codecs forces us to query the bogus fields. I'm not proposing we remove them from the schema, just keeping them out of the `Decoder` at least.

swalker2m:
> I'm not proposing we remove them from the schema, just keeping them out of the Decoder at least.

I think that's fine, but they do need to be in the `Encoder`.

rpiaggio:
Ok, I added them back in the `Encoder`. Why are they necessary there? When is the `Encoder` used?

swalker2m:
> Why are they necessary there? When is the Encoder used?

Check out the test case in https://github.com/gemini-hlsw/lucuma-odb/pull/303.  It fails when you remove the field from the `Encoder`.  The `ExecuctionConfig` is computed as a JSON result and Grackle looks at the structure of the JSON to decide which type it corresponds to (as I understand it).

swalker2m (approved):
This looks good to me. :+1:

PR #563, opened 2023-07-03
ITC result table

swalker2m:
This PR would:

* Store ITC results in a table to avoid future calls to the ITC unless the input parameters change.
* Mark the `useCache` flag deprecated and ignore it.
* Poll the ITC every 5 minutes asking for version updates.  When detected, the ITC results table is truncated.


modules/service/src/main/resources/db/migration/V0369__itc_result.sql
```diff
@@ -0,0 +1,19 @@
...
+CREATE TABLE t_itc_result (
+  c_program_id      d_program_id     NOT NULL,
+  c_observation_id  d_observation_id NOT NULL,
+  c_target_id       d_target_id      NOT NULL,
+  c_hash            text             NOT NULL,
+  c_exposure_time   interval         NOT NULL,
+  c_exposure_count  int4             NOT NULL,
+  c_signal_to_noise numeric(10,3)    NOT NULL,
+  c_version         text             NOT NULL,
+  c_data            text             NULL,
```
swalker2m, line 14: Storing the ITC version in `c_version` and `c_data` per target is probably not correct.  If either change we should just truncate the table.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/ObservationMapping.scala
```diff
@@ -115,90 +112,67 @@ trait ObservationMapping[F[_]]
...
-        case Select("itc", List(
-          BooleanBinding.Option("useCache", rUseCache)
-        ), child) =>
-          rUseCache.map { useCache =>
-            Environment(
-              Env(UseCacheParam -> useCache.getOrElse(UseCacheDefault)),
-              Select("itc", Nil, child)
-            )
-          }
+        case Select("itc", List(BooleanBinding.Option("useCache", rUseCache)), child) =>
```
swalker2m, line 115: If `useCache` is still in the schema I assume I need to include it here?

modules/service/src/main/scala/lucuma/odb/graphql/mapping/QueryMapping.scala
```diff
@@ -62,45 +60,44 @@ trait QueryMapping[F[_]] extends Predicates[F] {
...
+      ).success
 
-      (for {
-        p <- EitherT(s.transactionally(generatorService.selectParams(pid, oid)))
-        r <- EitherT(generatorService.generate(oid, p, useCache, futureLimit))
-      } yield r).value.map(result => mapResult(result.merge))
+    services.use { s =>
+      s.generator(commitHash, itcClient, plannedTimeCalculator)
+       .generate(pid, oid, futureLimit)
+       .map(_.bimap(mapError, mapSuccess).merge)
```
swalker2m, line 100: All of these updates in `QueryMapping` are duplicated code but they should be removed when we remove the corresponding deprecated schema bits.

modules/service/src/main/scala/lucuma/odb/service/ItcService.scala
```diff
@@ -0,0 +1,360 @@
...
+  )
+
+  private val hex = HexFormat.of
+
+  private def hash(input: SpectroscopyIntegrationTimeInput): String =
+    hex.formatHex(
+      MessageDigest
+        .getInstance("MD5")
+        .digest(input.asJson.noSpaces.getBytes("UTF-8"))
+    )
```
swalker2m, line 148: Here is the hash.  Any better ideas? This is simple but it would be better not to first encode everything to JSON.  On the other hand, if the ITC server input changes the JSON changes so there's no way to overlook it when computing the hash this way.

tpolecat: This is the safest way to do it, so I think it's fine to start here. May never be an issue.

tpolecat (commented):
This looks good so far. We discussed polling the ITC server periodically to see if we need to truncate the cache, which seems reasonable.

modules/service/src/main/resources/db/migration/V0369__itc_result.sql
cquiroz: Maybe we need to store it as a global value rather than per target?

cquiroz (approved):
LGTM as far as my basic understanding of odb

modules/service/src/main/scala/lucuma/odb/service/ItcService.scala
```diff
@@ -0,0 +1,379 @@
...
+        _ <- session.use { s => s.transaction.use(_ => s.execute(Statements.UpdateItcVersion)(v.server.some, v.data)) }
+      } yield ()
+
+    Stream
+      .fixedDelay(pollPeriod)
+      .zip(Stream.repeatEval(pollOnce))
+      .compile
+      .drain
+      .start
+      .void
```
swalker2m, line 173: * Is `.start` the right way to do this?  I want it to continue as long as the process is running.
* Note `pollVersionsForever` is called on startup and there's no associated user or services.  A session created via `Session.single` is used for this.
* I suppose this will be done in all instances of the database server.  Is this an issue?

modules/service/src/main/scala/lucuma/odb/Main.scala
```diff
@@ -303,6 +304,7 @@ object FMain extends MainParams {
       _  <- Applicative[Resource[F, *]].unlessA(skipMigration.isRequested)(Resource.eval(migrateDatabase[F](c.database)))
       ep <- entryPointResource(c)
       ap <- ep.wsLiftR(routesResource(c)).map(_.map(_.orNotFound))
+      _  <- Resource.eval(ItcService.pollVersionsForever(c.itcClient, singleSession(c.database), c.itc.pollPeriod))
```
swalker2m, line 307: Better place for this?

modules/service/src/main/scala/lucuma/odb/service/ItcService.scala
cquiroz: I  think it is fine to have it in all instances as they need to clean their local caches

modules/service/src/main/scala/lucuma/odb/Main.scala
tpolecat: Seems fine.

modules/service/src/main/scala/lucuma/odb/service/ItcService.scala
tpolecat: I think `start` is fine here. If you needed a finalizer on shutdown you could use `Supervisor` but I don't think you need it.

modules/service/src/main/scala/lucuma/odb/service/ItcService.scala
```diff
@@ -164,9 +165,14 @@ object ItcService {
         _ <- session.use { s => s.transaction.use(_ => s.execute(Statements.UpdateItcVersion)(v.server.some, v.data)) }
       } yield ()
 
+    val pollOnceLogError: F[Unit] =
+      pollOnce.handleErrorWith { t =>
+        Logger[F].info(t)(s"Error while polling ITC versions: ${t.getMessage}")
```
tpolecat, line 170: 👍🏻 you need to ensure that it stays alive

tpolecat (approved):
This all looks good.

PR #573, opened 2023-07-10
test for #1

### LINKED ISSUE #1: query fails with "No match"
Generated query is wrong and yields no rows.

```graphql
mutation {
  createProgram(input:{ }) {
    users {
      user {
        id
      }
    }
  }
}
```

response is

```json
{
  "errors": [
    {
      "message": "No match"
    }
  ],
  "data": null
}
```

tpolecat:
This works now. Added a test but no changes were needed.

resolves #1


toddburnside (approved):
LGTM

PR #574, opened 2023-07-10
change comments to doc strings

### LINKED ISSUE #35: schema doc has turned into comments
We need to turn `# foo` into `"foo"` in the schema so users will be able to see it.

tpolecat:
resolves #35 

The issue was that non-doc comments were also turned into doc, resulting in multiple consecutive doc strings, which isn't legal (or at least doesn't parse). Searching for `"""[\n\s]+"""` revealed the gaps between consecutive doc strings and from there it was easy to fix.

### DIFF
--- a/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -51,13 +51,17 @@ type AddStepEventResult {
 
 }
 
-# Air mass range creation and edit parameters
+"""
+Air mass range creation and edit parameters
+"""
 input AirMassRangeInput {
   min: PosBigDecimal
   max: PosBigDecimal
 }
 
-# Time allocation
+"""
+Time allocation
+"""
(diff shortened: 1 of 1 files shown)


toddburnside (approved):
It's hard to really review that big of a diff, but LGTM

swalker2m:
> The issue was that non-doc comments were also turned into doc, resulting in multiple consecutive doc strings, which isn't legal (or at least doesn't parse)

What are multiple consecutive doc strings?

Does this incorporate the latest schema changes?  If not, maybe I can update the script that produced it to fix the error.

tpolecat:
@swalker2m the issue was that

```graphql
# Commented out field
# field: Type

# Non-commented field
field: Type
```

would become

```graphql
"""
Commented out field
field: Type
"""

"""Non-commented field"""
field: Type
```

And that's not legal GraphQL. There were only a few instances of this so I just fixed them by hand.

This is built against `main` so it should be up to date.

swalker2m:
> This is built against main so it should be up to date.

Thanks, I see.  :+1:

PR #590, opened 2023-07-17
sequence optimization

swalker2m:
Pre-generates the potentially unique configurations in a GMOS long slit science sequence and then simply repeats them.  This makes a marked difference in performance (~60x faster in the 6400 step sequence I'm testing with).

### DIFF
--- a/modules/sequence/src/main/scala/lucuma/odb/sequence/gmos/longslit/Science.scala
+++ b/modules/sequence/src/main/scala/lucuma/odb/sequence/gmos/longslit/Science.scala
@@ -34,6 +34,8 @@ import lucuma.core.optics.syntax.optional.*
 import lucuma.odb.sequence.data.ProtoStep
 import lucuma.odb.sequence.data.SciExposureTime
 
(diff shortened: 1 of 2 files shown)


cquiroz (approved):
Very neat, LGTM

modules/sequence/src/main/scala/lucuma/odb/sequence/gmos/longslit/Science.scala
```diff
@@ -54,6 +56,16 @@ sealed trait Science[D, G, F, U] extends SequenceState[D] {
     sampling:      PosDouble
   ): Stream[Pure, Science.Atom[D]] = {
 
+    @tailrec def gcd(a: BigInt, b: BigInt): BigInt = if (b === 0) a else gcd(b, a%b)
+    def lcm(as: BigInt*): BigInt = as.reduce { (a, b) => a/gcd(a,b)*b }
```
cquiroz, line 60: Can `as` ever be empty?

PR #594, opened 2023-07-19
Missing gmos filters

cquiroz:


cquiroz:
It also bugs me because we were supposed to not do this. filters, etc should be added in only one place right?

swalker2m:
> It also bugs me because we were supposed to not do this. filters, etc should be added in only one place right?

Yes, that's ultimately the plan for enums of this type where we have a listing but don't really refer to specific instances in code.  They will be defined in the database and added to the schema dynamically.  There are a handful of them already (like observation attachment type) but it was thought to be too disruptive to convert all the others before XT0.

cquiroz:
In that case they won't be in the schema then?

swalker2m:
> In that case they won't be in the schema then?

They are added to the schema dynamically:

https://github.com/gemini-hlsw/lucuma-odb/blob/main/modules/service/src/main/scala/lucuma/odb/graphql/OdbMapping.scala#L191

PR #606, opened 2023-07-26
Digest vs Config

swalker2m:
A number of improvements with the goal of speeding up (on average) everything that relates to acquisition and science sequence generation (e.g., planned time calculation).

1. Updates the schema to move all sequence related fields to `query` / `observation` / `execution`.  So `query` / `sequence` and `query` / `observation` / `sequence` are then `query` / `observation` / `execution` / `config`. In the future, the idea is that everything execution related will go to `execution` so you'll be able to get the datasets and events and visits from there, just as in the old Sangria API.
2. The `query` / `observation` / `execution` / `digest` field contains everything for which the entire sequence must be examined to compute.  It's a fold over the entire sequence.  Accordingly, the `atomCount` is moved to here.  The digest is cached in a `t_execution_digest` table so it needn't be recomputed constantly.  It also now includes the `SetupTime` as well as the `PlannedTime`.
3. The `config` field contains the `static` configuration, the `acquisition` sequence, and the `science` sequence with the existing `nextAtom`, `possibleFuture` and `hasMore` fields.  Computing these values requires only looking at the comparatively small `futureLimit` next steps and is therefore deemed fast.
4. Removes the deprecated `query` / `itc` but leaves `query` / `observation` / `itc`, albeit without the `useCache` flag.
5. In the implementation, changed the smart gcal lookup to be a `Pipe` from `atom` to `Either[error, expanded atom]` where a `Left` indicates a missing Smart Gcal mapping.  This is slightly more complicated but saves an entire trip through the sequence to validate Smart Gcal and thereby makes the digest 2x faster to compute.  This is appreciable for sequences of 1000s of steps.

The digest hash includes the ITC results, the observing mode parameters, and the commit hash.  Depending on how we update smart gcal definitions, if it isn't via PR, then that will need to be included.  Execution events will obviously have to be taken into account as well.


swalker2m:
Another change: matching Explore, "signal to noise at wavelength" is now required in order to do an ITC lookup and therefore in order to generate a sequence.  A missing parameter is reported if not defined. While it works without this parameter the ITC lookup is an order of magnitude slower if it is missing and we got the green light to require it in GPP meeting.

swalker2m:
Admittedly strange notation, but a picture may (or may not) help

 query: Query

* observation:         Observation!
  * execution:         Execution!
    * __digest:          ExecutionDigest!__
      * setupTime:        SetupTime!
      * acquistion:       SequenceDigest!
        * observeClass:   ObserveClass!
        * plannedTime:    PlannedTime!
        * offsets:        [Offset!]!
        * atomCount.      NonNegInt!
      * science:          SequenceDigest! 
        * *same as above*
    * __config:          ExecutionConfig!__ (e.g., GmosNorthExecutionConfig)
      * static:           GmosNorthStatic!
      * acquisition:      GmosNorthExecutionSequence
        * nextAtom:       GmosNorthAtom!
        * possibleFuture: [GmosNorthAtom!]!
        * hasMore:        Boolean!
      * science:          GmosNorthExecutionSequence!
        *  *same as above*
    * datasets (TBD)
    * events (TBD)
    * visits (TBD)

  * itc:               ItcResultSet

  * ~sequence:          SequenceGenerationResult~
    * ~programId~
    *  ~observationId~
    *  ~itcResult:       ItcResultSet!~
    *  ~executionConfig: ExecutionConfig!~
    *  ~scienceDigest:   SequenceDigest~

* ~itc: ItcResultSet!~
* ~sequence: SequenceGenerationResult~

cquiroz:
> Another change: matching Explore, "signal to noise at wavelength" is now required in order to do an ITC lookup and therefore in order to generate a sequence. A missing parameter is reported if not defined. While it works without this parameter the ITC lookup is an order of magnitude slower if it is missing and we got the green light to require it in a GPP meeting.

Explore won't query until signal to noise at is included as a parameter

tpolecat (approved):
This looks very good. :+1:

PR #617, opened 2023-07-31
Unmapped delete mutations

swalker2m:
These were removed from the implementation some time ago (in favor of `updateObservations` and `updateTargets` with the `existence` flag set appropriately) but apparently we failed to remove them from the schema.  At any rate, they aren't hooked up to anything so they will fail if called until we get rid of them.


tpolecat (approved):
👍🏻 👍🏻 👍🏻

PR #627, opened 2023-08-03
Program Planned Time

swalker2m:
A preliminary implementation of program-level planned time.  To do this, I had to make some assumptions about how groups should be counted that may turn out to have been wrong.  More importantly, it suffers from a flaw in that the input parameters to the observation planned time calculations are gathered in separate independent transactions.  I'd like to offer a followup PR that collects all the inputs and potential cached results in a single transaction so that the answer at least faithfully reflects the planned time as it was at the time of the request.

I'm offering the PR now though so that it appears in the API and can be used in Explore.  It works by

* Adding a simple in-memory group tree representation and updating `GroupService` to select and build it.
* Walking through this tree recursively calculating a min and max planned time from the bottom up.
* When the containing group or program requires more fully-defined observations than are present, we cannot perform the calculation so it is marked optional.


tpolecat (approved):
This looks great :+1:

toddburnside (approved):
LGTM

PR #630, opened 2023-08-04
Adds the target role (science, guide) to the target table

toddburnside:
This PR just adds the role to the target table and makes sure the current API only gets/creates/edits `science` targets.


modules/service/src/main/resources/db/migration/V0371__target_role.sql
```diff
@@ -0,0 +1,66 @@
+create type e_target_role as enum('science', 'guide');
+comment on type e_existence is 'The role of the target.';
+
+alter table t_target
+add column c_role e_target_role;
```
tpolecat, line 5: Seems like this should be `not null` with a clause to set all existing targets to `science`.

tpolecat (commented):
Can you add this field to `createTarget.FullTargetGraph` in the tests? This should ensure that the field is replicated correctly during cloning.

modules/service/src/main/resources/db/migration/V0371__target_role.sql
```diff
@@ -0,0 +1,66 @@
...
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER t_target_role_immutable_trigger
+BEFORE UPDATE on t_target
+FOR EACH ROW
+EXECUTE FUNCTION target_role_immutable();
+
+-- Need to recreate target view because of added field in target table.
+-- Can't use `update_table_with_views` because the saved view definition has
+-- expanded the `*` to the explicit columns, so the new column won't get added.
```
swalker2m, line 53: Could you expand on this, I'm not sure I understand.  I guess maybe I expected the view definition itself to be stored and not the result of executing the view definition.  It sounds like that's not the case?

swalker2m (commented):
This seems good.

modules/service/src/main/resources/db/migration/V0371__target_role.sql
toddburnside: Oh, good idea. If I knew you could do that, I had forgotten.

modules/service/src/main/resources/db/migration/V0371__target_role.sql
toddburnside: I expected the same, but apparently the stored definition has already expanded the `*`. The output of 

> select pg_get_viewdef('v-target`, true);

is
```sql
 SELECT t_target.c_program_id,
    t_target.c_target_id,
    t_target.c_name,
    t_target.c_existence,
    t_target.c_type,
    t_target.c_sid_ra,
    t_target.c_sid_dec,
    t_target.c_sid_epoch,
    t_target.c_sid_pm_ra,
    t_target.c_sid_pm_dec,
    t_target.c_sid_rv,
    t_target.c_sid_parallax,
    t_target.c_sid_catalog_name,
    t_target.c_sid_catalog_id,
    t_target.c_sid_catalog_object_type,
    t_target.c_nsid_des,
    t_target.c_nsid_key_type,
    t_target.c_nsid_key,
    t_target.c_source_profile,
    t_target.c_role,
        CASE
            WHEN t_target.c_type = 'sidereal'::e_target_type THEN t_target.c_target_id::character varying
            ELSE NULL::character varying
        END AS c_sidereal_id,
        CASE
            WHEN t_target.c_type = 'nonsidereal'::e_target_type THEN t_target.c_target_id::character varying
            ELSE NULL::character varying
        END AS c_nonsidereal_id,
        CASE
            WHEN t_target.c_sid_catalog_name IS NOT NULL THEN t_target.c_target_id::character varying
            ELSE NULL::character varying
        END AS c_sid_catalog_info_id,
        CASE
            WHEN t_target.c_sid_pm_ra IS NOT NULL THEN t_target.c_target_id::character varying
            ELSE NULL::character varying
        END AS c_sid_pm_id,
        CASE
            WHEN t_target.c_sid_parallax IS NOT NULL THEN t_target.c_target_id::character varying
            ELSE NULL::character varying
        END AS c_sid_parallax_id,
        CASE
            WHEN t_target.c_sid_rv IS NOT NULL THEN t_target.c_target_id::character varying
            ELSE NULL::character varying
        END AS c_sid_rv_id
   FROM t_target;
```

modules/service/src/main/resources/db/migration/V0371__target_role.sql
```diff
@@ -0,0 +1,66 @@
+create type e_target_role as enum('science', 'guide');
+comment on type e_existence is 'The role of the target.';
```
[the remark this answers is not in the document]
toddburnside: Good catch.

modules/service/src/main/resources/db/migration/V0371__target_role.sql
toddburnside: I still don't know how to do this. All examples I can find on the web show adding the columns as nullable, updating the data, then setting the not null constraint. Do you have an example for this?

tpolecat: Yep, I thought you could use a `USING` clause but that only works when modifying an existing column. What you have here is fine.

toddburnside:
> Can you add this field to `createTarget.FullTargetGraph` in the tests? This should ensure that the field is replicated correctly during cloning.

I can't really do that because the role is not in the graphql api. Right now, cloning (in TargetService) only works for science targets, but I will have to allow cloning of guide stars while cloning observations once the observation has guide stars.

I am adding some tests, but I will need to query the db directly in order to verify that the role is correct.

modules/service/src/main/resources/db/migration/V0371__target_role.sql
swalker2m: Perhaps then `update_table_with_views` isn't very useful.

swalker2m (approved):
I think this looks great as is.  Just a few questions for the future, but  :+1:

modules/service/src/main/scala/lucuma/odb/graphql/mapping/QueryMapping.scala
```diff
@@ -274,10 +275,11 @@ trait QueryMapping[F[_]] extends Predicates[F] {
         Select("target", Nil,
           Unique(
             Filter(
-              And(
+              and(List(
                 Predicates.target.id.eql(pid),
                 Predicates.target.program.isVisibleTo(user),
-              ),
```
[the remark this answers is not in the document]
toddburnside: My understanding was that the fact that guide stars are stored in the target table is purely an internal implementation thing. Since guide stars are very different in that they don't have an existence separate from the observation they belong to, I was planning on never exposing the target id for guide stars to the API.

modules/service/src/main/resources/db/migration/V0371__target_role.sql
toddburnside: `update_table_with_views` works well if you modify a column, just not if you add or delete a column

modules/service/src/main/scala/lucuma/odb/graphql/mapping/QueryMapping.scala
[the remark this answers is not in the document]
swalker2m: Does this mean that we'll need to add controls for skipping to the "next" highest ranked guide star?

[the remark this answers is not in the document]
swalker2m: Yes, I think the main thing is that by default when you ask for an observation's targets you probably just want the science targets so that's what you should get.  Somehow though we'll need to be able to see what the guide star / blind offset / etc is in the UI. The `target` and `targets` queries, properly restricted in the `WHERE` clause, is one simple (?) way of getting this but maybe there are better options.

I think it is fine to merge this PR now though and worry about the rest later.

PR #636, opened 2023-08-07
Add ccd mode and roi parameters to itc calls

cquiroz:


modules/sequence/src/main/scala/lucuma/odb/sequence/gmos/longslit/Config.scala
```diff
@@ -162,6 +164,15 @@ object Config {
...
       defaultWavelengthDithersNorth(this.grating)
 
+    val ccdMode: GmosCcdMode =
+      GmosCcdMode(
+        explicitXBin.getOrElse(defaultXBin),
+        explicitYBin.getOrElse(defaultYBin),
+        GmosAmpCount.Twelve,
+        explicitAmpGain.getOrElse(defaultAmpGain),
+        explicitAmpReadMode.getOrElse(defaultAmpReadMode)
+      )
```
swalker2m, line 174: I think you can use the `Config` trait methods that do the explicit or default thing?

```scala
GmosCcdMode(
  xBin,
  yBin,
  GmosAmpCount.Twelve,
  ampGain,
  ampReadMode
)
```

Also, I think we could just define this up in `Config` and not have to duplicate it in `GmosSouth`?

modules/service/src/main/scala/lucuma/odb/service/GeneratorParamsService.scala
```diff
@@ -155,14 +155,30 @@ object GeneratorParamsService {
...
             params.traverse { p =>
-              spectroscopyParams(p, InstrumentMode.GmosNorthSpectroscopy(g, f, GmosFpu.North.builtin(u)), λ)
+              spectroscopyParams(
+                p,
+                InstrumentMode.GmosNorthSpectroscopy(
+                  g,
+                  f,
+                  GmosFpu.North.builtin(u),
+                  gn.ccdMode.some,
+                  gn.explicitRoi.getOrElse(gn.roi).some),
```
swalker2m, line 165: Here too just `gn.roi.some`.

PR #638, opened 2023-08-08
Program-wide PlannedTimeRange Improvements

swalker2m:
Improvements to the program-wide planned time calculation.  Most importantly, it calculates the program planned time by gathering up all parameters and potentially cached results in a single transaction, then performing ITC calls and digest calculations outside of the transaction where necessary.  In addition to not cycling so many transactions, it provides an answer that at least corresponds to a consistent program-wide planned time at some instant.

Also:

* Moves the code from `ExecutionMapping` into its own service because it had grown rather large.
* Adds an `Md5Hash` `opaque type`.
* Switches the hash type in the database from a `String` to a `bytea`.


tpolecat (approved):
Man, that looks like a lot of work. Looks great. :+1:

PR #641, opened 2023-08-09
Program planned time range calculation update

swalker2m:
After playing with it a bit, Andy recommended a couple of changes to the program planned time range calculation.  The main issue was that the presence of any incomplete observation could easily and often make it impossible to estimate the planned time range for the remainder of observations.  This was addressed in two ways:

1. Limit the calculation to only `active` and `approved` (or higher) observations.  Any others are ignored for the purpose of estimating the remaining time.
2. When the `minRequired` value of a group (or indeed the program as a whole) is not set, we only consider the complete, active and approved observations.  Previously `minRequired` would default to #children, resulting in the presence of even one incomplete child invalidating the calculation for a group (or program) as a whole.  Now `minRequired` defaults to the number of valid children only, which may even be `0`.

I suspect that this algorithm will evolve over time as we figure out what is needed.  In particular I think not having the program planned time range at TAC time will be an issue. It won't be calculated because presumably none of the observations will yet have `approved` status?

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/logic/PlannedTimeRangeService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/logic/PlannedTimeRangeService.scala
@@ -11,6 +11,7 @@ import cats.syntax.apply.*
 import cats.syntax.flatMap.*
 import cats.syntax.foldable.*
 import cats.syntax.functor.*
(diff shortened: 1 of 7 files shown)


toddburnside (approved):
LGTM

Issue #645, opened 2023-08-10
Collect restricted operations


cquiroz:
We are going to need to restrict some operations to different roles
We could start writing them down here:

* delete program (only admin and your own)
* change observation state (only staff)

tpolecat:
I have started writing tests for these. Will update the checklist above as I go.

PR #660, opened 2023-08-21
tests for program edit access

tpolecat:
This adds some tests for program edits (which for now include deletion). At this point anyone who can see the program can do anything. This implements access control for NGO roles, where you have access to programs if your partner has granted them time.

There is a disconnect between the model's `Partner` enum and the partner table in the database. I will circle back and fix that next.

This is in reference to #645


swalker2m (approved):
Looks good. :+1:

PR #661, opened 2023-08-21
Move some config methods to lucuma-core

cquiroz:


rpiaggio (approved):
LGTM

swalker2m (commented):
The defaults and calculations are at least in some cases specific to the observing mode. Is that reflected in core?

PR #662, opened 2023-08-22
WIP: static metadata endpoint

tpolecat:
This adds an unauthenticated endpoint at `/enum-metadata` that executes a static query for enum metadata.

```graphql
query {
  partnerMeta {
    tag
    shortName
    longName
    active
  }
  filterTypeMeta {
    tag
    shortName
    longName
  }
  obsAttachmentTypeMeta {
    tag
    shortName
    longName
    fileExtensions {
      fileExtension
    }
  }
  proposalAttachmentTypeMeta {
    tag
    shortName
    longName
  }
}
```

It's a simple HTTP `GET` that returns a javascript source file.

```
$ curl -i http://localhost:4082/export/enumMetadata
HTTP/1.1 200 OK
Content-Type: application/javascript
Vary: Origin
Date: Tue, 22 Aug 2023 18:53:16 GMT
Content-Length: 2028

export const enumMetadata ='{"data":{"obsAttachmentTypeMeta":[{"tag":"FINDER","shortName":"Finder",...
```

The intent here is that clients like Explore fetch this metadata at startup and use it to populate `Enumerated` instances for associated model traits.

@rpiaggio @toddburnside @armanbilge @cquiroz - wdyty?

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/Main.scala
+++ b/modules/service/src/main/scala/lucuma/odb/Main.scala
@@ -44,6 +44,8 @@ import skunk.{Command => _, _}
 import software.amazon.awssdk.services.s3.presigner.S3Presigner
 
 import scala.concurrent.duration._
+import lucuma.odb.graphql.OdbMapping
+import lucuma.graphql.routes.GrackleGraphQLService
 
 object MainArgs {
   opaque type ResetDatabase = Boolean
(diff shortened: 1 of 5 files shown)


rpiaggio:
Thank you! This is quite close to what we need. Ideally, we would get a JS file that we can directly bundle with the app in a webpage. The JS code can just export a string that we deal with in the Scala side. I think that wrapping the current result in `export const enumMetadata = '...'` would work.

rpiaggio (approved):
LGTM

PR #668, opened 2023-08-24
treat json values as terminal

tpolecat:
This works around a behavior in Grackle by which JSON cursors can walk into non-JSON data, which is a cool feature but [so far] in Lucuma we just want to treat JSON results as opaque blobs and not worry about other existing mappings for types that might be encountered when zippering around. So this removes the need for the `SwitchMapping` in `GmosNorthStaticMapping` and in the work @toddburnside is doing with guide targets.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/graphql/mapping/GmosNorthStaticMapping.scala
+++ b/modules/service/src/main/scala/lucuma/odb/graphql/mapping/GmosNorthStaticMapping.scala
@@ -8,7 +8,7 @@
 
  trait GmosNorthStaticMapping[F[_]] extends GmosStaticTables[F] {
 
-   private lazy val mapping: ObjectMapping =
+   lazy val GmosNorthStaticMapping: ObjectMapping =
      ObjectMapping(
        tpe = GmosNorthStaticType,
        fieldMappings = List(
@@ -19,14 +19,4 @@
        )
      )
 
-   // The GmosNorthStatic type appears in the middle of a computed JSON result.
-   // In order to avoid finding this mapping in that case, we need to be
-   // explicit about when to use this mapping.
-   lazy val GmosNorthStaticMapping: TypeMapping =
-     SwitchMapping(
-       GmosNorthStaticType,
-       List(
-         GmosNorthVisitType / "static" -> mapping
-       )
-     )
  }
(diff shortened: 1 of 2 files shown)


toddburnside (approved):
LGTM. Thanks!

modules/service/src/main/scala/lucuma/odb/graphql/mapping/GmosNorthStaticMapping.scala
```diff
@@ -19,14 +19,4 @@
...
-   // In order to avoid finding this mapping in that case, we need to be
-   // explicit about when to use this mapping.
-   lazy val GmosNorthStaticMapping: TypeMapping =
-     SwitchMapping(
-       GmosNorthStaticType,
-       List(
-         GmosNorthVisitType / "static" -> mapping
-       )
-     )
  }
```
swalker2m, line 22: This looks good.  We should probably remove this hack from `GmosSouthStaticMapping` at the same time?

swalker2m (approved):
Excellent.  :+1:  Is there any reason it shouldn't work this way by default in Grackle?

modules/service/src/main/scala/lucuma/odb/graphql/mapping/GmosNorthStaticMapping.scala
tpolecat: Oh, yes. I didn't know there was one. I'll do that.

tpolecat:
> Excellent. 👍 Is there any reason it shouldn't work this way by default in Grackle?

Well in general data types can be mapped in different ways and you can hop around between mappings. There is a test that tests this behavior for CirceMapping, in fact.

What I want to do is change `CirceField` and a few related things to allow you to say `opaque=true` and get our desired behavior. This would allow us to get rid of the hack, at the expense of an extra constructor argument. I need to talk to Miles about it.

swalker2m:
There are more of these, among them are:

https://github.com/gemini-hlsw/lucuma-odb/blob/main/modules/service/src/main/scala/lucuma/odb/graphql/mapping/GmosDynamicMapping.scala#L31-L43

https://github.com/gemini-hlsw/lucuma-odb/blob/main/modules/service/src/main/scala/lucuma/odb/graphql/mapping/GmosFpuMapping.scala#L26-L38

https://github.com/gemini-hlsw/lucuma-odb/blob/main/modules/service/src/main/scala/lucuma/odb/graphql/mapping/GmosGratingConfigMapping.scala#L27-L39

https://github.com/gemini-hlsw/lucuma-odb/blob/main/modules/service/src/main/scala/lucuma/odb/graphql/mapping/StepConfigMapping.scala#L23-L35

but we can also replace them later as we discover them.

tpolecat:
This was rolled into in another PR.

PR #671, opened 2023-08-25
Add guide environment

toddburnside:
Guide targets are currently found "on demand" with no DB or caching.


modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
```diff
@@ -0,0 +1,575 @@
...
+            // "Can't" fail because all `Usable` candidates have a position angle
+            EitherT.fromEither(usable.head.posAngle.toRight(Error.GeneralError("No position angle for guide target.")))
+          ge              = GuideEnvironment(posAngle, List(GuideTarget(guideProbe, target)))
+        } yield ge).value
+    }
+
+  object Statements {
+    import ProgramService.Statements.andWhereUserAccess
+
+    def getAsterism(user: User, pid: Program.Id, oid: Observation.Id): AppliedFragment =
```
swalker2m, line 395: Maybe we should move this to the `AsterismService`?

swalker2m (commented):
This was a lot of work and it looks great!  The main observation I have is using the generated sequence instead of the digest but also some questions about the missing schema decoders.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -6518,6 +6518,57 @@ type CreateGroupResult {
...
 }
 
+  """
+  The guide star(s) and related information
+  """
+type GuideEnvironment {
+  """
+  The position angle
+  """
+  posAngle: Angle!
```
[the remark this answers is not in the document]
toddburnside: This position angle is determined by the guide star that is selected and at the given time. A different guide star could have a different position angle, so it seems to make sense here. Also, the position angle was included in the guide environment in an earlier discussion.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -6518,6 +6518,57 @@ type CreateGroupResult {
...
+
+  """
+  Sidereal tracking information, if this is a sidereal target
+  """
+  sidereal: Sidereal
+
+  """
+  Nonsidereal tracking information, if this is a nonsidereal target
+  """
+  nonsidereal: Nonsidereal
```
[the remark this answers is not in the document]
toddburnside: That was my original approach. However, a `Target` has 2 fields that didn't make sense here (IMO). 

1. The target id - since it has no independent existence and can't be edited independently, I didn't think it made sense to include the target id. 
2. The program. Since it can't be queried independent of the observation, having the program added here seemed redundant since the observation has it.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -7085,6 +7136,7 @@ type Observation {
   """
   itc: ItcResultSet!
 
+  # Temporary!!!`
```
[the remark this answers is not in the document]
toddburnside: Nope.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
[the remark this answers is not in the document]
swalker2m: It depends on the position angle constraint.  The PI can constrain the position angle to θ and then the AGS algorithm doesn't modify it.  Or the PI can set it to the parallactic angle and again the AGS algorithm doesn't change it.

modules/schema/src/main/scala/lucuma/odb/json/cataloginfo.scala
```diff
@@ -0,0 +1,27 @@
...
+      Encoder.instance { ci =>
+        Json.obj(
+          "name" -> ci.catalog.asJson,
+          "id" -> ci.id.asJson,
+          "objectType" -> ci.objectType.asJson
+        )
+      }
+  }
+
+  object query extends QueryEncoder
```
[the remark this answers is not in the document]
toddburnside: I just checked and `lucuma-schemas` has a decoder for this. I'll move this to `UniversalCodecs`, add the `Decoder` and update lucuma-schemas.

modules/schema/src/main/scala/lucuma/odb/json/declination.scala
```diff
@@ -0,0 +1,26 @@
...
+      Encoder.instance { dec =>
+        Json.obj(
+          "dms"             -> Declination.fromStringSignedDMS.reverseGet(dec).asJson,
+          "degrees"         -> BigDecimal(dec.toAngle.toDoubleDegrees).asJson,
+          "microarcseconds" -> dec.toAngle.toMicroarcseconds.asJson
+        )
+      }
+  }
+
+  object query extends QueryEncoder
```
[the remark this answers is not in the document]
toddburnside: It turns out `lucuma-schemas` has `Decoder`s for all or most of these. I guess they should all be moved here. I'll do that.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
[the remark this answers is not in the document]
swalker2m: I get the relation between guide star and position angle so maybe this is the correct place to keep up with it.  Even before a target is selected or a guide star is found though, an observation can have a position angle.  If we want to know the position angle of the observation we'll need to look in the `GuideEnvironment` but it may be unset?

modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
```diff
@@ -0,0 +1,575 @@
...
+              s"Unable to get tracking information for asterism for observation $oid."
+            )
+          )
+      }
+
+      def callGaia(
+        oid:     Observation.Id,
+        obsTime: Instant,
+        query:   ADQLQuery
+      ): F[Either[Error, NonEmptyList[GuideStarCandidate]]] = {
```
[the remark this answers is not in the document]
toddburnside: The actual call to Gaia isn't part of `lucuma-catalog`, just some helpers for doing that. I've been wondering if `lucuma-catalog` should have a client to encapsulate this. I think that's probably a good idea. I'll add it. I need to think about the time range, though, after a call with Andy on Friday.

modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
```diff
@@ -0,0 +1,575 @@
...
+              s.possibleFuture.flatMap(_.steps.offsets)
+          }
+          .flatMap(NonEmptyList.fromList)
+
+      def getOffsets(
+        pid: Program.Id,
+        oid: Observation.Id
+      ): F[Either[Error, (Site, Option[NonEmptyList[Offset]])]] =
+        generator(commitHash, itcClient, plannedTimeCalculator)
+          .generate(pid, oid, FutureLimit.Default)
```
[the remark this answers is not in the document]
toddburnside: In my ignorance, I didn't know I could get them out of the digest. I was just plagiarizing code from Explore. I'll switch to using the digest.

modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
```diff
@@ -0,0 +1,575 @@
...
+        scienceCoords: List[Coordinates],
+        positions:     NonEmptyList[AgsPosition],
+        candidates:    NonEmptyList[GuideStarCandidate]
+      ): Either[Error, NonEmptyList[AgsAnalysis]] =
+        obsInfo.wavelength.flatMap { wavelength =>
+          val analysis = Ags
+            .agsAnalysis(obsInfo.constraints,
+                         wavelength,
+                         baseCoords,
+                         scienceCoords,
```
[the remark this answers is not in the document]
toddburnside: I believe that is the case, although I don't fully understand the analysis. @cquiroz probably nows.

modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
toddburnside: I was torn about that, since the `AsterismService` doesn't care about being able to actually get an in-memory Target or Asterism. It's methods are only concerned with mutations, not queries. I am open to the idea of moving it, but am not convinced it (and the associated skunk decoder) belong there.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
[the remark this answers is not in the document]
toddburnside: I don't know. It seems to me that you can get all the information about the position angle you can get without knowing the time and guide star from the position angle constraint. If it is constrained to 0 degrees, then that will be the angle of the observation and it won't be changed here. However, if it is unconstrained, then won't the position angle of the observation depend on the time and the guide star selected and so can't be known until we select a guide star?

modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
```diff
@@ -0,0 +1,575 @@
...
+    given Encoder[GuideTarget] =
+      Encoder.instance { gt =>
+        Json.obj(
+          "probe"         -> gt.probe.asJson,
+          "name"          -> gt.target.name.asJson,
+          "sourceProfile" -> gt.target.sourceProfile.asJson,
+          "sidereal"      -> gt.target.asJson
+        )
+      }
+  }
```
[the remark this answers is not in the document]
toddburnside: Oh, yeah. I remember one reason I had for not making the `GuideTarget` more "publicly" available. Clients can simply use the `Target` model decoder for the GuideTarget because the Target model does not include the `Target.Id` or the `Program`.  Explore decodes the API result into a `(Target.Id, Target)`. For this, it could just decode into a `(GuideProbe, Target)`. I was just thinking of `GuideTarget` as a convenience class for this service.

modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
swalker2m: `AsterismService` doesn't do queries or get an in-memory representation because it didn't need to before now.  I suggested it because if I needed to query for asterisms in the future the first place I'd probably look is in the `AsterismService` and not in the `GuideEnvironmentService`.

toddburnside: Good point. You've convinced me. I'll move them.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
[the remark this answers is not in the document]
swalker2m: That makes sense.  We might consider (apart from this PR) nesting a `TargetDefinition` type or something with the common information.  It just seems like the bulk of it should be shared.  If we do, that change seems like it would be beyond the scope of this PR through.

modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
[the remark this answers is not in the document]
swalker2m: That sounds great, but can probably wait for a future task.  Just need to keep track of it in an issue or we'll forget!

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -9073,6 +9124,16 @@ type TargetEnvironment {
...
 
+  """
+  The guide star(s) and related information
+  """
+  guideEnvironment(
+    """
+    The time used for the guide star calculations.
+    """
+    observationTime: Timestamp!
+  ): GuideEnvironment
```
swalker2m, line 9135: At some point in a future PR we'll need, I think, an option to specify a duration (defaulting to the entire planned time presumably). The duration will impact the average parallactic angle computation.

modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
```diff
@@ -0,0 +1,469 @@
...
+        val firstOfYear = ldt.`with`(ChronoField.DAY_OF_YEAR, 1L).`with`(ChronoField.NANO_OF_DAY, 0)
+        val start =
+          if (ldt.getMonthValue() < Month.JULY.getValue())
+            firstOfYear
+          else firstOfYear.`with`(ChronoField.MONTH_OF_YEAR, Month.JULY.getValue())
+        val end   = start.plus(1, ChronoUnit.YEARS)
+
+        (tracking.at(start.toInstant(ZoneOffset.UTC)), tracking.at(end.toInstant(ZoneOffset.UTC)))
+          .mapN { (a, b) =>
+            // Make a query based on two coordinates of the base of an asterism over a year
```
swalker2m, line 192: What will this do?  The candidates have to be valid for the entire semester?  I'm not sure we want to include scheduler concerns in this query.  That said, this is probably fine since any guide star that moves out of range over the course of the semester is borderline anyway.

modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
```diff
@@ -0,0 +1,469 @@
...
+
+      extension [D](steps: NonEmptyList[Step[D]])
+        def offsets: List[Offset] = steps.collect { case Step(_, _, StepConfig.Science(offset, _), _, _, _) =>
+          offset
+        }
+
+      def offsetsFromExecConfig[S, D](
+        oid:        Observation.Id,
+        execConfig: ExecutionConfig[S, D]
+      ): Option[NonEmptyList[Offset]] =
```
swalker2m, line 258: Is this still used anywhere?

modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
```diff
@@ -0,0 +1,469 @@
...
+
+      def getOffsets(
+        pid: Program.Id,
+        oid: Observation.Id
+      ): F[Either[Error, Option[NonEmptyList[Offset]]]] =
+        generator(commitHash, itcClient, plannedTimeCalculator)
+          .digest(pid, oid)
+          .map {
+            _.leftMap(Error.GeneratorError(_))
+              .map { d => NonEmptyList.fromFoldable(d.science.offsets) }
```
swalker2m, line 274: Do we need a union with acquisition offsets?  Seems like the guide star has to work for both.

modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
```diff
@@ -0,0 +1,469 @@
...
+        posAngleConstraint: PosAngleConstraint,
+        offsets:            Option[NonEmptyList[Offset]],
+        tracking:           ObjectTracking,
+        obsTime:            Instant
+      ): Either[Error, NonEmptyList[AgsPosition]] = {
+        val angles     =
+          posAngleConstraint.anglesToTestAt(site, tracking, obsTime).map(_.sorted(Angle.AngleOrder))
+        val newOffsets = offsets match
+          case None          => NonEmptyList.of(Offset.Zero)
+          case Some(offsets) => offsets.prepend(Offset.Zero).distinct
```
swalker2m, line 289: If there are offsets, why do we need to include (0, 0)?

swalker2m (approved):
I had a few Qs but it looks great!  :+1:

modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
toddburnside: Nope. Good catch.

modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
toddburnside: Good question. @cquiroz? I just stole this from explore.

modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
toddburnside: I think I need to defer to @cquiroz again. More plagiarism...

cquiroz: I noted that when we get offsets we don't always get the position at zero, this `prepend` ensures that is always present

modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
toddburnside: This was going to be used for some caching, but that is probably not going to be done - at least not here. So, this will probably be reworked in a future PR.

modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
swalker2m: I think if there are offsets but none of them are at (0, 0), then there should be no datasets collected at (0, 0) in which case the guide star doesn't need to be valid there.

modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
cquiroz: That's correct, we need to include acq offsets, Needs to be fixed in explore

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
toddburnside: Right. There is going to need to be some significant tinkering here. I'm not sure how average parallactic is being handled now without it. 🤔

modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
toddburnside: OK. I'll combine with the acq offsets.

modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
toddburnside: So, should I remove the prepend? @cquiroz @swalker2m

cquiroz: I'm trying to remember why.

We are finding the region that intersect all offsets such that a GS is valid at all of them. But that would be valid even if there is no (0, 0) and the list of offsets is certainly non empty

I'd say let's remove it here and in explore 

Also we should consider moving some of this logic to `lucuma-ags` such that it is common for the odb and explore

toddburnside: OK. I'll remove it here and put in a PR for explore.
> Also we should consider moving some of this logic to `lucuma-ags` such that it is common for the odb and explore

Yes. I can examine that as part of creating a gaia client.

PR #678, opened 2023-08-30
Select target on asterisms based on brightness

cquiroz:
explore and the odb differ on what target to select when an asterism is present
On explore we use the exposure time for the brightest target while on odb we are selecting the longest exposure time

This PR makes odb use the same logic as explore as requested at
https://app.shortcut.com/lucuma/story/2504/exposure-time-and-number-of-exposures-for-observations-of-asterisms#activity-2514

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/ItcService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/ItcService.scala
@@ -22,6 +22,7 @@ import cats.syntax.option.*
 import cats.syntax.order.*
 import cats.syntax.traverse.*
 import clue.ResponseException
+import eu.timepit.refined.cats.*
 import eu.timepit.refined.types.numeric.PosInt
 import fs2.Stream
 import io.circe.Encoder
@@ -136,6 +137,11 @@ object ItcService {
         s"ITC returned errors: ${ps.intercalate(", ")}"
       }
     }
+
+    case object TargetMissmatch extends Error {
+      def format: String =
+        s"ITC provide conflicting results"
+    }
   }
 
   sealed trait TargetResult {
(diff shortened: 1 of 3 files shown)


modules/service/src/main/scala/lucuma/odb/service/ItcService.scala
```diff
@@ -295,10 +323,10 @@ object ItcService {
         itcInputs(params)
           .traverse { case (tid, input) => OptionT(selectSingleTarget(tid, input)) }
           .value
-          .map(_.map{ lst =>
-            val it: NonEmptyList[TargetResult] = lst.map(_._1)
+          .map(_.flatMap{ lst =>
```
swalker2m, line 326: If we really only need to ask about the brightest target, we could go further here.  No need to call ITC for each target?  Maybe no need for `AsterismResult` at all.  This might simplify things a good bit.

swalker2m: I like the warning idea.  A question is whether we can tell just based on the magnitudes that the asterism may be problematic or do we actually have to call the ITC and compare the exposure times for all the targets?

If we can get rid of `AsterismResult` it will ripple through the API but it would be a good simplification (and faster for multi-target asterisms) so I'm hopeful we can do it.

swalker2m (approved):
Pending simplifications aside, I think we can merge these changes now.  :+1:  If possible we can come back and get rid of `AsterismResult` in another PR.

modules/service/src/main/scala/lucuma/odb/service/ItcService.scala
cquiroz: Right, even if the calls are done in parallel they endup as an extra load on the ITC server

PR #682, opened 2023-08-31
Record Atom

swalker2m:
When I started thinking in earnest about using Observe events to influence sequence generation, I realized I needed more information.  In particular the steps recorded by Observe need to be associated with a particular atom.  This PR adds a `recordAtom` mutation and updates the sequence related tables a bit.

There is a more depressingly detailed description in this [sequence generation gist](https://gist.github.com/swalker2m/cf3643d4b1d680eb43709616784bbb2c).


tpolecat (approved):
This looks good. Is there any validation that steps are executed in order and that the indices fall in the correct range (i.e., you can't record 5 steps in a 3-step atom?)

swalker2m:
> Is there any validation that steps are executed in order and that the indices fall in the correct range

Hmm, I was planning on using the indices just to sort the steps but otherwise they could be anything.  If executed out of order then the atom won't match anything that was generated and won't count toward completion.  Now that you mention it though, the indices are not reliable for this purpose and the order should probably just be based on the event times.  I'll delete this field altogether.

PR #695, opened 2023-09-07
Use new method to distinguish the brightest target in an asterism

cquiroz:


swalker2m (approved):
Looks great to me. :+1:

PR #700, opened 2023-09-08
Use extension to detect itc errors

cquiroz:


toddburnside (approved):
LGTM

modules/service/src/main/scala/lucuma/odb/service/ItcService.scala
```diff
@@ -388,7 +389,7 @@ object ItcService {
         client.imaging(ii, useCache = false)
           .map(_.result)
           .recover {
-            case ResponseException(errors, _) if errors.exists(_.message.contains("target is too bright")) =>
+            case ResponseException(errors, _) if errors.exists(_.extensions.exists(_.exists(_ === ("errorCode" -> ErrorCode.SourceTooBright.asJson)))) =>
```
swalker2m, line 392: If there were other errors, besides the source-too-bright one, would we still want to use the default integration time?

swalker2m (approved):
Looks good.  :+1:  If we do this much, we might want some clue syntax or something to simplify.

modules/service/src/main/scala/lucuma/odb/service/ItcService.scala
cquiroz: Good question. I'd say only this one, if e.g. the config is unsupported you ddon't really want to default to a valid value

PR #703, opened 2023-09-11
Acquisition signal to noise

cquiroz:
With this PR we are storing the acquisition times given by itc alongside the science results

The schema has also changed to return both results and the results themselves now follow the same structure as a `Zipper`

### DIFF
--- a/modules/service/src/main/resources/db/migration/V0375__acquisition_sn_itc.sql
+++ b/modules/service/src/main/resources/db/migration/V0375__acquisition_sn_itc.sql
@@ -1,0 +1,22 @@
+-- Let's drop the itc values, they will be recalculated soon
+TRUNCATE TABLE t_itc_result;
+
+-- Rename columns to indicaate whether they are science or acquisition
+ALTER TABLE t_itc_result
+  RENAME COLUMN c_exposure_time TO c_sci_exposure_time;
+
+ALTER TABLE t_itc_result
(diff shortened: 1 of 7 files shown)


modules/service/src/main/resources/db/migration/V0375__acquisition_sn_itc.sql
```diff
@@ -0,0 +1,22 @@
...
+
+ALTER TABLE t_itc_result
+  RENAME COLUMN c_signal_to_noise TO c_sci_signal_to_noise;
+
+ALTER TABLE t_itc_result
+  RENAME COLUMN c_acquisition_time TO c_acq_exposure_time;
+
+-- Add missig acquisition columns
+ALTER TABLE t_itc_result
+  ADD COLUMN c_acq_exposure_count integer NOT NULL;
```
swalker2m, line 19: Is this value computed by the ITC?  It wouldn't be used for GMOS at any rate.

swalker2m (approved):
Looks great! :+1:  Does explore use this?  If so we'll need to update the schemas library version and publish.

modules/service/src/main/resources/db/migration/V0375__acquisition_sn_itc.sql
cquiroz: It is computed but not used. We had some discussions with the user that it maybe used but it requires some changes on ITC

cquiroz:
Yes, this will break explore

PR #707, opened 2023-09-12
Scala 3.3.1 + Library Updates

swalker2m:
Updates to Scala 3.3.1, which required changing some `var`s to ` val`s.  Omnibus library updates.  Removed a few other gentle code warnings (unused values, missing parameter types on public `val`s).


toddburnside (approved):
LGTM. Thanks for fixing the tests!

PR #712, opened 2023-09-13
t_step_record.c_completed

swalker2m:
Adds a nullable `c_completed` `timestamp` column to `t_step_record` which is set when Observe events warrant it.  At the moment it simply sets the timestamp when we get an `END_STEP` event.  That will turn out to be too simplistic but it works as a starting point.

When a step is determined to be completed, we now also delete the cached execution digest (if any).  When the sequence digest is computed it will need to take into account that some steps are completed.  This prepares the way by clearing the cache.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/SequenceService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/SequenceService.scala
@@ -55,6 +55,11 @@ trait SequenceService[F[_]] {
     observationId: Observation.Id
   )(using Transaction[F]): F[Map[Step.Id, (GmosSouth, StepConfig)]]
 
+  def setStepCompleted(
+    stepId: Step.Id,
+    time:   Timestamp
+  )(using Transaction[F]): F[Unit]
+
   def insertAtomRecord(
     visitId:      Visit.Id,
     instrument:   Instrument,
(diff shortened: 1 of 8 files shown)


swalker2m:
The last commit is a bit unrelated (sorry).  It updates the `SequenceService` logic to query the view `v_step_record` once instead of making individual queries to each of the step type-specific tables.

modules/service/src/main/scala/lucuma/odb/service/SequenceService.scala
```diff
@@ -568,8 +529,47 @@ object SequenceService {
...
 
-    val SelectStepConfigSmartGcalForObs: Query[Observation.Id, (Step.Id, StepConfig.SmartGcal)] =
-      selectStepConfigForObs("t_step_config_smart_gcal", StepConfigSmartGcalColumns, step_config_smart_gcal)
+    private val step_config: Codec[StepConfig] =
+      (
+        step_type               *:
+        step_config_gcal.opt    *:
+        step_config_science.opt *:
+        step_config_smart_gcal.opt
+      ).eimap { case stepType *: oGcal *: oScience *: oSmart *: EmptyTuple =>
```
tpolecat, line 538: It seems pointlessly confusing that for codecs `a *: b *: c` is the only way to write it, but for tuples (which end in `EmptyTuple`) you can also write it as `(a, b, c)`.

image

tpolecat (approved):
This looks great 👍🏻

modules/service/src/main/scala/lucuma/odb/service/SequenceService.scala
swalker2m: Thanks for the tip, that's much better.

PR #716, opened 2023-09-15
ExecutionDigest Input Hash

swalker2m:
When the acquisition ITC lookup was added, we forgot to add it to the `ExecutionDigest` computation input hash.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/logic/Generator.scala
+++ b/modules/service/src/main/scala/lucuma/odb/logic/Generator.scala
@@ -174,9 +174,10 @@ object Generator {
           md5.update(params.observingMode.hashBytes)
 
           // Integration Time
-          val ing = scienceIntegrationTime
-          md5.update(BigInt(ing.exposureTime.toMicroseconds).toByteArray.reverse.padTo(8, zero))
-          md5.update(BigInt(ing.exposures.value).toByteArray.reverse.padTo(4, zero))
+          List(acquisitionIntegrationTime, scienceIntegrationTime).foreach { ing =>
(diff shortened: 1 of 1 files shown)


modules/service/src/main/scala/lucuma/odb/logic/Generator.scala
```diff
@@ -174,9 +174,10 @@ object Generator {
           md5.update(params.observingMode.hashBytes)
 
           // Integration Time
-          val ing = scienceIntegrationTime
-          md5.update(BigInt(ing.exposureTime.toMicroseconds).toByteArray.reverse.padTo(8, zero))
-          md5.update(BigInt(ing.exposures.value).toByteArray.reverse.padTo(4, zero))
+          List(acquisitionIntegrationTime, scienceIntegrationTime).foreach { ing =>
+            md5.update(BigInt(ing.exposureTime.toMicroseconds).toByteArray.reverse.padTo(8, zero))
+            md5.update(BigInt(ing.exposures.value).toByteArray.reverse.padTo(4, zero))
```
cquiroz, line 179: should `signalToNoise` be included?

cquiroz (approved):
LGTM

modules/service/src/main/scala/lucuma/odb/logic/Generator.scala
swalker2m: It doesn't impact the sequence generation, at least not yet.  That is, the requested signal to noise is an input to get the integration time and count results (which are included in the hash) but does not otherwise directly impact the sequence.

PR #723, opened 2023-09-22
Filter Executed Atoms

swalker2m:
## Atom Filtering

The main thrust of this PR filters executed atoms from sequence generation.  Toward that end:

* `CompletedAtomMap` is promoted as a named (albeit `opaque`) type instead of the last PR’s raw `Map`.  Recall that it is a map from atom configuration to the number of executed atoms that match that configuration.  `CompletedAtomMap` contains an extension for matching against generated atoms.

* The match key in `CompletedAtomMap` is augmented with the sequence type so that acquisition and science atoms don’t count toward the completion of the opposite sequence type.

* Another complication is that the generated Atom IDs are based in part on the atom index of the generated sequence.  Since executed atoms are going to be filtered out, we need to zip the atom sequence with its index in order to keep from changing the atom ids of subsequent atoms.

## `HashBytes`

Along the way, I discovered that the “namespace” UUID used in creating atom and step IDs had a bug.  Namely, two identical `GeneratorParams` instances according to `Eq` would produce different namespace UUIDs.  This is presumably due to a difference in the way the objects are serialized.  Instead of writing the `GeneratorParams` to a `ByteArrayOutputStream` then, I added a `HashBytes` typeclass to explicitly convert implementing instances to an `Array[Byte]` that can be used to produce an `MD5` hash.

### DIFF
--- a/modules/sequence/src/main/scala/lucuma/odb/sequence/util/HashBytes.scala
+++ b/modules/sequence/src/main/scala/lucuma/odb/sequence/util/HashBytes.scala
@@ -11,6 +11,9 @@ import lucuma.core.util.Gid
 import java.nio.charset.StandardCharsets.UTF_8
 import java.security.MessageDigest
 
+/**
+ * Typeclass for producing an `Array[Byte]` that represents an object.
+ */
 trait HashBytes[A] {
 
   def hashBytes(a: A): Array[Byte]
--- a/modules/sequence/src/main/scala/lucuma/odb/sequence/util/HashBytes.scala
+++ b/modules/sequence/src/main/scala/lucuma/odb/sequence/util/HashBytes.scala
@@ -29,6 +29,12 @@ object HashBytes {
 
   def apply[A](using ev: HashBytes[A]): ev.type = ev
 
+  def forJsonEncoder[A: Encoder]: HashBytes[A] =
+    new HashBytes[A] {
+      def hashBytes(a: A): Array[Byte] =
+        Encoder[A].apply(a).spaces2.getBytes(UTF_8)
+    }
+
   private def toByteArray(b: BigInt, pad: Int): Array[Byte] =
     b.toByteArray.reverse.padTo(pad, 0.toByte)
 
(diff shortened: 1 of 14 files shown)


modules/sequence/src/main/scala/lucuma/odb/sequence/util/HashBytes.scala
```diff
@@ -0,0 +1,77 @@
...
+        HashBytes[PosLong].hashBytes(Gid[A].isoPosLong.get(a))
+      )
+  }
+
+  given HashBytes[TimeSpan] with {
+    def hashBytes(a: TimeSpan): Array[Byte] =
+      HashBytes[Long].hashBytes(a.toMicroseconds)
+  }
+
+  given given_HashBytes_Json[A: Encoder]: HashBytes[A] with {
```
tpolecat, line 73: I'm a little concerned that making this `given` could accidentally rope in things that we don't want to hash this way, since so many things have JSON encoders. Like, `Int` will be hashed with the explicit instance above, but `List[Int]` would fall back to the JSON hash. Maybe I'm missing something.

modules/service/src/main/scala/lucuma/odb/service/SequenceService.scala
```diff
@@ -434,46 +400,21 @@ object SequenceService {
...
-      ).query(atom_id *: int2_nonneg *: step_id)
+      ).query(atom_id *: int2_nonneg *: sequence_type *: step_id)
 
     def encodeColumns(prefix: Option[String], columns: List[String]): String =
       columns.map(c => s"${prefix.foldMap(_ + ".")}$c").intercalate(",\n")
 
-// TODO: I would like to write the insert step config logic once, parameterized
-// by the table name, columns, and the encoder.  I think I'm being thwarted by
-// scala.quoted / macro issues?
-//
```
tpolecat, line 454: Do you still need a workaround for this?

tpolecat (approved):
One question about `Encoder[A]` implying `HashBytes[A]` but otherwise :+1:

modules/service/src/main/scala/lucuma/odb/service/SequenceService.scala
swalker2m: I decided I could live with the way I left it.

PR #725, opened 2023-09-25
Reconcile modes

swalker2m:
When there are multiple targets in an asterism, the computed observing mode configuration may differ across targets.  This happens for GMOS Long Slit, for example, because the source profile is used in calculating the object size which in turn is a parameter in the optimal binning calculation.  The current implementation throws up its hands when it encounters differing binning settings across targets.  Upon consultation, we learned that the proper way to handle this is to choose the minimum binning value of any of the targets.  This PR makes that change, but any other differences in the observing mode configuration derived from the various targets would continue to be rejected.

Note there will be additional updates to the binning calculation in future PRs based on new requirements, as soon as I can work out what they mean.

### DIFF
--- a/modules/sequence/src/main/scala/lucuma/odb/sequence/gmos/longslit/Config.scala
+++ b/modules/sequence/src/main/scala/lucuma/odb/sequence/gmos/longslit/Config.scala
@@ -1,7 +1,8 @@
 // Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
 // For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
 
-package lucuma.odb.sequence.gmos.longslit
+package lucuma.odb.sequence
+package gmos.longslit
 
 import cats.Eq
 import cats.syntax.option.*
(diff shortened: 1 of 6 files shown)


modules/sequence/src/main/scala/lucuma/odb/sequence/gmos/longslit/Config.scala
```diff
@@ -206,6 +209,26 @@ object Config {
...
+
+        case Some(b@GmosNorth(_, _, _, _, _, _, _, _, _, _, _, _)) =>
+          if (a === b)
+            reconcile(a, modes.tail)
+          else {
+            val x  = a.xBin min b.xBin
+            val aʹ = a.copy(explicitXBin = none, defaultXBin = x)
+            val bʹ = b.copy(explicitXBin = none, defaultXBin = x)
+            if (aʹ === bʹ) reconcile(aʹ, modes.tail) else none
+          }
```
swalker2m, line 225: It's a bit grim to repeat this for north and south, but I couldn't figure out a good way of making it work for both.

modules/service/src/main/scala/lucuma/odb/service/GeneratorParamsService.scala
```diff
@@ -148,7 +147,7 @@ object GeneratorParamsService {
           m  <- observingModeServices.selectObservingMode(oms)
         } yield
           ps.groupBy(_.observationId).map { case (oid, oParams) =>
-            oid -> toGeneratorParams(oParams, m.get(oid))
+            oid -> toGeneratorParams(NonEmptyList.fromListUnsafe(oParams), m.get(oid))
```
cquiroz, line 150: Do we know `oParams` is non-empty?

swalker2m: Yes, it is created by `groupBy` which will not create an empty grouping.

PR #734, opened 2023-10-01
make the refined codec names more consistent

swalker2m:
It bugged me that we had inconsistent naming for the skunk codecs related to `refined` types.  I think I introduced `pos_int` some time ago not knowing that the preferred form matches the underlying database type more closely.  At any rate we had `pos_int` and `int4_nonneg` and now it consistently matches `<database type>_<refinement>`:

* `int4_nonneg`
* `int4_pos`

etc.

### DIFF
--- a/modules/service/src/main/scala/db/migration/SmartGcalTable.scala
+++ b/modules/service/src/main/scala/db/migration/SmartGcalTable.scala
@@ -91,7 +91,7 @@ object SmartGcalTable {
   def valueEncoder[A](using e: Encoder[A]): Encoder[SmartGcalValue[A]] =
     (
       step_config_gcal *:
-      pos_int          *:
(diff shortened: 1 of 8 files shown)


modules/service/src/main/scala/db/migration/SmartGcalTable.scala
```diff
@@ -91,7 +91,7 @@ object SmartGcalTable {
   def valueEncoder[A](using e: Encoder[A]): Encoder[SmartGcalValue[A]] =
     (
       step_config_gcal *:
-      pos_int          *:
+      int4_pos          *:
```
cquiroz, line 94: missalignment?

cquiroz (approved):
LGTM

PR #739, opened 2023-10-02
Include group info on Observation view

### LINKED ISSUE #439: Include parent information on groupEdit subscription
This is so we can update Explore correctly when, for example, a group is moved. Both `parentGroup` and `parentGroupIndex` would be need to be included

### LINKED ISSUE #440: Include group info on observationEdit subscription
To move observations when they are moved between/outside groups in Explore. Both `groupId` and `groupIndex` would need to be included

hugo-vrijswijk:
Fixes #440

Adds group info to the `Observation` type

### DIFF
--- a/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -7145,7 +7145,7 @@ type Observation {
   """
   Index in enclosing group or at the top level if ungrouped. If left unspecified on creation, observation will be added last in its enclosing group or at the top level. Cannot be set to null.
   """
-  groupIndex: NonNegShort
+  groupIndex: NonNegShort!
 
 }
 
--- a/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -7137,6 +7137,16 @@ type Observation {
   """
   itc: Itc!
 
+  """
+  Enclosing group, if any.
+  """
+  groupId: GroupId
+
+  """
+  Index in enclosing group or at the top level if ungrouped. If left unspecified on creation, observation will be added last in its enclosing group or at the top level. Cannot be set to null.
+  """
+  groupIndex: NonNegShort
+
 }
 
 type Itc {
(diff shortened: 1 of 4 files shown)


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -7137,6 +7137,16 @@ type Observation {
...
 
+  """
+  Enclosing group, if any.
+  """
+  groupId: GroupId
+
+  """
+  Index in enclosing group or at the top level if ungrouped. If left unspecified on creation, observation will be added last in its enclosing group or at the top level. Cannot be set to null.
+  """
+  groupIndex: NonNegShort
```
rpiaggio, line 7148: Should this be `NonNegShort!` if it can't be null?

hugo-vrijswijk: Hmm, maybe? They are nullable in the database, but none of the rows on staging are.  I copied this over from another part of the schema.

image

image

It looks like `groupIndex` is nonnullable? @tpolecat ?

tpolecat: Should be `NOT NULL`. It starts off `NULL` here https://github.com/gemini-hlsw/lucuma-odb/blob/main/modules/service/src/main/resources/db/migration/V0290__groups.sql#L39 but it's set to `NOT NULL` at line 57.

tpolecat:
```
lucuma-odb=# \d t_observation
                                                                                                                                                                                                                        Table "public.t_observation"
           Column           |            Type             | Collation | Nullable |                                                                                                                                                                                          Default                                                                                                                                                                                          
----------------------------+-----------------------------+-----------+----------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 c_program_id               | d_program_id                |           | not null | 
...
 c_group_id                 | d_group_id                  |           |          | 
 c_group_index              | smallint                    |           | not null | 
...
```

hugo-vrijswijk: Thanks, fixed (and in the other place it was used too)

hugo-vrijswijk:
> Looks good so far. Can you add these fields to some of the tests (for obs creation maybe? group manipulation?) to ensure that they're selected properly?

Added, I wasn't sure where the tests were at first 😅

tpolecat (approved):

PR #744, opened 2023-10-04
add parent information to groups

### LINKED ISSUE #439: Include parent information on groupEdit subscription

tpolecat:
Adds `parentId` and `parentIndex` to `type Group`. This just required adding them to the schema and exposing existing columns. Also added an improved test method for subscriptions. Resolves #439


swalker2m (approved):
Looks great.  :+1:

hugo-vrijswijk (approved):

PR #745, opened 2023-10-04
remove cached itc result when replacing a target with its clone

### LINKED ISSUE #687: Internal error on mutation
This mutation produces an internal error (Note this is a target shared by more than one observation)

```graphql
mutation($input: CloneTargetInput!) {
  cloneTarget(input: $input) {
    newTarget {
      id
    }
  }
}
```

```json
{
    "input": {
        "targetId": "t-105",
        "REPLACE_IN": [
            "o-10a"
        ]
    }
}
```

Produces this response:
```json
{
  "errors": [
    {
      "message": "An internal error of type PostgresErrorException occurred."
    }
  ]
}
```

tpolecat:
Resolves #687, which was triggered when a target is replaced with its own clone in an observation with cached ITC results. I'm deleting the ITC results in `TargetService` because you can't easily get ahold of an `ItcService` from there.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/TargetService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/TargetService.scala
@@ -165,8 +165,10 @@ object TargetService {
 
         def replaceIn(tid: Target.Id): F[Unit] =
           input.REPLACE_IN.traverse_ { which =>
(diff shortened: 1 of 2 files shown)


swalker2m (approved):
Thanks for taking care of this.  I didn't notice when I added the ITC cache. :+1:

PR #750, opened 2023-10-10
update libs: Adds duration to parallactic angle calculation.

toddburnside:
How the duration is going to provided is an open question at the moment. In the meantime, I just used the full observation duration - which is not correct in cases where the observation is being split. It may be that the duration will need to be specified in the API along with the observation time.


modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
```diff
@@ -251,15 +252,18 @@ object GuideEnvironmentService {
...
         oid: Observation.Id
-      ): F[Either[Error, Option[NonEmptyList[Offset]]]] =
+      ): F[Either[Error, (TimeSpan, Option[NonEmptyList[Offset]])]] =
         generator(commitHash, itcClient, plannedTimeCalculator)
           .digest(pid, oid)
           .map {
             _.leftMap(Error.GeneratorError(_))
-              .map { d => NonEmptyList.fromFoldable(d.science.offsets.union(d.acquisition.offsets)) }
+              .map { d => 
+                (d.fullPlannedTime.sum,
```
toddburnside, line 264: @swalker2m Is this the correct time to use for the duration for the calculation? Or should it be something like `digest.science.plannedTime.sum`?

swalker2m (approved):
Thanks, looks good. :+1:

PR #754, opened 2023-10-10
Dataset queries

swalker2m:
Switches `Dataset.Id` / `DatasetId` to a GID-based id and implements the `dataset` and `datasets` queries.


modules/service/src/main/scala/lucuma/odb/graphql/mapping/DatasetSelectResultMapping.scala
```diff
@@ -0,0 +1,20 @@
...
+  extends DatasetTable[F]
+     with ResultMapping[F] {
+
+  lazy val DatasetSelectResultMapping: TypeMapping =
+    SwitchMapping(
+      DatasetSelectResultType,
+      List(
+        QueryType / "datasets" -> topLevelSelectResultMapping(DatasetSelectResultType)
+      )
+    )
```
swalker2m, line 18: `datasets` can also be queried from the `observation`, so we'll justify the `SwitchMapping` when that is added.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -4819,12 +4819,20 @@ type Coordinates {
   dec: Declination!
 }
 
+"DatasetId id formatted as `o-[1-9a-f][0-9a-f]*`"
+scalar DatasetId
```
swalker2m, line 4823: Copy/paste typo here with the `o` prefix.  I'll fix it.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/MutationMapping.scala
```diff
@@ -606,25 +606,25 @@ trait MutationMapping[F[_]] extends Predicates[F] {
...
 
   private lazy val UpdateObsAttachments =
-      MutationField("updateObsAttachments", UpdateObsAttachmentsInput.binding(Path.from(ObsAttachmentType))) { (input, child) =>
-        services.useTransactionally {
-          val filterPredicate = and(List(
-            Predicates.obsAttachment.program.id.eql(input.programId),
-            Predicates.obsAttachment.program.isWritableBy(user),
-            input.WHERE.getOrElse(True)
-          ))
+    MutationField("updateObsAttachments", UpdateObsAttachmentsInput.binding(Path.from(ObsAttachmentType))) { (input, child) =>
```
swalker2m, line 609: Unrelated.  This was just indented too much.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/QueryMapping.scala
```diff
@@ -113,6 +118,53 @@ trait QueryMapping[F[_]] extends Predicates[F] {
...
+        NonNegIntBinding.Option("LIMIT", rLIMIT)
+      ), child) =>
+        (rWHERE, rOFFSET, rLIMIT).parTupled.flatMap { (WHERE, OFFSET, LIMIT) =>
+          val limit = LIMIT.foldLeft(ResultMapping.MaxLimit)(_ min _.value)
+          ResultMapping.selectResult("datasets", child, limit) { q =>
+            FilterOrderByOffsetLimit(
+              pred = Some(and(List(
+                OFFSET.map(Predicates.dataset.id.gtEql).getOrElse(True),
+                Predicates.dataset.observation.program.isVisibleTo(user),
+                WHERE.getOrElse(True)
```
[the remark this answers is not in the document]
swalker2m: Nope, it looks like the `Join` goes in the `nestedSelectResultMapping` of `DatasetSelectResultMapping`.  That works.

modules/service/src/main/scala/lucuma/odb/service/DatasetService.scala
```diff
@@ -66,26 +69,28 @@ object DatasetService {
...
-        val insert: F[Either[InsertDatasetFailure, Dataset.Id]] =
+        val insert: F[Either[InsertDatasetFailure, (Dataset.Id, Step.Id, PosShort)]] =
           session
             .unique(Statements.InsertDataset)(stepId, filename, qaState)
             .map(_.asRight[InsertDatasetFailure])
             .recover {
               case SqlState.UniqueViolation(_)     => ReusedFilename(filename).asLeft
               case SqlState.ForeignKeyViolation(_) => StepNotFound(stepId).asLeft
+              case SqlState.NotNullViolation(ex) if ex.getMessage.contains("c_observation_id") =>
+                StepNotFound(stepId).asLeft
```
toddburnside, line 80: Is `c_observation_id` correct? It doesn't seem to have anything to do with the insert.

swalker2m: A question is whether it's a good idea to include the observation id since it can be looked up via the step.  It's needed quite a bit though, is very convenient to have, and is a few joins away so it _seems_ like a good idea?  <grimace>

toddburnside: OK. Just making sure. 😄

modules/service/src/main/resources/db/migration/V0378__datasets_update.sql
```diff
@@ -0,0 +1,64 @@
...
+    RAISE EXCEPTION 'Cannot change observation id once set';
+  END IF;
+  RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER check_dataset_observation_immutable_trigger
+BEFORE UPDATE ON t_dataset
+FOR EACH ROW
+EXECUTE FUNCTION check_dataset_observation_immutable();
```
tpolecat, line 64: 👍🏻 👍🏻 Good. Delete and truncate are ok I guess?

tpolecat (approved):
This looks great 👍🏻

modules/service/src/main/resources/db/migration/V0378__datasets_update.sql
swalker2m: I think so yes, since the entire row is being removed in those cases we can't end up with a dataset without an observation.

PR #761, opened 2023-10-11
updateDatasets mutation

swalker2m:
Adds the `updateDatasets` mutation.  The only thing that can be updated at this point is the QA state.


toddburnside (approved):
LGTM

PR #762, opened 2023-10-12
set dataset times on event reception

swalker2m:
The presence of a dataset end time (as recorded in the nullable `t_dataset.c_end_time` column) marks a completed dataset.  The dataset start and end times are inferred from dataset events.  We expect a number of events from Observe as it executes, but we need to know when it is really considered "done".  As it stands, the reception of an `EndWrite` dataset event that occurs sometime after a `StartObserve` dataset event will cause the end time to be set and hence mark the dataset as done.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/graphql/table/DatasetTable.scala
+++ b/modules/service/src/main/scala/lucuma/odb/graphql/table/DatasetTable.scala
@@ -36,7 +36,7 @@ trait DatasetTable[F[_]] extends BaseMapping[F] {
 
     object Time {
       val Start: ColumnRef     = col("c_start_time", core_timestamp.opt)
-      val End: ColumnRef       = col("c_start_time", core_timestamp.opt)
+      val End: ColumnRef       = col("c_end_time",   core_timestamp.opt)
     }
   }
 
--- a/modules/service/src/main/scala/lucuma/odb/service/DatasetService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/DatasetService.scala
@@ -15,6 +15,7 @@ import lucuma.core.enums.DatasetQaState
 import lucuma.core.model.User
 import lucuma.core.model.sequence.Dataset
 import lucuma.core.model.sequence.Step
+import lucuma.core.util.Timestamp
 import lucuma.odb.data.Nullable
 import lucuma.odb.graphql.input.DatasetPropertiesInput
 import lucuma.odb.util.Codecs.*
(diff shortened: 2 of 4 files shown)


swalker2m:
Something like this

```scala
    val SelectCompletedStepRecordsForObs: Query[Observation.Id, (Atom.Id, NonNegShort, SequenceType, Step.Id)] =
      (sql"""
        SELECT
          a.c_atom_id,
          a.c_step_count,
          a.c_sequence_type,
          s.c_step_id
        FROM
          t_step_record s
        INNER JOIN
          t_atom_record a
        ON a.c_atom_id = s.c_atom_id
        WHERE """ ~> sql"""a.c_observation_id = $observation_id AND s.c_completed IS NOT NULL""" <~ sql"""
          AND NOT EXISTS (
            SELECT 1
            FROM   t_dataset d
            WHERE
              d.c_step_id = s.c_step_id
              AND (
                d.c_end_time IS NULL
                OR (d.c_qa_step IS NOT NULL AND c.c_qa_step <> 'Pass'::e_dataset_qa_state)
              )  
          )                   
        ORDER BY s.c_completed
      """).query(atom_id *: int2_nonneg *: sequence_type *: step_id)
```

I think would become the new completed step query.  Here what's new is the `AND NOT EXISTS ( ... )` section which is trying to say that there are no datasets for this step which either haven't ended (`d.c_end_time IS NULL`) or which have an assigned QA state other than `Pass`.

Maybe the dataset end time check is superfluous if we are checking the dataset QA state though?  If so, we don't really need the dataset time update offered in this PR and we might in fact even remove `c_start_time` and `c_end_time`.

modules/service/src/main/scala/lucuma/odb/graphql/table/DatasetTable.scala
```diff
@@ -36,7 +36,7 @@ trait DatasetTable[F[_]] extends BaseMapping[F] {
 
     object Time {
       val Start: ColumnRef     = col("c_start_time", core_timestamp.opt)
-      val End: ColumnRef       = col("c_start_time", core_timestamp.opt)
+      val End: ColumnRef       = col("c_end_time",   core_timestamp.opt)
```
swalker2m, line 39: Bugfix.

modules/service/src/main/scala/lucuma/odb/service/DatasetService.scala
```diff
@@ -144,5 +167,20 @@ object DatasetService {
...
       }
     }
+
+    val SetStartTime: Command[(Timestamp, Dataset.Id)] =
+      sql"""
+        UPDATE t_dataset
+           SET c_start_time = $core_timestamp,
+               c_end_time   = NULL
+         WHERE c_dataset_id = $dataset_id
+      """.command
```
swalker2m, line 177: Resets the end time, if it has been previously set.

swalker2m (commented):
I'll mark this ready now.

tpolecat (approved):

PR #765, opened 2023-10-16
Fixes skunk decode exception when observation has no configuration

toddburnside:
I guess I must have thought that the decoder error message would be reported to the user, but it is not. Since not having a configuration is a common thing, the user needs a better error message than `An internal error of type DecodeException occurred`. 😆

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
@@ -129,13 +129,15 @@ object GuideEnvironmentService {
     constraints:        ConstraintSet,
     posAngleConstraint: PosAngleConstraint,
     optWavelength:      Option[Wavelength],
-    fpu:                Either[GmosNorthFpu, GmosSouthFpu],
+    optFpu:             Option[Either[GmosNorthFpu, GmosSouthFpu]],
     explicitBase:       Option[Coordinates]
   ) {
-    def agsParams: AgsParams                  = AgsParams.GmosAgsParams(fpu.some, PortDisposition.Side)
+    def agsParams: AgsParams                 = AgsParams.GmosAgsParams(optFpu, PortDisposition.Side)
     def wavelength: Either[Error, Wavelength] =
       optWavelength.toRight(Error.GeneralError(s"No wavelength defined for observation $id."))
-    def site: Site = fpu.fold(_ => Site.GN, _ => Site.GS)
+    def fpu: Either[Error, Either[GmosNorthFpu, GmosSouthFpu]] =
+      optFpu.toRight(Error.GeneralError(s"No configuration defined for observation $id."))
+    def site:       Either[Error, Site] = fpu.map(_.fold(_ => Site.GN, _ => Site.GS))
   }
 
   def instantiate[F[_]: Concurrent](
@@ -317,6 +319,7 @@ object GuideEnvironmentService {
       ): F[Either[Error, GuideEnvironment]] =
         (for {
           obsInfo        <- EitherT(getObservationInfo(pid, oid))
+          site           <- EitherT.fromEither(obsInfo.site)
           asterism       <- EitherT(getAsterism(pid, oid))
           tracking        = ObjectTracking.fromAsterism(asterism)
           tAndO          <- EitherT(getTimeAndOffsets(pid, oid))
@@ -345,7 +348,7 @@ object GuideEnvironmentService {
                                 )
                             )
           positions      <- EitherT.fromEither(
-                              getPositions(oid, obsInfo.site, obsInfo.posAngleConstraint, offsets, tracking, obsTime, obsDuration)
+                              getPositions(oid, site, obsInfo.posAngleConstraint, offsets, tracking, obsTime, obsDuration)
                             )
           usable         <- EitherT.fromEither(
                               processCandidates(obsInfo, baseCoords, scienceCoords, positions, candidates)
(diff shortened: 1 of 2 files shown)


modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
```diff
@@ -129,13 +129,15 @@ object GuideEnvironmentService {
     constraints:        ConstraintSet,
     posAngleConstraint: PosAngleConstraint,
     optWavelength:      Option[Wavelength],
-    fpu:                Either[GmosNorthFpu, GmosSouthFpu],
+    optFpu:             Option[Either[GmosNorthFpu, GmosSouthFpu]],
     explicitBase:       Option[Coordinates]
   ) {
-    def agsParams: AgsParams                  = AgsParams.GmosAgsParams(fpu.some, PortDisposition.Side)
+    def agsParams: AgsParams                 = AgsParams.GmosAgsParams(optFpu, PortDisposition.Side)
```
swalker2m, line 135: I guess we're anticipating a refactor when we introduce some other instrument?

modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
```diff
@@ -317,6 +319,7 @@ object GuideEnvironmentService {
       ): F[Either[Error, GuideEnvironment]] =
         (for {
           obsInfo        <- EitherT(getObservationInfo(pid, oid))
+          site           <- EitherT.fromEither(obsInfo.site)
```
swalker2m, line 322: This is what introduces the "No configuration defined ..." message?

modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
```diff
@@ -437,22 +440,22 @@ object GuideEnvironmentService {
...
+              case (Some(north), None) => north.asLeft.some
+              case (None, Some(south)) => south.asRight.some
+              case _                   => none
             }
 
           val explicitBase: Option[Coordinates] =
             (ra, dec).mapN(Coordinates(_, _))
 
-          (elevRange, fpu).mapN((elev, f) =>
+          elevRange.map(elev =>
```
swalker2m, line 453: Was the problem that they are `Either`s but switching to `Validated` and then back would work or that the message just doesn't make it up to where it needs to go?  If the later, do we fail to see the "Invalid elevation range in observation" message when it happens too?

swalker2m (approved):
Looks good. :+1:

modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
toddburnside: Yes, we would fail to see the `invalid elevation range in observation` message as well, but this should never happen, right? Observations are required to have contraints which include an elevation range, IIRC.

modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
toddburnside: Yes. I get the site by knowing what the configuration is. Is there a better way of knowing the site that I should use?

modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
toddburnside: Yeah. I think we're going to need to do some refactoring to get whatever information is needed for other `AgsParams` types. Right now, there is only `GmosAgsParams`. We'll also need to have different guide probe(s), etc. If you know of things that would help mitigate that future refactoring, I could do those now. Maybe `guideProbe` could be a value in `ObservationInfo`? But, I don't really know how other instruments will work.

toddburnside: Oh, now that I look, the `guideProbe` is part of `AgsAnalysis.Usable`. I missed that before. I can switch to that.

modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
swalker2m: No, I was just looking to find how the message would be triggered but I see it's via the `site` which asks for the `fpu`.  You're using the `fpu` anyway to define the AGS params.

For what it's worth, the observation also directly defines the `c_observing_mode_type` which will tell you `GmosNorthLongSlit` or `GmosSouthLongSlit` or neither.  The observing mode defines which instrument it applies to, and we probably need something that facilitates mapping from `Instrument` to `Site` somewhere at some point if it doesn't exist.

modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
toddburnside: Actually, I'll do that in my next PR as I'm making some changes that will make that easier.

modules/service/src/main/scala/lucuma/odb/service/GuideEnvironmentService.scala
toddburnside: Let me see if there is a mapping from Instrument to Site and I'll add that into my next PR if possible. I guess `c_observing_mode_type` would be null if there isn't a configuration?

toddburnside: OK. I couldn't find one either.

PR #776, opened 2023-10-23
Use data from Generator in AGS

toddburnside:
Adds `Generator.digestWithParmsAndHash()` for use by the `GuideEnvironmentService`. The hash is not needed quite yet, but guide star availability is also going to be cached. Since it depends on the digest, it seemed simpler to make use of the hash from the Generator as part of the hast for invalidating the cached GS availability. And, since the Generator was already getting the necessary configuration information, it seemed silly to get it again in the GuideEnvironmentService.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/logic/Generator.scala
+++ b/modules/service/src/main/scala/lucuma/odb/logic/Generator.scala
@@ -57,14 +57,26 @@ import Generator.FutureLimit
 sealed trait Generator[F[_]] {
 
   /**
-   * Looks up the parameters required to calculate the ExecutionDigest, performs
-   * the calculation, and caches the results. If the observation is not
-   * completely defined (e.g., if missing the observing mode), an Error is produced.
+   * Looks up the parameters required to calculate the ExecutionDigest, and checks
+   * the cache. If not in the cache, it performs the calculation and caches the 
+   * results. If the observation is not completely defined (e.g., if missing the 
+   * observing mode), an Error is produced.
    */
   def digest(
     programId:     Program.Id,
     observationId: Observation.Id
   )(using NoTransaction[F]): F[Either[Error, ExecutionDigest]]
+
+  /**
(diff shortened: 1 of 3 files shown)


modules/service/src/main/scala/lucuma/odb/logic/Generator.scala
```diff
@@ -57,15 +57,27 @@ import Generator.FutureLimit
...
+  /**
+   * The same is `digest()`, but it also returns the GeneratorParms and the hash used
+   * to determine if the digest needed to be recalculated. This is useful in things
+   * like the guide star availability calculations which depend on the digest and are
+   * also cached.
+   */
+  def digestWithParamsAndHash(
+    programId:     Program.Id,
+    observationId: Observation.Id
+  )(using NoTransaction[F]): F[Either[Error, (ExecutionDigest, GeneratorParams, Md5Hash)]]
```
swalker2m, line 79: There are certainly things included in the params and hash that do not impact the guide star calculation.  I guess that's fine though since it will just make the cache a little less efficient as it kicks out entries that are only replaced with the same values.  I wonder if the reverse is also true?  It seems possible that there are things that the sequence generation doesn't need but which the guide star calculation would.  The observing conditions for example are taken into account only indirectly via the integration time calculation from ITC?

toddburnside: For the first part, I wasn't sure of a better approach. I'd have to know which things were effecting the things that affected the guide star calculation and then make a hash of them. Maybe it would be better to take the GeneratorParams and make a hash of only the stuff that impacted the guide star calculation (the offsets, etc.)? But, I don't know what those are. 

For the second part, there definitely are additional things that the guide star calculations require. I was going to use the generator hash as part of a new hash. But this kept me from having to understand/duplicate what the generator was already doing.

swalker2m (approved):
Looks fine; I don't see any problem exposing a bit more information.  :+1:

PR #785, opened 2023-10-30
Select atom records

swalker2m:
API Updates

* Drops instrument-specific atom records in favor of `AtomRecord`.
* Adds a `StepRecord` interface, with instrument-specific implementations.
* Adds a `steps` query to `AtomRecord` for retrieving executed step data as `StepRecord`.
* Switches `stepId: StepId!` fields for `step: StepRecord!`.

This is mostly mechanical but there are a couple of concerns:

1. As you drill down into collections in collections each individual query is limited in result size (currently to 1,000).  Nothing limits the overall size of the result though.  So if you actually had 1,000 visits each with 1,000 atoms with 1,000 steps with 1,000 datasets it would ... not work.  Not sure what to do about that.
2. I was able to drop the instrument-specific atom records but at a cost.  When you get the visits for an observation you have to match on the instrument-specific type if you want to see the `static` configuration.  Then you can ask for its atoms and the steps in those atoms.  If you want the `instrumentConfig` of the step, you again would have to match on the instrument-specific type.

```
        query {
          observation(observationId: "o-100") {
            execution {
              visits {
                matches {
                  ... on GmosNorthVisit {
                    static {
                      stageMode
                    }
                    atomRecords {
                      matches {
                        steps {
                          matches {
                            stepConfig {
                              stepType
                            }
                            ... on GmosNorthStepRecord {
                              instrumentConfig {
                                fpu {
                                  builtin
                                }
                              }
                            }
                          }
                        }
                      }
                    }
                  }
                }
              }
            }
          }
        }

```


modules/service/src/main/scala/lucuma/odb/graphql/mapping/LookupFrom.scala
```diff
@@ -0,0 +1,37 @@
...
+
+  private def defineLookup(tpes: List[Type], path: Seq[String], mapping: ObjectMapping): List[(Path, ObjectMapping)] =
+    tpes.toList.map { tpe =>
+      (path.foldLeft(Path.from(tpe)) { case (t, p) => t / p }, mapping)
+    }
+
+  def lookupFromStepRecord(mapping: ObjectMapping, path: String*): List[(Path, ObjectMapping)] =
+    defineLookup(StepRecordTypes, path, mapping)
+
+  def lookupFromVisit(mapping: ObjectMapping, path: String*): List[(Path, ObjectMapping)] =
```
swalker2m, line 34: Not particularly pleased by this, but wherever there is a `SwitchMapping` from `Visit` (in this case), we have to define the same mapping from each of the interface type `Visit` and the two concrete types `GmosNorthVisitType` and `GmosSouthVisitType`.  Since this is needed multiple times it seemed important to avoid repeating the type list.

tpolecat:
> 1 ... if you actually had 1,000 visits each with 1,000 atoms with 1,000 steps with 1,000 datasets it would ... not work. Not sure what to do about that.

We could limit the fetched resultset to some arbitrary size, but we need to be sure that paging is available in cases where the user is likely to hit it. We could also have a time limit in case the generated query is overly complex. So I think we have some [draconian] options.

> 2 ... When you get the visits for an observation you have to match on the instrument-specific type if you want to see the static configuration

This seems reasonable to me. The syntax is a little bulky but this does seem like a fair use case for matching on an interface type.

swalker2m:
Alright, I've marked this ready.

tpolecat (approved):
This looks very good. :+1:

PR #787, opened 2023-10-31
improve group events

tpolecat:
This changes the group event stream to trigger when the contents of a group change, so you get an event when an observation is created, and two events when an observation is moved.

### DIFF
--- a/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -3630,7 +3630,7 @@ input TargetEditInput {
 
 input GroupEditInput {
     """
-    Group ID
+    Group ID, or Null to watch top-level group(s), or omit to watch all groups.
     """
     groupId: GroupId
 
@@ -3686,10 +3686,13 @@ type GroupEdit {
   """
   editType: EditType!
 
+  "Assocated program"
+  program: Program!
+
   """
-  Edited object
+  Edited group, or null if it's the top-level group.
   """
-  value: Group!
+  value: Group
   id: Long! @deprecated(reason: "id is no longer computed; a constant value is returned")
 }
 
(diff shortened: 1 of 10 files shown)


tpolecat:
@hugo-vrijswijk maybe you can say a few words about how you intend to consume this event. We could add more information (like a `CONTENTS_CHANGED` event type for instance)  but I'm not sure what you need.

swalker2m (commented):
This looks good / makes sense to me.  Is it still a WIP?

tpolecat:
It's ready, assuming it does what Hugo needs. I want to double-check with him before merging.

hugo-vrijswijk:
I think this is good. Ideally there would be three events (or 2 when moved inside the same group)

- `groupEdit` if a group is moved (with `parentGroupId` and/or `parentGroupIndex` changed)
- `observationEdit` when an obs is moved (with `groupId` and/or `groupIndex` changed)

And for the old/new groups
- `groupEdit` for the new group (with new `elements`)
- `groupEdit` for the old group (with changed `elements`)
With this being a single event if an element is moved inside the same group.

For the root group it works if there is no event for the root group, only for the new group (and for elements in the root group being moved/updated). I.e:
1. Obs is moved outside of root group
2. `observationEdit` event with a new `groupId` (already works)
3. `groupEdit` event for the group the obs is moved to
4. `observationEdit` with `groupIndex` updated for the elements in the root group (already works, I think)

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -3686,10 +3686,13 @@ type GroupEdit {
...
   editType: EditType!
 
+  "Assocated program"
+  program: Program!
+
   """
-  Edited object
+  Edited group, or null if it's the top-level group.
   """
-  value: Group!
```
hugo-vrijswijk, line 3692: I don't think this has to be nullable. If the root group is updated it's fine if just the elements inside it are updated (with a new index)

tpolecat: I think for consistency it needs to be nullable, but you can ignore this case on your end if you like.

tpolecat:
Need an amen from someone.

swalker2m (approved):
Amen.

PR #788, opened 2023-10-31
Checklist and `interface`

swalker2m:
The `Checklist` "test" is very useful for finding unmapped fields / types / errors in the schema.  Unfortunately it would show fields implemented in an interface as unmapped.  This PR, while admittedly destroying the erstwhile beauty of the `Checklist` code, takes into account interfaces.

For example, before the `instrument`, `observation`, etc. fields would be marked as unmapped.  Now they are shown as implemented along with the interface name in which they are implemented.

```
- [x] GmosNorthVisit
  - [x] id
  - [x] instrument (via Visit)
  - [x] observation (via Visit)
  - [x] created (via Visit)
  - [ ] startTime
  - [ ] endTime
  - [ ] duration
  - [x] atomRecords (via Visit)
  - [x] datasets (via Visit)
  - [x] events (via Visit)
  - [x] static
```

### DIFF
--- a/modules/service/src/test/scala/lucuma/odb/graphql/Checklist.scala
+++ b/modules/service/src/test/scala/lucuma/odb/graphql/Checklist.scala
@@ -5,9 +5,10 @@ package lucuma.odb.graphql
 
 import cats.effect.IO
 import cats.syntax.all._
+import grackle.Field
 import grackle.ObjectType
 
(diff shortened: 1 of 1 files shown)


tpolecat (approved):
Oh, fantastic! 👍🏻 👍🏻

PR #794, opened 2023-11-02
rename migration

tpolecat:
This renames a duplicate migration number. The check @armanbilge added should have caught this but somehow it didn't. :-\


armanbilge:
Thanks for letting me know, I'll take a look what went wrong.

tpolecat:
I think the issue was interleaved PRs. My branch wasn't up to date with `main` but there were no conflicts, so it allowed the merge. Maybe a branch protection rule would fix it (you can turn this on if you like).

toddburnside (approved):
👍

tpolecat:
Well now the migration validator is failing 😭 … it's not happy that I renamed the old file. Urk.

tpolecat:
So uh @armanbilge do you have any ideas? An invalid set of migrations made it into the git history and now the migration checker won't let me change it.

armanbilge:
I think we just have to ignore it in this case? In theory it's doing the right thing: you should not be able to delete migrations.

If we don't like that, then we need to add some whitelist capability (similar to MiMa exceptions).

tpolecat:
How do we ignore it? If we can do that for this one PR then we should be good? It only looks at the previous state right?

tpolecat:
Huh really weird. We have the status check for "up to date" enabled.

armanbilge:
> How do we ignore it?

Slam the merge button like there's no tomorrow 😁 

> Huh really weird. We have the status check for "up to date" enabled.

Yeah, I did the diagnostics and this is an actual bug in the validation action unfortunately. I'm figuring out how to fix it.

tpolecat:
👍🏻 We're back up and running. Thanks @armanbilge

PR #795, opened 2023-11-02
Remove unused ObservationEffectHandler effectHandler.fieldName

toddburnside:
Unless someone knows a reason that the field name should exist...


swalker2m (approved):
We can always revive it if we discover a use for it in the future.

PR #797, opened 2023-11-03
Rename GuideEnvironmentService to GuideService

toddburnside:
It handles both `guideEnvironment` requests and `guideAvailability` requests, so this name makes more sense.

I didn't want to include the renaming in the previous PR because it sometimes makes diffs harder to read.


swalker2m (approved):
Makes sense to me. :+1:

PR #798, opened 2023-11-03
Tighten constraints on guide star queries to gaia

toddburnside:
Uses the observation constraints and wavelength to constrain the gaia query for guide star candidates. In the observations I tested this on, it reduced the candidates by up to a half without changing the end results after analysis.

@cquiroz Since you wrote the code behind all of this, could you take a quick look and make sure it's OK?

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/GuideService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/GuideService.scala
@@ -21,6 +21,7 @@ import lucuma.ags.GuideProbe
 import lucuma.ags.GuideStarCandidate
 import lucuma.ags.*
 import lucuma.catalog.votable.*
+import lucuma.core.enums.GuideSpeed
 import lucuma.core.enums.PortDisposition
 import lucuma.core.enums.Site
 import lucuma.core.geom.ShapeExpression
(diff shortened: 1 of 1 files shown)


modules/service/src/main/scala/lucuma/odb/service/GuideService.scala
```diff
@@ -219,15 +220,18 @@ object GuideService {
...
       ): Either[Error, ADQLQuery] =
         (tracking.at(start.toInstant), tracking.at(end.toInstant))
           .mapN { (a, b) =>
+            val brightnessConstraints = gaiaBrightnessConstraints(constraints, GuideSpeed.Slow, wavelength)
             // Make a query based on two coordinates of the base of an asterism over a year
             CoordinatesRangeQueryByADQL(
               NonEmptyList.of(a.value, b.value),
               shapeConstraint,
-              ags.widestConstraints.some
+              brightnessConstraints.some
```
cquiroz, line 234: the reason too use the widest constraiints is for caching. You'll get more stars but you'll be able to reuse the results for more stringent queries
As a plain function using widest will be less efficient but with cached results (like explore) it won't. I guess it depends on the context

toddburnside: Makes sense. In the odb, we won't be caching the results from gaia, so I'm trying to reduce processing time by having fewer candidates.

cquiroz: Should we have two entry points then?  so it is suitable for either case?

toddburnside: I'm not sure I see the point. This can only be called from within the GuideService, so if we wanted to start caching, other changes would be required as well. Alternatively, `getGaiaQuery` could be called with `ags.widestConstraints`

cquiroz (approved):
LGTM

PR #805, opened 2023-11-07
Library Updates

swalker2m:
A handful of easy updates.

We're held back on clue 0.34.0 at the moment because we depend on `lucuma-itc` which depends on `lucuma-graphql-routes` which depends on Grackle and clue.    Updating the train requires updating Grackle in the ITC.


tpolecat:
Cool, thanks. I'm working on the Grackle update and will see if I can update lucuma-itc.

PR #808, opened 2023-11-07
Execution Event Table

swalker2m:
I have waffled on whether I wanted an execution event table per type (sequence, step, and dataset) with a shared key sequence and a view that combines them or else just a single wider table with nullable columns a check constraint.  I think I'll switch to a single table because for time accounting I'll need a couple of different indices on the events (by ids in one case and by timestamp in another).


tpolecat (approved):
This seems like a reasonable simplification. 👍🏻

PR #809, opened 2023-11-07
Clue 0.34, Grackle 0.16, other minor changes

tpolecat:


swalker2m (approved):
Looks great. :+1:

build.sbt
```diff
@@ -1,33 +1,33 @@
 val catsParseVersion           = "0.3.10"
```
swalker2m, line 1: I think we can update `cats-parse` since you updated core.

tpolecat: The latest tagged core doesn't include that change, so I'll leave it for now.

PR #824, opened 2023-11-15
update deps for cats-parse 1

tpolecat:


toddburnside:
I think we should be able to get rid of the 4 lines like this, too:

> libraryDependencySchemes += "org.typelevel" %% "cats-parse" % "always", // Workaround until http4s updates to 1.0.0 as well

toddburnside (approved):
LGTM

swalker2m (approved):
Thanks for handling all this.  We probably should remove the `always` lines as Todd says?

PR #831, opened 2023-11-17
Add observe class to step record

swalker2m:
Time accounting needs the observe class of each step in order to charge the proper time category.  Unfortunately I left this out of the step record table.  This PR fixes that oversight.


modules/service/src/main/scala/lucuma/odb/graphql/input/RecordGmosVisitInput.scala
```diff
@@ -0,0 +1,35 @@
...
+        staticMatcher("static", rStatic)
+      ) =>
+        (rObservationId, rStatic).parMapN(RecordGmosVisitInput(_, _))
+    }
+
+  val GmosNorthBinding: Matcher[RecordGmosVisitInput[GmosNorth]] =
+    binding(GmosNorthStaticInput.Binding)
+
+  val GmosSouthBinding: Matcher[RecordGmosVisitInput[GmosSouth]] =
+    binding(GmosSouthStaticInput.Binding)
```
swalker2m, line 33: This change is unrelated to the goal of adding the observe class.  I refactored the GMOS step recording input to combine the north and south versions and wanted to make the visit match.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/AtomRecordMapping.scala
```diff
@@ -13,13 +13,13 @@ import lucuma.odb.graphql.binding.TimestampBinding
 import lucuma.odb.graphql.predicate.Predicates
 
 import table.AtomRecordTable
-import table.StepRecordTable
+import table.StepRecordView
```
swalker2m, line 16: This is unrelated as well, but `StepRecordTable` should have been (re)named `StepRecordView` since it refers to columns in the view.

tpolecat (approved):
Looks great! 👍🏻

PR #835, opened 2023-11-21
Cache guide availability results

toddburnside:
The implementation avoids recalculating anything that it already has in the cache. So, if you request a subset of what is in the cache, it does nothing. If you request something that overlaps or abuts, it only calculates what is missing. If the request is completely disjoint, it will fill in the intermediate time as well, as long as the gap is not too large.

Most of the decision making for what will be calculated is encapsulated in the `ContiguousTimestampMap`. I don't know if this will be generally useful elsewhere, but it seemed better to extract it for testing, and it was easy to generalize it.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/data/ContiguousTimestampMap.scala
+++ b/modules/service/src/main/scala/lucuma/odb/data/ContiguousTimestampMap.scala
@@ -64,6 +64,9 @@ final case class ContiguousTimestampMap[A: Eq] private (
 object ContiguousTimestampMap {
   def empty[A: Eq]: ContiguousTimestampMap[A] = ContiguousTimestampMap(none, SortedMap.empty)
 
+  def single[A: Eq](period: TimestampInterval, a: A): ContiguousTimestampMap[A] =
+    ContiguousTimestampMap(period.some, SortedMap.from(List((period, a))))
+
   def fromList[A: Eq](list: List[(TimestampInterval, A)]): Option[ContiguousTimestampMap[A]] =
     list.sortBy(_._1).foldRight(empty.some){ case ((i, a), ectm) => ectm.flatMap(_.add(i, a)) }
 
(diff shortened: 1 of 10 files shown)


modules/service/src/main/scala/lucuma/odb/data/ContiguousTimestampMap.scala
```diff
@@ -0,0 +1,80 @@
...
+package lucuma.odb.data
+
+import cats.Eq
+import cats.Order.*
+import cats.syntax.all.*
+import lucuma.core.util.TimestampInterval
+
+import scala.collection.immutable.SortedMap
+
+final case class ContiguousTimestampMap[A: Eq] private (
```
tpolecat, line 13: This has some similarities to `TimestampUnion` (which is set rather than a map) so it might make sense to put it in core, call it `TimestampMap`, and make the APIs similar. It's possible that they could share some logic. You may have considered this already.

toddburnside: I considered it briefly, but there seemed to be some significant differences. I will take another look now that I have completed this one and see if there is enough of a shared API to make that seem reasonable. I think the shared logic is unlikely, but I'll take a look.

I was going to call this `TimestampMap`, but it seemed to me that it would imply something more like a regular `Map` with Timestamp keys. The requirement for keeping it contiguous and the merging of adjacent periods with the same values made me think it required a different name. There could be a better name than the current one, though.

toddburnside: I think this is superficially similar to `TimestampUnion` in that both are collections of `TimestampInterval`s. However, the job of `TimestampUnion` is to allow disjoint, but not overlapping or abutting intervals, effectively merging any that would abut or overlap. `ContiguousTimestampMap` explicitly disallows any disjoint intervals and requires that all the contained intervals exactly abut to form a contiguous block of time and that each added interval abuts exactly with one end or the other of the contiguous time block. Most of the "shared" logic resides in the `TimestampInterval`s. The specific requirements of `ContiguousTimestampMap` are required by the guide environment cache, but so specific that I think it unlikely to be useful elsewhere. I am open to persuasion otherwise.

tpolecat (approved):
This was a lot of work. Looks good to me. 👍🏻

PR #837, opened 2023-11-22
Optimize file size counter in S3FileService

armanbilge:


modules/service/src/main/scala/lucuma/odb/service/S3FileService.scala
```diff
@@ -114,7 +114,7 @@ object S3FileService {
           val f = for {
             ref  <- Ref.of(0L)
             pipe  = s3.uploadFileMultipart(awsConfig.bucketName, filePath.toKey, partSize)
-            _    <- data.evalTapChunk(c => ref.update(_ + 1)).through(pipe).compile.drain
+            _    <- data.chunks.evalTap(c => ref.update(_ + c.size)).unchunks.through(pipe).compile.drain
```
armanbilge, line 117: `evalTapChunk` runs the effect once per `Byte` i.e. `c` was a `Byte` and not a `Chunk`. Confusingly the "`Chunk`" aspect of its name relates to how the stream is running internally i.e. it is grabbing an entire `Chunk`, then running the effect, instead of grabbing a single element and then running the effect.

toddburnside (approved):
LGTM. Thanks, Arman!

PR #841, opened 2023-11-23
Time Accounting API Updates

swalker2m:
Updates the API to add initial time accounting support, moving to `CategorizedTime` from https://github.com/gemini-hlsw/lucuma-core/pull/839.

The updates can be summarized:

```
Program:

  plannedTimeRange:  PlannedTimeRange @deprecated(reason: "Use timeEstimateRange instead.")

  # This is the new field for planned time.
  timeEstimateRange: CategorizedTimeRange

  # Time accounting information for the program as a whole.
  timeCharge:        CategorizedTime!


Execution (Observation):
  digest:
    acquisition:
      plannedTime: PlannedTime! @deprecated(reason: "Use timeEstimate instead.")
      timeEstimate: CategorizedTime
    science
      plannedTime: PlannedTime! @deprecated(reason: "Use timeEstimate instead.")
      timeEstimate: CategorizedTime

  # Time accounting information for the observation as a whole.
  timeCharge: CategorizedTime!

Visit:

  # Detailed time accounting information for the visit.
  timeChargeInvoice: TimeChargeInvoice!
    executionTime: CategorizedTime!
    discounts:     [TimeChargeDiscount!]!
    corrections:   [TimeChargeCorrection!]!
    finalCharge:   CategorizedTime!
```

And are geared toward supporting a view such as this one in the inception doc, where each entry for an observation corresponds to a visit.

Screen Shot 2023-11-23 at 2 35 10 PM


build.sbt
```diff
@@ -61,6 +61,7 @@ lazy val schema =
         "io.circe"       %%% "circe-testing"                   % circeVersion               % Test,
         "edu.gemini"     %%% "lucuma-core"                     % lucumaCoreVersion,
         "edu.gemini"     %%% "lucuma-core-testkit"             % lucumaCoreVersion          % Test,
+        "edu.gemini"     %%% "lucuma-sso-frontend-client"      % lucumaSsoVersion,
```
swalker2m, line 64: I added this just to pick up the circe codec for `User` so that I could in turn provide a codec for `TimeChargeCorrection`.  Is this ill advised?  Maybe I should just supply a mapping.

tpolecat: I think this is ok. We could factor out another tiny module with just the codec at some point. Because the codec is used for encoding/decoding the JWT payload it has to be very stable and has to be owned by SSO I think.

tpolecat (approved):
👍🏻 This was a lot of not-fun work. Looks great.

PR #842, opened 2023-11-24
Make digest and config optional

swalker2m:
The `ExecutionDigest` and `ExecutionConfig` are currently required with a bang `!`. If the observation is not fully specified, these fail because we cannot generate its sequence.  This PR makes `execution / digest`  and `execution / config` both optional, while preserving any errors in the top-level error channel.  Is this a good idea?

  For example the query

```
             query {
               program(programId: "p-123") {
                 observations {
                   matches {
                     id
                     execution {
                       digest {
                         setup {
                           full { seconds }
                         }
                       }
                     }
                   }
                 }
               }
             }
```

might produce

 ```
            {
              "program": {
                "observations": {
                  "matches": [
                    {
                      "id": "o-123",
                      "execution": {
                        "digest": null
                      }
                    },
                    {
                      "id": "o-124",
                      "execution": {
                        "digest": {
                          "setup" : {
                            "full" : {
                              "seconds" : 960.000000
                            }
                          }
                        }
                      }
                    }
                  ]
                }
              }
            }
```

along with the error

```
Could not generate a sequence from the observation o-123: observing mode
```

### DIFF
--- a/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -5183,11 +5183,15 @@ type Execution {
 
   """
   Calculations dependent on the sequence, such as planned time and offsets.
+  If a sequence cannot be generated for this observation, `null` is returned
+  along with warning messages.
   """
-  digest: ExecutionDigest!
+  digest: ExecutionDigest
 
   """
-  Full execution config, including acquisition and science sequences.
+  Full execution config, including acquisition and science sequences.  If a
+  sequence cannot be generated for this observation, `null` is returned along
+  with warning messages.
   """
   config(
 
@@ -5201,7 +5205,7 @@ type Execution {
     """
     futureLimit: NonNegInt = 25
 
-  ): ExecutionConfig!
+  ): ExecutionConfig
 
   "Executed (or at least partially executed) atom records, across all visits."
   atomRecords(
--- a/modules/service/src/main/scala/lucuma/odb/graphql/mapping/ExecutionMapping.scala
+++ b/modules/service/src/main/scala/lucuma/odb/graphql/mapping/ExecutionMapping.scala
@@ -103,7 +103,7 @@ trait ExecutionMapping[F[_]] extends ObservationEffectHandler[F]
 
   extension (e: Generator.Error) {
     def toResult: Result[Json] =
-      Result.failure(e.format)
+      Result.warning(e.format, Json.Null)
   }
 
   private lazy val configHandler: EffectHandler[F] = {
--- a/modules/service/src/test/scala/lucuma/odb/graphql/OdbSuite.scala
+++ b/modules/service/src/test/scala/lucuma/odb/graphql/OdbSuite.scala
@@ -3,6 +3,7 @@
 
 package lucuma.odb.graphql
 
+import cats.data.Ior
 import cats.effect.*
 import cats.effect.std.Supervisor
 import cats.effect.unsafe.IORuntime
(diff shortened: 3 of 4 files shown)


hugo-vrijswijk (commented):
The API looks good to me. I'll leave reviewing the implementation to someone else

tpolecat:
Do we actually get success with warnings from the Clue front end? It looks like you're catching exceptions in the testcase which makes me think the end result for users would be a failure ... ?

toddburnside:
> Do we actually get success with warnings from the Clue client? It looks like you're catching exceptions in the testcase which makes me think the end result for users would be a failure. Maybe I handle that in graphql-routes. A ver..
> 
> Ok I was confused. It's up to clients to handle this case or not. Seems fine.

I think the clue client uses policy classes for this. By default, IIRC, this would be a failure, but specifying a policy class would allow handling this situation. Of course, @rpiaggio would know better.

swalker2m:
It's legitimate GraphQL maybe but is it preferable to instead raise warnings to the level of the schema instead of stuffing them in the error message array?  That is, define some object to hold either a valid result or else an error message.

toddburnside:
It seems the same thing should be done for the `guideEnvironments` and `guideAvailability`. 

Actually, at one point the execution stuff was in it's own base level query when Grackle didn't support having it in the Observation in the schema. I sometimes wonder if that wouldn't be better for the guide star stuff (and maybe execution stuff). The problem I can see is if they are in the observation, the user can include them in a query for all observations as in the example in the comment. Since these effects take time to run, it seems we could easily hit problems with timeouts. I'm not sure if this is a problem with the execution digest, but calculating the guide stars for a bunch of observations can take significant time.

swalker2m:
> Yeah if you expect the client to handle it then it might make sense to encode it in the schema. If you just expect the client to say "the operation reported warnings: X, Y, Z" then the error channel is ok. So maybe it depends on what Explore is likely to do here?

This was a request from @hugo-vrijswijk so perhaps he can confirm, but I think the idea is to ignore the `null` results (and corresponding warnings) and gather the valid ones.  I was hesitant to make it optional w/o any kind of feedback on why it fails to compute sometimes so I tried this approach.

@toddburnside is right that if you have many observations it may be an expensive query and presumably may time out.  I don't know what we can do about that other than narrow the result set in the `observations` query.

hugo-vrijswijk:
> This was a request from @hugo-vrijswijk so perhaps he can confirm, but I think the idea is to ignore the null results (and corresponding warnings) and gather the valid ones. 

Correct, I figured logging the warnings would be fine. As it is a valid case for an observation not to have this data (yet).

> The problem I can see is if they are in the observation, the user can include them in a query for all observations as in the example in the comment. Since these effects take time to run, it seems we could easily hit problems with timeouts. I'm not sure if this is a problem with the execution digest, but calculating the guide stars for a bunch of observations can take significant time.

I see your point for calculating each one taking a lot of time in the query for all observations. But we do need this data in the observation list in explore. And in the program tab to sum them. I suppose we could load them lazily for each observation in view if performance is a concern. I can't say what the implication would be for including the digest in the query for all observations

swalker2m:
> I see your point for calculating each one taking a lot of time in the query for all observations. But we do need this data in the observation list in explore. And in the program tab to sum them. I suppose we could load them lazily for each observation in view if performance is a concern. I can't say what the implication would be for including the digest in the query for all observations

The digest is cached so it would be an initial call to calculate them all that will suffer.  (This will happen when we push an update to the code as well since it includes the commit in the input hash.)

Note that at the program level, there is a `plannedTimeRange` that sums the observations considering groups.  Of course it has to do the underlying calculations for each valid observation so it could also take a while.

I'll go ahead and merge this since I think Hugo needs it.

PR #848, opened 2023-11-28
Default Proposal ToO Activation to None.

toddburnside:
The `ToOActivation` field in the Proposal was optional in the API query. Although the GraphQL documentation stated that it was mandatory on creation, it was nullable in the database. Furthermore, it is not optional in the `Proposal` class in lucuma core, which is what Explore uses. This PR makes it not nullable in the database and defaults it to `none` which is a reasonable default value.


tpolecat (approved):
:+1:

PR #855, opened 2023-12-01
Fix constraints for the proposal class.

toddburnside:
The constraints for  `c_min_percent_total` and `c_total_time` checked the values of the `c_category` problem instead of the `c_class` column, leading to odd constraints that weren't caught by the tests.

This changes the constraints and also adds/updates some tests that would have failed with the previous constraints. 

In addition, adding a new test caused the `edit proposal (non-class properties` test to fail because the order of the partner splits changed, so an elaborator was added to make the order of the partner splits deterministic.


tpolecat (approved):
👍🏻 good catches here. Thanks!

PR #859, opened 2023-12-04
TimeChargeCorrectionInput

swalker2m:
I mistakenly used a `type` in an `input` in my [last PR](https://github.com/gemini-hlsw/lucuma-odb/pull/841).  This was an API update and necessary changes to keep the existing code working, but the new API hasn't been implemented so I didn't notice.  Unfortunately it breaks introspection as @toddburnside discovered.

### DIFF
--- a/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -64,7 +64,7 @@ that will be corrected and describes the correction itself.
 """
 input AddTimeChargeCorrectionInput {
   observation: VisitId!
-  correction:  TimeChargeCorrection!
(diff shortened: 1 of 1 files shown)


swalker2m:
Would it be possible to add a sanity check test case that just tries to do introspection?

tpolecat:
> Would it be possible to add a sanity check test case that just tries to do introspection?

Yes, that's a good idea. Introspection is just a user-level query, although I don't know what it looks like offhand. The playground sends it of course.

tpolecat:
The playground gets its introspection query from here https://github.com/graphql/graphql-js/blob/688ee2f2a2f938e80bcf3808bc435d3d87a3ff3e/src/utilities/getIntrospectionQuery.ts#L41

swalker2m:
I might circle back to this to add a sanity check, but I'll merge this to fix the playground now.

PR #869, opened 2023-12-08
Time Accounting (Part 1)

swalker2m:
This PR is an initial step in time accounting. It includes only execution time for individual visits along with:

* A `TimeAccounting` object containing a few simple case classes for tracking time accounting information.  This is only used in the implementation.
* A `TimeAccountingState` which is like a union of `TimestampInterval` but where each interval is associated with a `TimeAccounting.Context`.  It is limited to functionality supporting time accounting calculations and is intended to be used as the state of a `cats.data.State` as various discounts are applied.
* A `TimeAccountingService` which is intended to be what the mappings will interact with to get the time accounting results.  The `VisitMapping` uses it here.


modules/service/src/main/scala/lucuma/odb/graphql/mapping/VisitMapping.scala
```diff
@@ -126,4 +130,44 @@ trait VisitMapping[F[_]] extends VisitTable[F]
       selectWithOffsetAndLimit(rOFFSET, rLIMIT, ExecutionEventType, "id", Predicates.executionEvent.id, Predicates.executionEvent.observation.program)
   }
 
+  // WIP.  At the moment the `invoiceHandler` calls the time accounting service
+  // on demand and returns an "invoice" with just the whole execution time.
+  private lazy val invoiceHandler: EffectHandler[F] =
```
swalker2m, line 135: While this is done on demand at the moment, ultimately the results will be stored (per visit).  In addition to making it a fast/simple lookup for executed visits, if the rules change after the fact then the change won't impact past calculations. Stability is important because otherwise changing the implementation invalidates any manual corrections which may have been applied.

swalker2m:
Alright, I updated this to lookup an `ObserveClass` to use as the default instead of tracking intervals with no corresponding step as "uncategorized".  This sands the edge a little bit.   I'll mark it ready now.

Keep in mind, this is part of a series of time accounting PRs and it may take a circuitous path to get to the final destination.

modules/service/src/main/scala/lucuma/odb/service/TimeAccountingService.scala
```diff
@@ -0,0 +1,115 @@
...
+      val event: Decoder[TimeAccounting.Event] =
+        (core_timestamp *: context).to[TimeAccounting.Event]
+    }
+
+    val SelectEvents: Query[(ObserveClass, Visit.Id), TimeAccounting.Event] =
+      sql"""
+        SELECT
+          e.c_received,
+          e.c_visit_id,
+          COALESCE(s.c_observe_class, $obs_class),
```
toddburnside, line 101: OK. So...The `MIN` observe class for the observation is looked up in `SelectObserveClass` and defaults to `science`. Then, here, if the observe class column is null, we use the one from `SelectObserveClass`?

toddburnside (approved):
LGTM

modules/service/src/main/scala/lucuma/odb/service/TimeAccountingService.scala
swalker2m: Yes, that's the intent.  The sequence level events are not tied to any step so for those there is no explicitly specified observe class.  In the `ocs` at any rate, in that situation we charge according to the observe class for the observation as a whole.

tpolecat (approved):
This looks very good. I wish there were a better word than "context" but I can't think of one. 👍🏻

PR #872, opened 2023-12-11
Switch obs status used for time calculations from Approve to Included

toddburnside:
This is based on a discussion with Andy. The proposal needs the time for observations in the `Included` status and above.


swalker2m (approved):
:+1:

PR #874, opened 2023-12-12
Remove `description` and `checked` from ProposalAttachments

toddburnside:
This removes the `description` and `checked` columns from the DB and the GraphQL API. The mutation was completely removed because only `description` and `checked` were able to be modified. The Rest API was updated to remove the description.

I `think` I found all the little bits that need to be removed...


tpolecat (approved):
deletion good 👍🏻

PR #885, opened 2023-12-15
Store time accounting

swalker2m:
Adds storage of time accounting results.  This works by adding a `t_time_accounting` table with columns for raw and final time charges in each category.  When a visit is created, an entry for it is added to the time accounting table.  Each time a new event arrives, the time accounting is recalculated and updated in the table.

My original plan was to calculate the values on demand, after first checking whether the value had been previously stored.  Each new event would then just remove the time accounting row from the table.  This was to be done with an `EffectField` which did the check/calculation and then proceeded normally as if it were a simple `SqlObject` mapping.  Unfortunately I couldn't get this to work with Grackle.

I'm a little worried about repeatedly calculating time accounting on each event reception, especially as it becomes more costly to compute, but at this point I don't have a better idea.


modules/service/src/main/scala/lucuma/odb/graphql/mapping/VisitMapping.scala
```diff
@@ -129,41 +116,4 @@ trait VisitMapping[F[_]] extends VisitTable[F]
...
-                  }
-          res <- ResultT(
-                   vids
-                     .flatMap { vid => invs.find(r => r._1 === vid).map(_._2).toList }
-                     .zip(queries)
-                     .traverse { case (result, (query, parentCursor)) =>
-                       Query.childContext(parentCursor.context, query).map { childContext =>
-                         CirceCursor(childContext, result, Some(parentCursor), parentCursor.fullEnv)
-                       }
-                     }.pure[F]
```
swalker2m, line 164: With the time accounting data that is currently ready, this `CirceCursor` / `Json` result approach would work fine. Once the discounts and corrections are added though, there will be fields that require database lookups depending on the query.

swalker2m:
I think if I modify the time accounting part of the schema to avoid references (e.g., the associated `Visit`, the `Atom`s that were impacted by a discount, the `Dataset`s with a poor QA state, etc.) I could pretty easily go back to the original idea.  That is, 

* delete the time accounting data when a new event is received
* make an `EffectField` that reads the time accounting data if it exists and returns `Json`.  otherwise it does the calculation and writes it and then returns `Json`

Maybe I should do that?  It may be kind of a lot of `Json` though.  I think doing it correctly (i.e., making the `EffectField` do the check/update and then continue as if it were a simple `SqlObject` mapping) is probably trivial but I couldn't discover it.

swalker2m:
Closed in favor of #896

PR #895, opened 2023-12-27
Visit, Atom, Step, Dataset time interval

swalker2m:
Visits, atoms, steps and datasets are associated with execution events from Observe.  The first event in each case signifies the start of the visit, atom, step or dataset and the last event signifies the end.  The period between the start and the end is the duration.  These were represented in the schema, but inconsistently.  This PR makes the event-derived intervals uniform by adding a `TimestampInterval` to the schema and an `interval` field of this type to `Visit`, `AtomRecord`, `StepRecord` and `Dataset` replacing the existing `start`, `startTime`, etc.  In fact `AtomRecord` was missing the interval altogether so it was added there.

For example:

```
          observation(observationId: "o-123") {
            execution {
              visits {
                matches {
                  interval {
                    start
                    end
                    duration { seconds }
                  }
                }
              }
            }
          }
```


toddburnside (approved):
LGTM

PR #896, opened 2023-12-27
Time Accounting (Part II)

swalker2m:
Snapshot of the current state of the time accounting implementation.  This PR supersedes #885.  It will continue to grow until morale improves.

* Storage of time accounting data.
* Discount for daylight (pre-dusk and/or post-dawn).
* `TimeAccountingState` updates.

The storage implementation is the same as #885, in that it is updated upon event reception.  I expect this decision to be revisited.


swalker2m:
Superseded by #900.

PR #898, opened 2023-12-28
Add ProposalStatus, make `Enums` available in `Services`

toddburnside:
I added the dynamic Enums to the `Services` and made them available with all the other syntax. However, usage is not completely straight forward. For one, see the comments in `Services.scala`. Also, I did not see anyway to get an instance of one of the Enums straight from the API, so I converted between `Tag` and `ProposalStatus` manually. Ditto for values from the DB. If someone has better ideas for dealing with the Enums, I'd be happy to hear them.


modules/service/src/main/scala/lucuma/odb/graphql/enums/Enums.scala
```diff
@@ -86,10 +87,33 @@ final class Enums(
 
   }
 
+  enum ProposalStatus(val tag: String, ordinal: Short) {
+    private val meta = enumMeta.proposalStatus.getOrElse((tag, ordinal), sys.error(s"t_proposal_status missing tag `$tag`, ordinal `$ordinal`"))
+
+    def name: String = meta.name
+
+    // Note: The order here controls the Order of the Enumerated in the ODB, so keep them in order.
```
toddburnside, line 95: All the existing dynamic enums get ordered by the `tag` in the API. I wanted a consistent, non-tag order between API and clients so I added the ordinal. Couldn't think of a better way.

modules/service/src/main/scala/lucuma/odb/graphql/input/ProgramPropertiesInput.scala
```diff
@@ -7,19 +7,21 @@ package input
...
 import lucuma.odb.data.Existence
+import lucuma.odb.data.Tag
 import lucuma.odb.graphql.binding._
 
 object ProgramPropertiesInput {
 
   case class Create(
     name: Option[NonEmptyString],
-    proposal: Option[ProposalInput.Create],
-    existence: Option[Existence]
```
toddburnside, line 18: `existence` is not actually used in creation. So, I removed it from here. Users can still specify it in the API, but it is ignored, just as it was before. Since only half the fields in `ProgramPropertiesInput` are used for creation, it makes me wonder if creation and update should have different `Inputs`?

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -7907,6 +7927,11 @@ type Query {
   """
   proposalAttachmentTypeMeta: [ProposalAttachmentTypeMeta!]!
 
+  """
+  Metadata for `enum ProposalStatus
+  """
+  proposalStatusMeta: [ProposalStatusMeta!]!
```
swalker2m, line 7933: I assume there will be quite a few of these before we're done.  Maybe they should be separated into a `EnumMetaType` or something in order to keep the individual entries out of the top level?

modules/service/src/main/scala/lucuma/odb/graphql/input/ProgramPropertiesInput.scala
swalker2m: I'm confused.  `ProgramPropertiesInput` has

```
  name: NonEmptyString
  proposal: ProposalInput
  existence: Existence
```

We don't use these for program creation?  Is it just that we haven't implemented it yet rather than that there's a fundamental reason why they don't make sense at creation time?

toddburnside: `ProgramPropertiesInput` now also has `proposalStatus`. AFAICT, `existence`, although available in `ProgramPropertiesInput.Create` has always just been ignored. I assume because it makes no sense to create a `deleted` program. 
It also makes no sense to create a program with a proposal status of anything except `NOT_SUBMITTED` since it cannot have any observations, yet, having just been created.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
toddburnside: Seems like a reasonable idea. The rest api that provides the enum metadata would just need to be changed to use the new path. Explore (and probably the other UIs) don't care about it directly in the GraphQL API because they get the metadata through the rest api to create the dynamic enums.

modules/service/src/main/scala/lucuma/odb/graphql/input/ProgramPropertiesInput.scala
swalker2m: `Existence` is `PRESENT` vs `DELETED` but maybe should have been `PRESENT` vs `HIDDEN` or something since it is not really deleted.  I could _maybe_ see creating a hidden program and editing it until ready and then making it appear by "undeleting" it.

toddburnside: I suppose they could do that via the API (but isn't support in Explore at the moment). If we want to allow that, I could put it back and actually implement it in program creation. Since it wasn't actually being used, I thought it was clearer to remove it from the Create case class.

toddburnside: However, it still don't think it makes sense to create one with a non NOT_SUBMITTED proposal class. If it is SUBMITTED, it will be `locked` and will have no observations.

modules/service/src/main/scala/lucuma/odb/graphql/enums/Enums.scala
tpolecat: You could prepend a number to the tag since the tags aren't supposed to be for human consumption, but I guess we have kind of violated that idea already by exposing the tag as the enum value :-\

So this seems ok.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/QueryMapping.scala
```diff
@@ -120,6 +122,12 @@ trait QueryMapping[F[_]] extends Predicates[F] {
         OrderBy(OrderSelections(List(OrderSelection[Tag](ProposalAttachmentTypeMetaType / "tag"))), child)
       }
 
+  private lazy val ProposalStatusMeta: PartialFunction[(TypeRef, String, List[Binding]), Elab[Unit]] =
+    case (QueryType, "proposalStatusMeta", Nil) =>
+      Elab.transformChild { child =>
+        OrderBy(OrderSelections(List(OrderSelection[Short](ProposalStatusMetaType / "ordinal"))), child)
```
tpolecat, line 128: At some point we should introduce a helper to make this easier.

modules/service/src/main/scala/lucuma/odb/service/ProgramService.scala
```diff
@@ -162,13 +179,60 @@ object ProgramService {
           }
 
         // Update proposals. This can fail in a few ways.
-        val updateProposals: F[List[Program.Id]] =
-          SET.proposal.fold(Nil.pure[F]) {
-            proposalService.updateProposals(_)
+        val updateProposals: F[Either[UpdateProgramsError, List[Program.Id]]] =
```
tpolecat, line 182: At some point we should really change this to return `F[Result[List[Program.Id]]]` since all we do with the structured error is turn it into a `Problem` up in the mutation. It wasn't initially clear what kind of values the services needed to compute.

modules/service/src/main/scala/lucuma/odb/service/ProgramService.scala
```diff
@@ -162,13 +179,60 @@ object ProgramService {
...
 
-        // Combie the results
-        (setup >> updatePrograms, updateProposals).mapN(_ |+| _)
+        // A stable identifier (ie. a `val`) is needed for the enums.
+        val enumsVal = enums
+
+        def tagToProposalStatus(tag: Tag): Either[UpdateProgramsError, enumsVal.ProposalStatus] =
+          Enumerated[enumsVal.ProposalStatus].fromTag(tag.value).toRight(UpdateProgramsError.InvalidProposalStatus(tag))
+
+        def userCanChangeProposalStatus(ps: enumsVal.ProposalStatus): Option[Unit] =
```
tpolecat, line 198: Would it complicate things too much for this to be `Boolean`?

tpolecat (approved):
A few comments but this seems fine. 👍🏻

modules/service/src/main/scala/lucuma/odb/service/ProgramService.scala
toddburnside: Ah. OK. I'll change that. I had initially started in that direction, but changed when the services I was using for inspiration weren't return `Result`s. I thought there might be some reason for separating the types that was beyond my ken.

modules/service/src/main/scala/lucuma/odb/service/ProgramService.scala
toddburnside: Nope. I'm not sure why I did it that way. Probably distracted by other issues. 😄

modules/service/src/main/scala/lucuma/odb/service/ProgramService.scala
tpolecat: Originally I had wanted to keep the service layer independent of the GraphQL layer (`Result` is a Grackle data type) but it resulted in a lot of indirection that was probably worse than just using `Result`. I don't have strong feelings about it anymore.

PR #900, opened 2023-12-29
Time Accounting (Part III)

swalker2m:
Snapshot of the current state of the time accounting implementation. This PR supersedes #896 which in turn supersedes #885. It will continue to grow until morale improves.

* Storage of time accounting data.
* Discount for daylight (pre-dusk and/or post-dawn).
* Discount for QA state.
* TimeAccountingState updates.

The storage implementation is the same as #885, in that it is updated upon event reception. I expect this decision to be revisited.


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -9427,54 +9440,67 @@ uncharged.
...
+  program: TimeSpan!
 
-  "Optional additional detail."
-  comment: String
+  "Additional detail."
+  comment: String!
 
-  "Datasets with non-passing QA states."
+  "Datasets associated with the discount."
   datasets: [Dataset!]!
```
swalker2m, line 9496: I tried to make this into a query with offset and limit such as appears elsewhere but was soundly defeated by Grackle.  The only difference that I could appreciate from other cases (such as in the `VisitMapping`) was the usage of an intermediate link table between discounts and datasets.

swalker2m:
Closed by #905.

PR #905, opened 2024-01-04
Time Accounting (Part 4)

swalker2m:
Current state of time accounting, now including:

* Storage of time accounting data
* Discount for daylight (pre-dusk and/or post-dawn)
* Discount for QA state
* TimeAccountingState updates
* Scanty testing

Incorporates #900 #896 #885.

PR #908, opened 2024-01-08
Time Accounting (Part 5)

swalker2m:
Current state of time accounting, now including:

* Storage of time accounting data
* Discount for daylight (pre-dusk and/or post-dawn)
* Discount for QA state
* TimeAccountingState updates
* Observation and Program time charge sum
* Scanty testing

Incorporates #905 #900 #896 #885.


modules/service/src/main/scala/lucuma/odb/service/TimeAccountingService.scala
```diff
@@ -3,59 +3,328 @@
...
+          _  <- session.execute(PerformCorrection)(visitId, input).void
+        } yield id
+
+      def selectObservation(
+        observationId: Observation.Id
+      )(using Transaction[F]): F[CategorizedTime] =
+        session
+          .unique(SelectObservation)(observationId)
+          .recoverWith { case SqlState.DatetimeFieldOverflow(_) =>
+            TimeAccounting.CategorizedTimeMax.pure[F]
```
swalker2m, line 243: This seems sketchy in that if any of non-charged, partner or program sums overflows they are all set to the maximum time span.  I couldn't figure away around it though other than performing the query individually for each category (i.e., three times instead of once).  This is a case that in reality doesn't occur except deliberately or perhaps via a scripting error by an API user or maybe a bug in the client.

swalker2m:
Superseded by #909

PR #912, opened 2024-01-10
Step QA State

swalker2m:
Adds a missing field definition, for overall step QA state.  It is a rollup of all the step's dataset's QA states.  There was a specific, `StepQaState` type for this but I couldn't remember why.  I removed it and just return the worst `DatasetQaState` for the step.  For example, if any dataset is marked `FAIL` the step QA state will also be `FAIL`.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/graphql/mapping/StepRecordMapping.scala
+++ b/modules/service/src/main/scala/lucuma/odb/graphql/mapping/StepRecordMapping.scala
@@ -5,17 +5,27 @@ package lucuma.odb.graphql
 package mapping
 
 import cats.effect.Resource
+import cats.syntax.applicative.*
+import cats.syntax.eq.*
+import cats.syntax.functor.*
 import cats.syntax.option.*
+import cats.syntax.traverse.*
 import grackle.Cursor
 import grackle.Predicate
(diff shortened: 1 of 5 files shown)


modules/service/src/main/scala/lucuma/odb/graphql/mapping/StepRecordMapping.scala
```diff
@@ -53,9 +63,7 @@ trait StepRecordMapping[F[_]] extends StepRecordView[F]
         EffectField("interval",  intervalHandler, List("id")),
         SqlObject("stepConfig"),
         SqlField("observeClass", StepRecordView.ObserveClass),
-
-        // TBD: stepQaState
-
+        EffectField("qaState", qaStateHandler, List("id")),
```
swalker2m, line 66: I think it likely possible to do this without resorting to an `EffectField` but was stymied by the fact that the data it needs is not found in a simple `Leaf`, like for example instances where a `CirceFieldJson` is used.  Also, the underlying query is trivial, though the `EffectHandler` trappings are convoluted.

tpolecat (approved):
This looks good. It's annoying that we're working around so many limitations of Grackle. Hopefully these things will improve and we can remove some of this complexity.

PR #914, opened 2024-01-11
Fix t_chron_program_update migration

toddburnside:


swalker2m (approved):
Looks good, assuming the update was never successfully applied and Flyway permits it.

PR #915, opened 2024-01-11
user invitations

tpolecat:
Kind of a long discussion here, sorry. The feature was never fleshed out in a lot of detail so I wanted to record my thinking.

This adds an API to allow users to create, redeem, and revoke invitations to participate in a particular science program. Redeeming an invitation performs a `linkUser` operation, adding the current user to the target program in a specified program role.

The workflow is as follows:
- `createUserInvitation` creates a `pending` invitation for a specified program and returns a single-use key, which is not stored (we only know its hash). Its structure is the same as an API key; i.e., `<id>.<body>`. The invitation itself records the issuer, redeemer (if any), associated program, and state (`pending`, `redeemed`, `declined`, `revoked`). A list of invitations is exposed on `type Program`.
  - Users acting in a `Guest` role cannot create invitations.
  - Users acting in a `Pi` (science) role can create invitations for programs where they are the PI, but only for `coi` and `observer` (onlooker) program roles.
  - Users acting in an `Ngo` role can create invitations for programs where their partner has allocated time, but only for the NGO support program role, and only for their own partner.
  - Users acting in a `Service`, `Admin`, or `Staff` role can create any kind of invitation on any program, including staff support.
- `redeemUserInvitation` redeems an invitation, given a single-use key. Invitations can only be redeemed if they are currently `pending`. A user may also use this operation to `decline` an invitation.
  - `Guest` and `Service` users cannot redeem invitations (program roles apply only to `StandardUser`s).
- `revokeUserInvitation` revokes an `pending` invitation by id (the key is not needed)
  - Users acting in a `Guest` role cannot revoke invitations.
  - Users acting in a `Pi` or `Ngo` role can revoke pending invitations they issued.
  - Users acting in a `Service`, `Admin`, or `Staff` role can revoke any pending invitation.

There is no UI for this yet. The ultimate intent is to be able to send an invitation by email and have a mechanism to redeem it via clicking a link. As a first step it might make sense to just have a dialog for creating an invitation and another for redeeming one.

Because there is no intended recipient associated with the invitation, there's a security risk if the invitation is inadvertently shared. We *could* require (or allow) the user to specify the intended recipient's ORCiD Id, but we need to allow for people who have not even registered with ORCiD yet. Not sure what makes sense here.


swalker2m:
> We could require (or allow) the user to specify the intended recipient's ORCiD Id, but we need to allow for people who have not even registered with ORCiD yet.

To make sure I understand:

At the point that the intended recipient decides to accept the invitation, they will have to first register with ORCiD?

The single-use key is only good for creating the link between the new user and the program?  Once that is accomplished it may remain recorded for posterity but serves no authentication purpose.

tpolecat:
> At the point that the intended recipient decides to accept the invitation, they will have to first register with ORCiD?

Right. If an invitation is intended for a specific ORCID id then we can ensure that *only* that person can redeem it. But if the intended recipient isn't registered we have no way to tie them to the invitation, except possibly by email, but this is sketchy because they might not expose the same email address in their ORDID profile, which is all we know. So in that case we will necessarily need to have an invitation out in the wild that anyone could redeem.
 
> The single-use key is only good for creating the link between the new user and the program? Once that is accomplished it may remain recorded for posterity but serves no authentication purpose.

That's correct. They hang around for posterity but can only be used once. `redeemInvitation` and `revokeInvitation` permanently set the invitation into an immutable state.

tpolecat:
An invitation is kind of like a program key in OCS. Maybe I should explain it like that.

modules/service/src/main/scala/lucuma/odb/service/UserInvitationService.scala
```diff
@@ -0,0 +1,212 @@
...
+          case Access.Admin | Access.Service | Access.Staff =>
+            session.prepareR(Statements.revokeUserInvitationUnconditionially).use: pq =>
+              pq.option(input.id).map:
+                case Some(id) => Result(id)
+                case None     => Result.failure(s"Invitation does not exist or is no longer pending.")
+          case Access.Ngo | Access.Pi =>
+            session.prepareR(Statements.revokeUserInvitation).use: pq =>
+              pq.option(input.id, user.id).map:
+                case Some(id) => Result(id)
+                case None     => Result.failure(s"Invitation does not exist, is no longer pending, or was issued by someone else.")
```
swalker2m, line 130: Grackle should maybe have some `Option` to `Result` `.toSuccess` thing like the `.toRight`.

swalker2m (approved):
This looks great, very clear.

modules/service/src/main/resources/db/migration/V0842__invitations.sql
```diff
@@ -0,0 +1,57 @@
...
+  p_program_id      d_program_id,                
+  p_role            e_program_user_role,         
+  p_support_type    e_program_user_support_type, 
+  p_support_partner d_tag                  
+) RETURNS text AS $$
+  DECLARE
+    invitation_key_id d_invitation_id := to_hex(nextval('s_invitation_id'::regclass));
+    invitation_key text := md5(random()::text) || md5(random()::text) ||  md5(random()::text);
+    invitation_key_hash text := md5(invitation_key);
+    rec record;
```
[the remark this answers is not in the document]
tpolecat: Ah, good catch, that's unused.

modules/service/src/main/scala/lucuma/odb/service/UserInvitationService.scala
swalker2m: Never mind, I guess that is there in `grackle.syntax`.  It wouldn't really make any difference to employ it though.

PR #918, opened 2024-01-12
Query -> events

swalker2m:
`ExecutionEvent` top-level query.  It can match on observation id, step id, dataset id, timestamp, etc. but is hampered by lack of interface support in Grackle.  I could not, for example, add a predicate for dataset filename for dataset events because those would require matching on `DatasetEvent` in particular or perhaps adding a hidden `SqlObject` (which is not supported).

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/graphql/input/WhereExecutionEvent.scala
+++ b/modules/service/src/main/scala/lucuma/odb/graphql/input/WhereExecutionEvent.scala
@@ -1,0 +1,77 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.odb.graphql
+package input
+
+import cats.syntax.parallel.*
+import grackle.Path
+import grackle.Predicate
+import grackle.Predicate.*
+import lucuma.core.enums.DatasetStage
+import lucuma.core.enums.SequenceCommand
+import lucuma.core.enums.StepStage
+import lucuma.core.model.ExecutionEvent
+import lucuma.core.model.Observation
+import lucuma.core.model.Visit
+import lucuma.core.model.sequence.Dataset
+import lucuma.core.model.sequence.Step
+import lucuma.core.util.Timestamp
+import lucuma.odb.data.ExecutionEventType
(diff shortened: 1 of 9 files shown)


modules/service/src/main/scala/lucuma/odb/graphql/input/WhereExecutionEvent.scala
```diff
@@ -0,0 +1,66 @@
...
+
+object WhereExecutionEvent {
+
+  def binding(path: Path): Matcher[Predicate] = {
+    val WhereOrderExecutionEventIdBinding = WhereOrder.binding[ExecutionEvent.Id](path / "id", ExecutionEventIdBinding)
+    val WhereEqVisitIdBinding             = WhereEq.binding[Visit.Id](path / "visit" / "id", VisitIdBinding)
+    val WhereOrderObservationIdBinding    = WhereOrder.binding[Observation.Id](path / "observation" / "id", ObservationIdBinding)
+    val WhereOrderTimestampBinding        = WhereOrder.binding[Timestamp](path / "received", TimestampBinding)
+    val WhereExecutionEventTypeBinding    = WhereEq.binding[ExecutionEventType](path / "eventType", enumeratedBinding[ExecutionEventType])
+//    val WhereSequenceEventBinding         = WhereSequenceEvent.binding(path)
```
swalker2m, line 26: When I attempt this I'm told:

```
java.lang.Throwable: No mapping for field 'command' of type [ExecutionEvent]
	at grackle.Result$.internalError(result.scala:203)
	at grackle.sql.SqlMappingLike.columnsForLeaf(SqlMapping.scala:891)
	at grackle.sql.SqlMappingLike.columnsForLeaf$(SqlMapping.scala:37)
	at grackle.skunk.SkunkMapping.columnsForLeaf(SkunkMapping.scala:33)
	at grackle.sql.SqlMappingLike$MappedQuery$.loop$11(SqlMapping.scala:2937)
	at grackle.sql.SqlMappingLike$MappedQuery$.group$1$$anonfun$1(SqlMapping.scala:2805)
```

which is certainly true.  I attempted to use a `path` rooted at the `SequenceEvent` type instead of `ExecutionEvent` for this part but no luck.  I need some kind of `Predicate` that will match when the `ExecutionEvent` is a `SequenceEvent`.

tpolecat (approved):
This looks good. 👍🏻

PR #921, opened 2024-01-15
Short Cut 2429: Unify updateObservations and updateTargets inputs

swalker2m:
This fulfills a request from Andy to "make the input for updating observations and updating targets more similar, specifically either removing the requirement for a `ProgramId` or requiring it in both".  See [ShortCut Issue 2429](https://app.shortcut.com/lucuma/story/2429/normalize-updateobservationsinput-and-updatetargetsinput).

I'm not sure why I required the `ProgramId` for `updateObservations`.  There may have been a good reason (?) or it may more likely have simply been dumb.  At any rate, it feels onerous to require that so I removed it to match `updateTargets`.

### DIFF
--- a/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -10429,6 +10429,11 @@ input WhereObservation {
   id: WhereOrderObservationId
 
   """
+  Matches the id of the associated program.
+  """
+  programId: WhereOrderProgramId
+
+  """
   Matches the subtitle of the observation.
   """
   subtitle: WhereOptionString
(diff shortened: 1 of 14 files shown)


swalker2m:
I suppose this will break Explore, which is a favorite pastime of mine.  Should I continue to accept `Program.Id` and just ignore it?

toddburnside:
> I suppose this will break Explore, which is a favorite pastime of mine. Should I continue to accept `Program.Id` and just ignore it?

I suppose we could deprecate it. But, it's a fairly simple change. If you wait until tomorrow, we could use this to test Hugo's new github action to update lucuma-schemas. Explore dev will be broken for a little bit, but 🤷

toddburnside (approved):
LGTM

swalker2m:
>  If you wait until tomorrow

:+1:

tpolecat (approved):

PR #924, opened 2024-01-18
execution permissions

tpolecat:
This implements `ExecutionUserCheck#checkUser`, disallowing execution event writes from non-service users. Everything else is just adjustments in the tests to accommodate this.


modules/service/src/main/scala/lucuma/odb/graphql/OdbMapping.scala
```diff
@@ -38,8 +36,6 @@ import org.typelevel.log4cats.Logger
 import scala.io.AnsiColor
 import scala.io.Source
 
-import mapping.RedeemUserInvitationResultMapping
```
swalker2m, line 41: Hmm, if these are unused I wonder how this passed checks before.

swalker2m (approved):
Looks great, pending fetching an update to the test code I suppose.  :+1:

tpolecat:
Very confused. The file with the error on line 114 only has 106 lines. Running again.

PR #930, opened 2024-01-22
An alternate approach to enums

toddburnside:
This PR is meant as a topic for discussion.

The "dynamic enums" work well for enumerations that we think might change with some frequency in the future. However, they are more cumbersome to use, especially on the ODB side where we need to treat them as `Tag`s and convert to enumerations when desired. 

For enumerations that are unlikely to change, or will change infrequently, the value of the dynamic approach seems questionable. This PR presents a possible alternate solution to making sure that "static" enumerations defined in Scala match the values defined in a Postgres `enum` and in the GraphQL API. To me, this seems to be a simpler alternative to the dyamic implementation for enums like `ProposalStatus`, `ObsAttachmentType` and `ProposalAttachmentType`.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/graphql/enums/Enums.scala
+++ b/modules/service/src/main/scala/lucuma/odb/graphql/enums/Enums.scala
@@ -4,10 +4,9 @@
 package lucuma.odb.graphql.enums
 
 import cats.Monad
+import cats.MonadThrow
 import cats.data.NonEmptyList
-import cats.syntax.flatMap.*
-import cats.syntax.functor.*
-import cats.syntax.traverse.*
+import cats.syntax.all.*
 import grackle.DirectiveDef
 import grackle.EnumType
 import grackle.NamedType
@@ -15,11 +14,16 @@ import grackle.Schema
 import grackle.SchemaExtension
 import grackle.TypeExtension
 import lucuma.core.enums.Instrument
+import lucuma.core.enums.ToOActivation
 import lucuma.core.util.Enumerated
 import lucuma.core.util.TimeSpan
 import org.tpolecat.sourcepos.SourcePos
 import org.typelevel.log4cats.Logger
+import skunk.Codec
+import skunk.Fragment
 import skunk.Session
+import skunk.data.Arr
+import skunk.data.Type
 
 /**
  * Enums loaded from the database on startup.  These fall into two categories:
@@ -113,7 +117,7 @@ final class Enums(
       def pos: SourcePos = SourcePos.instance
       def baseTypes: List[NamedType] = 
         Enumerated[ProposalStatus].toEnumType("ProposalStatus", "Enumerated type of ProposalStatus")(_.name) ::
-          enumMeta.unreferencedTypes
+          enumMeta.unreferencedTypes ++ enumMeta.scalaEnumTypes
       def directives: List[DirectiveDef] = Nil
       def schemaExtensions: List[SchemaExtension] = Nil
       def typeExtensions: List[TypeExtension] = Nil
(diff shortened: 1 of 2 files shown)


modules/service/src/main/scala/lucuma/odb/graphql/enums/Enums.scala
```diff
@@ -122,15 +126,27 @@ final class Enums(
...
 object Enums {
+  case class ScalaEnum(
+    postgresName: String,
+    enumType: EnumType
+  )
+
+  // enums defined in scala for which we want to validate the postgres enums and add to the schema
+  val scalaEnums: List[ScalaEnum] =
+    List(
+      ScalaEnum("e_too_activation", Enumerated[ToOActivation].toEnumType("ToOActivation", "Target of Opportunity Activation")(_.label))
```
toddburnside, line 137: The description string ("Target of Opportunity Activiation") doesn't seem to be actually be used for anything?

toddburnside: However, we can add comments before the `DUMMY` entries in OdbSchema.graphql and they will be retained in the final schema.

cquiroz: What happens if the validation with postgres fails?

cquiroz:
Is the idea that enums thta change more often, e.g. `Filters` would remain in the db while those that don't e.g. `SIte` will remain as code?

toddburnside:
> Is the idea that enums thta change more often, e.g. `Filters` would remain in the db while those that don't e.g. `SIte` will remain as code?

Yes, that is what I was thinking.

And, Shane has informed me that `ToOActivation` may be a bad example. 😆 It appears that Bryan is poised to complicate the TOO process a bit.

swalker2m:
I like having the option, and working with actual Scala enums is much easier.  Seeing `Tag` in the code instead of actual enumeration values feels wrong.  Still, if there is much of a chance that an enum will be updated it will be problematic if we have to simultaneously update all the services and clients when a new value is added or removed.  I guess you're saying those remain dynamic though.

modules/service/src/main/scala/lucuma/odb/graphql/enums/Enums.scala
```diff
@@ -144,6 +160,19 @@ object Enums {
...
+  def validateScalaEnums[F[_]: MonadThrow: Logger](s: Session[F], scalaEnum: ScalaEnum): F[Unit] =
+      val _text_enum: Codec[List[String]] = 
+        Codec.array[String](_.toString, _.asRight[String], Type(s"_${scalaEnum.postgresName}", List(Type(scalaEnum.postgresName))))
+          .imap(_.toList)(l => Arr.fromFoldable(l))
+      // This is probably abusing skunk - is there a better way?
+      val query = Fragment(List(Left(s"select enum_range(null::${scalaEnum.postgresName})")), skunk.Void.codec, skunk.util.Origin.unknown)
+      val scalaValues = scalaEnum.enumType.enumValues.map(_.name.toLowerCase)
+      s.unique(query.query(_text_enum)).flatMap(l =>
+        // May need to allow for different ordering and conversions
+        if (l === scalaValues) Monad[F].unit
```
rpiaggio, line 174: `raiseUnless` ?

toddburnside: Nice.

modules/service/src/main/scala/lucuma/odb/graphql/enums/Enums.scala
toddburnside: The service doesn't start. Same thing as if there is a problem loading the dynamic enums.

rpiaggio:
I'm going to take a step back here and reevaluate our approach to dynamic enums. Now that we have them in place, they seem overly complicated to me. They require thinking of the instance and the meta level when programming. Furthermore, in several places we want our code to react to specific instances and referring to them is unsafe, we are assuming their existence and the compiler can't check that.

My opinion is to roll back dynamic enums. Let the source of truth be Scala enums. We are building web systems, in part, because they are easy and straightforward to redeploy. So, when we add an enum instance, we can press a button and redeploy everything. If we really want Postgres enums in the DB to represent them, we can generate migrations from the Scala enums.

Alternatively, we could have Postgres be the source of truth and generate Scala code from that, like we did before. But I'd rather have Scala be the source of truth for the simple reason that lucuma-odb is upstream in the dependency chain.

If we have an "enum" that changes frequently enough that it becomes tedious to redeploy because of it, then maybe it should be its own entity (eg: `*Filter`).

toddburnside:
I want to be clear that the purpose of this PR is to spur discussion. Like Raúl, I'm skeptical of the value of the dynamic enums given their complexity, but am amenable to having some of them dynamic if that it seems best. However, I'd prefer if we didn't use dynamic enums for everything. For example, all of the dynamic enums I have created (ProposalStatus, ObsAttachmentType and ProposalAttachmentType) don't really seem to need to be dynamic. I don't care if it is an approach similar to the one in this PR, or something different, but I'd like to be able to use scala enumerations where possible.

swalker2m:
FWIW I spent some time figuring out why this was evaluating `true` in some PL/pgSQL code ...

```
    IF NEW.c_proposal_status <> 'not_submited'::d_tag THEN
```

I think if it had been an `enum` instead it would have been obvious.

cquiroz:
What's the state of this discussion?

PR #939, opened 2024-01-30
Program reference

swalker2m:
Adds part of the [program reference proposal](https://gist.github.com/swalker2m/3b321fe96b195cb489b4954db66f3f44). There is more work to be done but I'll offer this in its current state since it is getting long.

## What's Present

- [x] The proposal part of Program Reference values for normal programs.  E.g., `G-2026A-0001`.
- [x] Updates to the API to accept `ProgramReference` in the `program` query.
- [x] Updates to the `WhereProgram` filter to accept program reference fields.

## What's Missing

- [ ] Additional queries that take a `ProposalId` need to accept a `ProposalReference` (e.g., `asterismGroup`).
- [ ] Support for the program type: `C`, `D` ... `Q` etc.
- [ ] Additional `ProgramReference` formats (`CAL`, `ENG`, `LIB`, `XPL` etc.)
- [ ] Observation reference, dataset reference etc.
- [ ] Push updates to `Semester` (and semester parsing) down to core.
- [ ] Move `ProgramReference` to core.


modules/service/src/main/scala/lucuma/odb/data/ProgramReference.scala
```diff
@@ -0,0 +1,100 @@
...
+import lucuma.core.model.Semester
+import lucuma.core.optics.Format
+import lucuma.core.parser.MiscParsers.intN
+import lucuma.core.parser.MiscParsers.posInt
+
+import java.time.DateTimeException
+import java.time.Year
+import scala.util.control.Exception
+
+case class ProgramReference(semester: Semester, index: PosInt) {
```
swalker2m, line 25: As mentioned, this is missing the program type and will probably become a `trait` with implementations for the various forms: science vs `CAL` / `ENG` / `LIB` / `XPL`.  Then I suppose it will go into `core` along with updates to `Semester` and its parsing.

modules/service/src/main/resources/db/migration/V0843__program_reference.sql
```diff
@@ -0,0 +1,90 @@
+-- Add a semester column to program.  It is nullable but there will be a
+-- constraint that checks that it is set when the proposal status is not
+-- 'not_submitted'.
+
+CREATE DOMAIN d_semester AS varchar
+  CHECK (VALUE ~ '^[0-9][0-9][0-9][0-9]+[AB]$');
```
tpolecat, line 6: Is there any practical reason this can't be `'^20[0-9][0-9][AB]$'`? I can't imagine any of this will still be running in 75 years.

modules/service/src/main/resources/db/migration/V0843__program_reference.sql
```diff
@@ -0,0 +1,90 @@
...
+RETURNS INT
+AS $$
+DECLARE
+    sequence_name VARCHAR := CONCAT('s_', semester);
+BEGIN
+    EXECUTE 'CREATE SEQUENCE IF NOT EXISTS ' || sequence_name;
+    RETURN nextval(sequence_name)::INT;
+END;
+$$ LANGUAGE plpgsql;
+
```
tpolecat, line 37: Oh, I like the idea of creating the sequences lazily. I would not have thought of this.

modules/service/src/main/resources/db/migration/V0843__program_reference.sql
```diff
@@ -0,0 +1,90 @@
...
+BEGIN
+    RETURN CASE
+        WHEN semester IS NULL OR index_value IS NULL THEN NULL
+        ELSE CONCAT('G-', semester, '-', LPAD(index_value::text, 4, '0'))
+    END;
+END;
+$$ LANGUAGE plpgsql IMMUTABLE;
+
+ALTER TABLE t_program
+  ADD COLUMN c_program_reference d_program_reference GENERATED ALWAYS AS (format_program_reference(c_semester, c_semester_index)) STORED UNIQUE;
```
tpolecat, line 90: 👍🏻

tpolecat (approved):
This looks very good.

modules/service/src/main/resources/db/migration/V0843__program_reference.sql
tpolecat: In fact I guess we could constrain the year to be 2024 through current year + 1, with a looser constraint on the Scala side.

swalker2m: Thanks, that's a good idea.  I'll make the lower limit be 2000 though since I think eventually we may want to import old data in some form?

tpolecat (approved):
Good 👍🏻

PR #942, opened 2024-01-31
Update Core and Grackle (Rename DatasetStage enum vals)

swalker2m:
Updates core and grackle.  This started as the ODB counterpart to https://github.com/gemini-hlsw/lucuma-core/pull/854 but then evolved into a general update the world PR.  Still, the biggest part is updating core for the new dataset event names as agreed upon with Carlos and Javier.


toddburnside (approved):
LGTM after updating with the real lucuma-core version.

swalker2m:
> after updating with the real lucuma-core version

I got a bit stuck trying to do this.  To update `core` in `odb`, I have to update it in `itc` and `sso`.  When I tried updating `sso` I found to my surprise that SSO has `grackle` 0.17.2 and yet `routes` 0.8.6.   The problem there is that `routes` 0.8.6 needs `grackle` 0.18.0:

```
[info] Fetched artifacts of 
[info] found version conflict(s) in library dependencies; some are suspected to be binary incompatible:
[info] 	* org.typelevel:grackle-core_3:0.18.0 (early-semver) is selected over {0.17.2}
[info] 	    +- edu.gemini:lucuma-graphql-routes_3:0.8.6           (depends on 0.18.0)
[info] 	    +- org.typelevel:grackle-sql_3:0.17.2                 (depends on 0.17.2)
[info] 	    +- org.typelevel:grackle-circe_3:0.17.2               (depends on 0.17.2)
```

I'm not sure how it ended up in that state.  No problem I thought, I'll just update grackle too.  That seems to work fine for SSO (or at least compiles) but now I need to update grackle in `itc` as well.  That's where the real trouble starts.  So `itc` compiles fine with grackle `0.18.0` but tests issue a mysterious complaint:

```
==> X lucuma.itc.service.GraphQLImagingTimeSuite.gmos north case  0.314s munit.ComparisonFailException: /Users/swalker/dev/lucuma-itc/modules/tests/src/test/scala/lucuma/itc/service/GraphQLSuite.scala:43
42:      .flatMap(_.as[Json])
43:      .assertEquals(expected)
44:
values are not the same
=> Obtained
JObject(
  value = object[errors -> [
  {
    "message" : "at offset 416\nexpectation:\n* must fail: exceeded maximum input value depth"
  }
]]
)
 JObject(
-  value = object[errors -> [
-  {
-    "message" : "at offset 416\nexpectation:\n* must fail: exceeded maximum input value depth"
+  value = object[data -> {
+  "imagingIntegrationTime" : {
+    "mode" : {
...
```

swalker2m:
It seems that `mergify` merged it: https://github.com/gemini-hlsw/lucuma-sso/pull/777.  I'm not sure how that passes CI checks though.

swalker2m:
This is getting closer but our tests expose a bug in `0.18.1` that needs to be fixed before we can continue.  There's supposedly a workaround in args to `compileAndRun` but we don't call `compileAndRun`, at least not directly.

PR #943, opened 2024-01-31
WHERE: { program: { reference: { EQ: "G-2024B-0001" } } }

swalker2m:
To allow the `observations` and `targets` queries to filter on `ProgramReference` as well as `ProgramId`, this PR updates `WhereObservation` and `WhereTarget` to remove the `programId` entry and replace it with `program: WhereProgram` (which in turn has a `reference` filter).

### Old Way

```
observations(WHERE: {
  programId: { EQ: "p-123" }
})
```

### New Way

```
observations(WHERE: {
  program: { id: { EQ: "p-123" } }
})
```
which also opens up

```
observations(WHERE: {
  program: { reference: { EQ: "G-2024B-0001" } }
})
```

among other program filtering options.

### Misc

* The `observations` query had a redundant, optional `programId` parameter in addition to the `WHERE` input.  This doesn't exist in the `targets` query.  It seemed best to provide just one way to filter on program id and to make it consistent with `targets` so I removed it.

* `WhereDataset` had an `observationId` field which I replaced with `observation: WhereObservation` to be consistent with the others and to open up filtering on other observation attributes.

* Changed the `programReference` field to `reference` to be consistent with `id` fields.  Also, `program: { reference: { EQ: "G-2024B-0001" } }` reads better than `program: { programReference: { EQ: "G-2024B-0001" } }`.

* There were __two__  identical `WhereTarget` Scala binding definitions: `WhereTarget.scala` and `WhereTargetInput.scala`.  I deleted one after about an hour trying to figure out why my changes weren't working.


tpolecat (approved):
👍🏻 It hadn't occurred to me that this change would be so pervasive, but it makes sense and seems good.

PR #945, opened 2024-02-01
*Group with ProgramReference

swalker2m:
Updates the `asterismGroup`, `constraintSetGroup`, and `targetGroup` queries to accept either a program id (as they do currently) or a program reference.  This depends on having a `program` field in the corresponding group element.  It uses this to create the predicate which matches the program.  Of the three:

* __`TargetGroup`__ already had a `program` field but it was marked `deprecated` and the comment said it shouldn't be used?  

```
  Link back to program, temporary, to avoid some mapping issues.
  Please don't use this in API calls.
```

* __`AsterismGroup`__ already had a `program` field with neither `deprecated` nor other admonitions of any kind.
* __`ConstraintGroup`__ had just a `programId` which I turned into `program`.

So, is this not kosher for some reason?  I believe I saw that the latest Grackle has support for hidden `SqlObject`s in the mappings that we might use if necessary?

### DIFF
--- a/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -8148,13 +8148,17 @@ type Query {
   # ): TargetEnvironmentGroupSelectResult!
 
   """
-  Observations grouped by commonly held targets
+  Observations grouped by commonly held targets.  Specify one of programId or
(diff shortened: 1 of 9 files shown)


tpolecat (approved):
:+1: I had to add the program reference in order to construct a predicate over the program id, so basically the situation we were just talking about. I don't think there's any harm in leaving it in the API though, if you just want to undeprecate it.

PR #949, opened 2024-02-05
Program reference in mutations

swalker2m:
Adds a `programReference` option to 

* `createGroup`,
* `createObservation` and
* `createTarget`

As before, the inputs which previously required a `ProgramId` have to be modified to accept an optional `ProgramId` or `ProgramReference`.  If a reference is supplied, we first lookup the corresponding pid and then pass that through to the service which does the mutation.

Other mutations which would need a similar treatment (in a future PR) include:

* `createUserInvitation`
* `linkUser`
* `setAllocation`
* `unlinkUser`
* `updateAsterisms`
* `updateObsAttachments`

Mutations such as `updateObservations` and `updateTargets` now work with `WHERE` clauses and in those cases a program reference may be supplied in the `WhereProgram` `reference` field.


modules/service/src/main/scala/lucuma/odb/service/ProgramService.scala
```diff
@@ -283,6 +291,16 @@ object ProgramService {
...
 
+    val selectPid: Query[ProgramReference, Program.Id] =
+      sql"""
+        SELECT
+          c_program_id
+        FROM
+          t_program
+        WHERE
+          c_program_reference = $program_reference
+      """.query(program_id)
```
swalker2m, line 302: I'd originally thought that wherever a program reference like `G-2025A-0001-Q` would be accepted, we would also accept `G-2025A-0001` (since the proposal reference prefix uniquely identifies the program).  There are several places which will need an update if we do this, including the PID lookup here.

Another would be in the `WhereOptionOrderProgramReference` which has the familiar `EQ`, `NEQ`, `IN`, etc. These would (somehow) have to be changes so that proposal prefix matches would also work.  I'm wondering if we should just require exact matches.  If the proposal is accepted and has a program type like `Q` it becomes required and supplying just `G-2025A-0001` would fail.  What do you think?

swalker2m:
@tpolecat you're right in the middle of refactoring error message handling, if I recall correctly.  Does this work need to wait for a bit until that's done?

modules/service/src/main/scala/lucuma/odb/service/ProgramService.scala
tpolecat: It makes sense to me to allow the non-suffixed reference, but we probably need to compute and store it so we can index on it. We may already be doing this ... ?

tpolecat:
> Does this work need to wait for a bit until that's done?

No, carry on and I will merge. Most of what I'm doing is pretty mechanical.

tpolecat (approved):
This looks good. Most of the time client software will be using GIDs, so the extra work in `selectPid` seems fine to me. 👍🏻

modules/service/src/main/scala/lucuma/odb/service/ProgramService.scala
swalker2m: I wasn't planning on it but it's trivial to have columns for each.  I'll add that.  There's just the one type `ProgramRerence` for these.  I could add `ProposalReference` but then everywhere that we accept a `ProgramId` or a `ProgramReference` we'd also need to accept a `ProposalReference`.  Maybe that's the most straightforward, easy to understand and easy to implement solution.

tpolecat: Having distinct schema types would definitely be the most transparent way to do it, but it's fine with me if we just fall back to proposal reference as an internal thing, if no suffix is specified. My concern was just that doing a prefix match would end up being slow.

PR #964, opened 2024-02-16
Bump lucuma-core

rpiaggio:


swalker2m:
To make this happen, we'll have to get ITC up-to-date with Grackle as well and there is an unresolved issue with that.

> Trying to update Grackle in one of our projects results in test case failures like "message" : "at offset 481\nexpectation:\n* must fail: exceeded maximum input value depth". Any pointers where we might look into this would be appreciated.
[18:35](https://gemini-software.slack.com/archives/GN10WURRD/p1708119329189509)
The query that fails doesn’t appear to be particularly deeply nested. https://github.com/gemini-hlsw/lucuma-itc/blob/main/modules/tests/src/test/scala/lucuma/itc/service/GraphQLImagingTimeSuite.scala

swalker2m:
There’s a migration associated with this so I’ll close in favor of #942.

PR #973, opened 2024-02-22
Program Types

swalker2m:
Adds program types according to the program reference proposal. 

## Schema Changes

* Adds `ProposalReference`, which before was subsumed into `ProgramReference`.
* Made `ProgramReference` and `ProposalReference` into `type`s with their own fields, of which are `ProgramReferenceLabel` and `ProposalReferenceLabel` scalar `label`s.
* Updates queries and mutations to accept `programId` (PID), `proposalReference`, or `programReference`.
* Adds a `setProgramReference` mutation to change program references.
* Adds `WhereProposalReference` and `WhereProgramReference` for filtering programs.
* Comments out (for now) some unused fields in `WhereProposal`

## Pending

Right now semester is still an independently modifiable `Program` property, and it must be set in order to successfully submit a proposal.  I want to rework the proposal submission, acceptance and rejection in another PR.  I believe submission should require a semester and no longer be offered as a general property.  It makes no sense to supply a semester for a LIB, or XPL program for example.  Acceptance has different permissions than submission.  I think explicit mutations for the proposal status might be the way to go.


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -11508,29 +11806,34 @@ input WhereProposal {
...
+#
+#  """
+#  Matches the proposal abstract.
+#  """
+#  abstract: WhereOptionString
+#
+#  """
+#  Matches proposal partners.
+#  """
+#  partners: WhereProposalPartners
```
toddburnside, line 11836: Commented out `Where`s. Do we think we'll implement one or more of these later? Should we also comment out or  remove the `WhereProposalClass`, `WhereEqToActivation`, etc?

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -10217,6 +10401,33 @@ input WhereEqPartner {
...
+
+  """
+  Matches if the program type is not the supplied value.
+  """
+  NEQ: ProgramType
+
+  """
+  Matches if the program type is any of the supplied options.
+  """
+  IN: [ProposalType!]
```
toddburnside, line 10423: `ProgramType`?

swalker2m: D'oh, thanks!

toddburnside (approved):
LGTM

PR #1000, opened 2024-02-27
Record planned time

swalker2m:
Some time ago we discussed recording the step time estimate as steps are recorded by Observe.  This PR adds an `estimate` field to the `StepRecord` interface which reads the time estimate calculated at the time a new step record is received.  If this seems acceptable, a future PR would add a simple query to sum the estimates for the executed part of the observation.


tpolecat (approved):
This looks great. Hopefully this will resolve the "original estimate" discussion. 👍🏻

PR #1014, opened 2024-02-28
Structured ODB Errors

tpolecat:
This took way too long because I was waffling between semi-structured errors and a proper ADT, and whether or not to include the current user in all errors (which ended up being complicated). So after trying all the hard solutions I ended up with the simplest possible thing and it seems totally fine.

So what we have is `lucuma.odb.data.OdbError` which is a ADT defined in the schema module to make it available to clients. Errors coming from the server _may_ have an extension called `odb_error` that contains a serialized instance. The intent is that all application-generated errors should include this extension.

On the server there's convenience syntax to turn an `OdbError` into a Grackle `Problem`, `Failure`, or `Warning`. So anywhere we're in `Result` we should use this technique to allow for better error reporting on the client. Nothing bad will happen if you just do `Result.failure("oops")` but the `odb_error` extension won't be present when it gets to the client. `OdbSuite` will emit a warning if it sees such an error.

This PR also includes a few bits of cruft from when I was dicking around trying to figure out what I should do first. They're harmless so I just left them in. See notes below.


modules/service/src/main/scala/lucuma/odb/graphql/input/CoordinatesInput.scala
```diff
@@ -13,38 +13,26 @@ import lucuma.core.math.RightAscension
...
+    dec: Option[Declination]
+  )
+  object Edit:
+    val Binding: Matcher[Edit] =
+      ObjectFieldsBinding.rmap {
+        case List(
+          RightAscensionInput.Binding.Option("ra", rRa),
+          DeclinationInput.Binding.Option("dec", rDec)
+        ) => (rRa, rDec).parMapN(Edit(_, _))
+      }
```
tpolecat, line 36: This change moves the ra/dec check to validation time; it had been happening at the service level which is somewhat wasteful as a transaction is allocated and immediately discarded if the check fails. This is kind of random and not really related to the rest of the PR, sorry.

modules/service/src/main/scala/lucuma/odb/service/ExecutionUserCheck.scala
```diff
@@ -17,4 +20,10 @@ trait ExecutionUserCheck {
       case _       => f(user).asLeft
     }
 
+  def checkUser2[F[_]](using Services[F]): Result[Unit] =
+    user.role.access match {
+      case Service => Result.unit // TODO: should specifically be the observe service user, whose name should be passed in the app config
+      case _       => OdbError.NotAuthorized(user.id).asFailure
+    }
+
```
tpolecat, line 28: This is part of the access control review that I started on but got distracted. It will go away.

modules/service/src/main/scala/lucuma/odb/data/OdbErrorExtensions.scala.scala
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
```
swalker2m, line 1: `OdbErrorExtensions.scala.scala`?

modules/service/src/main/scala/lucuma/odb/graphql/input/CoordinatesInput.scala
```diff
@@ -13,38 +13,26 @@ import lucuma.core.math.RightAscension
...
-  object messages {
-    val BothRaAndDecNeeded: String =
-      "Both ra and dec are required in order to specify a coordinate."
-  }
+  type Create = Coordinates
+  object Create:
+    val Binding: Matcher[Create] =
+      Edit.Binding.emap:
+        case Edit(Some(ra), Some(dec)) => Right(Coordinates(ra, dec))
+        case _ => Left("Both ra and dec are required in order to specify a coordinate.")
```
swalker2m, line 23: Should this be an `.rmap` maybe that fails with `Matcher.validationFailure("Both ra and dec ....")`?  Just trying to get a sense for when to use it.

modules/service/src/main/scala/lucuma/odb/graphql/input/TargetEnvironmentInput.scala
```diff
@@ -10,26 +10,36 @@ import lucuma.odb.data.Nullable
...
-}
+trait TargetEnvironmentInput:
+  def asterism: Nullable[List[Target.Id]]
+
+object TargetEnvironmentInput:
+
+  final case class Create(
+    explicitBase: Nullable[CoordinatesInput.Create],
+    asterism:     Nullable[List[Target.Id]]
+  ) extends TargetEnvironmentInput
```
swalker2m, line 21: I'm curious about the `Nullable` in the `Create` as opposed to `Option`.  Maybe it's all the same.

modules/service/src/main/scala/lucuma/odb/service/extensions.scala
```diff
@@ -3,9 +3,23 @@
...
 import skunk.data.Completion
 
 extension [F[_]: MonadCancelThrow](s: Session[F]) def executeCommand(af: AppliedFragment): F[Completion] =
-  s.prepareR(af.fragment.command).use(_.execute(af.argument))
\ No newline at end of file
+  s.prepareR(af.fragment.command).use(_.execute(af.argument))
+
+extension (self: ResultT.type)
+  def fromResult[F[_]]: FromResultPartiallyApplied[F] = new FromResultPartiallyApplied
+  def success[F[_], A](fa: F[A])(using Functor[F]) = ResultT(fa.map(Result.success))
```
swalker2m, line 20: These would be handy in Grackle.

swalker2m (approved):
This looks excellent!  :+1: Just a couple of questions for my own edification as I use this going forward.

modules/service/src/main/scala/lucuma/odb/data/OdbErrorExtensions.scala.scala
tpolecat: doh

modules/service/src/main/scala/lucuma/odb/graphql/input/CoordinatesInput.scala
tpolecat: If you use `emap` here the left side will get turned into a validation error, so either way.

modules/service/src/main/scala/lucuma/odb/graphql/input/TargetEnvironmentInput.scala
tpolecat: Good point, I'll change that.

PR #1017, opened 2024-02-29
Use reference model in lucuma-core v0.93.0

swalker2m:
Now using `ProgramReference` etc. that were moved to core.


toddburnside (approved):
LGTM. I think we'll need a release of the schemas with this update.

tpolecat (approved):

Issue #1023, opened 2024-02-29
Interface Issues


swalker2m:
The query:

```graphql
query {
  observation(observationId: "o-151") {
    execution {
      config(futureLimit: 100) {
        ... on GmosNorthExecutionConfig {
          static {
            stageMode
            detector
          }
        }
        ... on GmosSouthExecutionConfig {
          static {
            stageMode
            detector
          }
        }
      }
    }
  }
}
```

Produces: 

![image_720](https://github.com/gemini-hlsw/lucuma-odb/assets/4906023/3ffbcddb-d622-4543-9730-5a8ad75f493d)

swalker2m:
Check #1024 to see this problem in action.

swalker2m:
Discussion with Miles indicates that the query / schema violate the spec and Grackle is now conforming to it whereas before it was not.  The issue as I understand it is that two concrete implementation of an interface cannot have leaf fields (at least) with the same name and yet with different types.  This despite the fact that the fields are not part of the interface itself.  So in this case we have `stageMode: GmosNorthStageMode!` in `GmosNorthStatic` and `stageMode: GmosSouthStageMode!` in `GmosSouthStatic`, which don't even share an interface but appear in `GmosNorthExecutionConfig` and `GmosSouthExecutionConfig` both of which implement `ExeuctionConfig`.

```
interface ExecutionConfig {
  instrument: Instrument!
}

type GmosNorthExecutionConfig implements ExecutionConfig {
  static: GmosNorthStatic!
  ...
}

type GmosSouthExecutionConfig implements ExecutionConfig {
  static: GmosSouthStatic!
  ...
}

type GmosNorthStatic {
  stageMode: GmosNorthStageMode!
  ...
}

type GmosSouthStatic {
  stageMode: GmosSouthStageMode!
  ...
}
```

swalker2m:
To me this seems to render `interface`s dangerously unusable.  When you make a concrete `interface` implementation you have to check all other implementations and ensure that you've invented unique names for the fields which have differing types.

swalker2m:
So one solution would be to ban interfaces, at least where this is likely to be a problem as in the case of `ExecutionConfig`. If we do this, `ExecutionConfig` would become a type with optional fields, all but one of which will always be `null`.

```
type ExecutionConfig {
  gmosNorth: GmosNorthExecutionConfig
  gmosSouth: GmosSouthExecutionConfig
  ghost: GhostExecutionConfig
[...]

PR #1024, opened 2024-02-29
Test case

swalker2m:
Grackle 0.18.1 revealed that we were unknowingly taking advantage of a nonconformity of Grackle with respect to the GraphQL spec.  In particular, the path down to a leaf must always resolve to the same type regardless of the particular interface implementation (see #1023).  To make this possible, we can rename fields at a high level to introduce a distinguishing prefix.  That turns, for instance, 

* `execution` / `config` / `static` / `stageMode` into
*  `execution` / `config` / `gmosNorthStatic` / `stageMode` and
*  `execution` / `config` / `gmosSouthStatic` / `stageMode`.

For which a query might produce, for example:

```json
"execution": {
  "config": {
    "gmosNorthStatic": {
      "stageMode": "FOLLOW_XY"
    }
  }
}
```

but never the ambiguous

```json
"execution": {
  "config": {
    "static": {
      "stageMode": "FOLLOW_XY"
    }
  }
}
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -31,7 +31,7 @@ val skunkVersion               = "0.6.3"
 val pprintVersion              = "0.8.1"
 val testcontainersScalaVersion = "0.40.14" // N.B. 0.40.15 causes java.lang.NoClassDefFoundError: munit/Test
 
-ThisBuild / tlBaseVersion      := "0.11"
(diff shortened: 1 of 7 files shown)


swalker2m:
I'll need to rework some of the other interfaces along similar lines.

swalker2m:
Fixed by #1032

PR #1032, opened 2024-03-05
Interface Issue

swalker2m:
Works around the GraphQL `interface` path limitation by avoiding `interface`s.


toddburnside (approved):
LGTM. I think this approach works best for our API given our constraints. Thanks for tackling this.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -9395,6 +9094,17 @@ interface StepRecord {
...
+  GMOS North instrument configuration for this step, if any.  This will be null
+  unless the `instrument` discriminator is "GMOS_NORTH".
+  """
+  gmosNorth: GmosNorthDynamic!
+
+  """
+  GMOS South instrument configuration for this step, if any.  This will be null
+  unless the `instrument` discriminator is "GMOS_SOUTH".
+  """
+  gmosSouth: GmosSouthDynamic!
```
swalker2m, line 9107: Oops, these should be optional now.  Not sure how that worked.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -10110,6 +9820,16 @@ interface Visit {
...
 
+  """
+  GmosNorth static instrument configuration, for GMOS North visits.
+  """
+  gmosNorth: GmosNorthStatic!
+
+  """
+  GmosSouth static instrument configuration, for GMOS South visits.
+  """
+  gmosSouth: GmosSouthStatic!
```
swalker2m, line 9831: Make optional.

tpolecat (approved):
This all kind of sucks but it's better to discover it now I guess. 👍🏻

PR #1035, opened 2024-03-06
UpdateAsterismsInput

swalker2m:
There is a collection of `updateFoo` mutations in the schema: `updateAsterisms`, `updateDatasets`, etc.  They all take a corresponding `UpdateFooInput` (e.g., `UpdateAsterismsInput`) which have a similar collection of fields:

* `SET` - what to change
* `WHERE` - which to change
* `LIMIT` - max result count to return
* `includeDeleted` - consider deleted observations, targets, etc.

A couple of these (`UpdateAsterismsInput` and `UpdateObsAttachmentsInput`) require an additional `programId` argument rather than using the `WHERE` clause to restrict the scope of the update.  I'd like to make it consistent across all these similar mutations.  Assuming you can filter on programs in the `WHERE` clause, this way you can also use program references or proposal references instead of just program ids.  In particular, for the asterism update:

```
WHERE: {
  program: { reference: { EQ: "G-2024A-0001-Q" }  }
}
```

The drawback of gaining consistency and making it possible to use references is that it becomes hidden that the update *requires* that all the observations passed by the `WHERE` filter have the same program.  If you try to update the asterism across programs you get an error message.  Still it *seems* worth it to me.

This PR updates `UpdateAsterismsInput` to remove the `programId` field.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/AsterismService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/AsterismService.scala
@@ -13,6 +13,8 @@ import cats.syntax.foldable.*
 import cats.syntax.functor.*
 import eu.timepit.refined.types.string.NonEmptyString
 import grackle.Result
+import grackle.ResultT
+import grackle.syntax.*
 import lucuma.core.math.Coordinates
 import lucuma.core.math.ProperMotion
 import lucuma.core.model.CatalogInfo
@@ -74,7 +76,6 @@ trait AsterismService[F[_]] {
    * deleting targets as indicated.
    */
   def updateAsterism(
-    programId:      Program.Id,
     observationIds: NonEmptyList[Observation.Id],
     ADD:            Option[NonEmptyList[Target.Id]],
     DELETE:         Option[NonEmptyList[Target.Id]]
@@ -96,6 +97,22 @@ object AsterismService {
   def instantiate[F[_]: MonadCancelThrow: Concurrent](using Services[F]): AsterismService[F] =
 
     new AsterismService[F] {
+
+      private def selectProgramId(
+        observationIds: NonEmptyList[Observation.Id]
+      ): F[Result[Program.Id]] = {
+        val af = Statements.selectProgramId(observationIds)
+        session.prepareR(af.fragment.query(program_id)).use { p =>
+          p.stream(af.argument, chunkSize = 16)
+           .take(2)
+           .compile
+           .toList
+           .map {
+             case List(pid) => pid.success
+             case _         => OdbError.InvalidObservationList(observationIds).asFailure
+           }
+        }
+      }
 
       override def insertAsterism(
         programId:      Program.Id,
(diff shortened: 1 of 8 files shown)


modules/service/src/main/scala/lucuma/odb/service/AsterismService.scala
```diff
@@ -97,6 +98,24 @@ object AsterismService {
...
     new AsterismService[F] {
 
+      // Selects the program id that is shared by all the given observations,
+      // or else fails.
+      private def selectProgramId(
+        observationIds: NonEmptyList[Observation.Id]
+      ): F[Result[Program.Id]] = {
+        val af = Statements.selectProgramId(observationIds)
+        session.prepareR(af.fragment.query(program_id)).use { p =>
+          p.stream(af.argument, chunkSize = 16)
```
toddburnside, line 108: Could you just use `unique` instead of stream? Or is it just easier to handle the error case like this?

toddburnside (approved):
LGTM

modules/service/src/main/scala/lucuma/odb/service/AsterismService.scala
swalker2m: Yeah, that was my first attempt but I wasn't happy with the error handling.  I guess it could go either way.

toddburnside: No strong opinion here. This is fine.

swalker2m: I want it to fail unless there is exactly one result.  `unique` would fail with an exception when more than one is returned by the query or if none are returned, which I could have `recoverWith`ed I think.  That might be clearer I'm not sure.

tpolecat: Gotcha, yeah either way works. What you have is fine.

tpolecat (approved):

modules/service/src/main/scala/lucuma/odb/service/AsterismService.scala
swalker2m: Thanks.  I went back and tried again with the `unique` / `recover` option and remembered the reason I switched was that I would have to catch `skunk.exception.SkunkException` which seemed overly broad. I'll just leave it this way

```
skunk.exception.SkunkException: Exactly one row was expected, but more were returned.
	at skunk.exception.SkunkException$.fromQueryAndArguments(SkunkException.scala:145)
	at skunk.PreparedQuery$$anon$1.fail(PreparedQuery.scala:134)
	at skunk.PreparedQuery$$anon$1.unique$$anonfun$1(PreparedQuery.scala:118)
	at cats.effect.IOFiber.succeeded(IOFiber.scala:1204)
	at cats.effect.IOFiber.runLoop(IOFiber.scala:245)
	at cats.effect.IOFiber.asyncContinueSuccessfulR(IOFiber.scala:1372)
	at cats.effect.IOFiber.run(IOFiber.scala:113)
	at cats.effect.unsafe.WorkerThread.run(WorkerThread.scala:634)
```

PR #1036, opened 2024-03-06
UpdateObsAttachmentInput

swalker2m:
This is a similar treatment as that described in #1035, but for `UpdateObsAttachmentInput`.  In this case, I needed to add the `WhereProgram` filter.  It now allows updating attachments in multiple programs, but you still need to have permission to update them.


toddburnside (approved):
LGTM

tpolecat (approved):
Looks good 👍🏻

PR #1037, opened 2024-03-06
remove OdbError construction from mapping layer

tpolecat:
I have come to the conclusion that the mapping layer should be entirely mechanical; i.e., all business logic (including interpreting results and turning them into user errors) should live in the service layer. This will make service composition much safer because we won't need to worry about logic that's living somewhere else (an example of this is `cloneObservation`, which currently does *not* clone the asterism; instead it's done in two steps in the mapping logic), and in principle should make testing and automation easier because API and direct service calls will behave (and fail) the same way.

This PR gets us some of the way there by pushing error construction down into service calls (which now return `Result`s).  A side-effect is that some services now construct structured responses and then immediately turn them into `OdbError`s, which is harmless but dumb so I will remove this in a followup PR.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/graphql/mapping/MutationMapping.scala
+++ b/modules/service/src/main/scala/lucuma/odb/graphql/mapping/MutationMapping.scala
@@ -5,7 +5,6 @@ package lucuma.odb.graphql
 package mapping
 
 import cats.Functor
-import cats.data.Ior
 import cats.data.Nested
 import cats.data.NonEmptyList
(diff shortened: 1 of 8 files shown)


swalker2m (commented):
Looks great so far.  Pushing this stuff down seems like a good improvement.

tpolecat:
Ok that's all I'm going to do for this one, so it's ready when you are.

swalker2m (commented):
Looks great. 👍

swalker2m (approved):

PR #1040, opened 2024-03-07
integrate time accounting for event insertion

tpolecat:
Similar to the last PR. This one moves event-related time accounting updates out of the mapping and into the event service.


swalker2m (approved):
Excellent that’s better

PR #1041, opened 2024-03-07
ObservationReference

swalker2m:
This is a work in progress on observation references.  If possible I'd appreciate feedback on a couple of points in particular, which I'll detail in comments but:

* Use of a view for the observation reference
* Somewhat dodgy exception catching

### DIFF
--- a/modules/service/src/main/resources/db/migration/V0847__observation_reference.sql
+++ b/modules/service/src/main/resources/db/migration/V0847__observation_reference.sql
@@ -1,0 +1,92 @@
+-- Add an index column to the observation table.
+ALTER TABLE t_observation
+  ADD COLUMN c_observation_index int4 CHECK (c_observation_index > 0),
+  ADD CONSTRAINT unique_observation_index_per_program UNIQUE (c_program_id, c_observation_index);
+
+-- A function to obtain the next observation number for a particular program.
+CREATE OR REPLACE FUNCTION next_observation_index(pid d_program_id)
+RETURNS INT
+AS $$
+DECLARE
+    sequence_name VARCHAR;
+BEGIN
+    sequence_name := CONCAT('s_', REPLACE(pid::text, '-', '')); -- a hyphen (as in p-123) will not work
+
+    BEGIN
+        EXECUTE 'CREATE SEQUENCE IF NOT EXISTS ' || sequence_name;
+    EXCEPTION WHEN unique_violation THEN
+        NULL; -- sequence exists, this was an attempt to create it in parallel
+    END;
+
+    RETURN nextval(sequence_name)::INT;
+END;
+$$ LANGUAGE plpgsql;
+
+-- Update all existing observations so that they have an index defined.
+UPDATE t_observation
+  SET c_observation_index = next_observation_index(c_program_id);
+
+-- Make the index required.
+ALTER TABLE t_observation
+  ALTER COLUMN c_observation_index SET NOT NULL;
+
+-- Add a trigger that sets the observation index when a new observation is
+-- created.
+CREATE OR REPLACE FUNCTION set_observation_index()
+RETURNS TRIGGER AS $$
+BEGIN
+  NEW.c_observation_index = next_observation_index(NEW.c_program_id);
+  RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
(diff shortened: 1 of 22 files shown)


modules/service/src/main/resources/db/migration/V0847__observation_reference.sql
```diff
@@ -0,0 +1,92 @@
...
+DECLARE
+    sequence_name VARCHAR;
+BEGIN
+    sequence_name := CONCAT('s_', REPLACE(pid::text, '-', '')); -- a hyphen (as in p-123) will not work
+
+    BEGIN
+        EXECUTE 'CREATE SEQUENCE IF NOT EXISTS ' || sequence_name;
+    EXCEPTION WHEN unique_violation THEN
+        NULL; -- sequence exists, this was an attempt to create it in parallel
+    END;
```
swalker2m, line 19: I did this trick of creating a sequence on the fly elsewhere.  I pulled it out again for maintaining the observation index for each program.  Unfortunately it produced unique constraint violations in one particular unrelated test case.

```
skunk.exception.PostgresErrorException: 
🔥  
🔥  Postgres ERROR 23505 raised in _bt_check_unique (nbtinsert.c:664)
🔥  
🔥    Problem: Duplicate key value violates unique constraint
🔥             "pg_class_relname_nsp_index".
🔥     Detail: Key (relname, relnamespace)=(s_p101, 2200) already exists.
```

The test case was in `lucuma.odb.graphql.query.constraintSetGroup` where we do:

```
createObservation(user, pid, iq, sb).replicateA(2)
```

which I assume is somehow creating the same sequence in parallel.  Adding this exception handler "fixes" the problem but I wonder if it discredits the entire approach.  I'm not sure what else to do though other than reserving a sequence when the program is created I suppose.

modules/service/src/main/resources/db/migration/V0847__observation_reference.sql
```diff
@@ -0,0 +1,92 @@
...
+
+-- Create a view for the observation reference, including only observations with
+-- a defined reference.  This makes mapping the reference as optional easier.
+CREATE VIEW v_observation_reference AS
+  SELECT
+    o.c_observation_id,
+    o.c_program_id,
+    o.c_observation_index,
+    p.c_program_reference,
+    CONCAT(p.c_program_reference::text, '-', LPAD(o.c_observation_index::text, 4, '0')) AS c_observation_reference
```
swalker2m, line 88: We'd discussed making the program reference part of a compound PK for `t_program` and then adding a FK reference in `t_observation` for which updates cascade.  It occurred to me that it might also just be computed in a view and let searching be handled by a combination of the program reference and the observation index.  What do we think?

modules/service/src/main/resources/db/migration/V0847__observation_reference.sql
swalker2m: I could do that pretty easily in this case.  For the semester counters I could just write a mutation that reserves them for the foreseeable future.

tpolecat: I think maybe go with the exception handler. It's weird but it does seem to fix the issue?

modules/service/src/main/resources/db/migration/V0847__observation_reference.sql
tpolecat: My thinking was that if prog/obs references are actual [computed] columns then we can do indexed searches without having to parse the input references. If we compute these values in a view then I think searches will end up being linear ... ? 

I may misunderstand your proposal here.

swalker2m: Given an observation reference like "G-2025A-0001-Q-0001" you can find it, I think quickly, with

```scala
    val selectOid: Query[ObservationReference, Observation.Id] =
      sql"""
        SELECT o.c_observation_id
        FROM t_observation o
        INNER JOIN t_program p ON p.c_program_id = o.c_program_id
        WHERE p.c_program_reference = $program_reference
          AND o.c_observation_index = $int4_pos
      """.query(observation_id).contramap[ObservationReference] { ref =>
        (ref.programReference, ref.observationIndex)
      }
```

I may be overlooking the obvious.

PR #1042, opened 2024-03-07
push program reference resolution into service layer

tpolecat:
More moving composed service calls into the service layer.


swalker2m (approved):
Looks good.  :+1:

PR #1045, opened 2024-03-11
ObservationReference (Take 2)

swalker2m:
This is a second attempt at observation reference, or more accurately an evolution from #1041.  Observation references consist of the corresponding program reference plus a unique (to the program) observation index.  The data migration:

* Adds an observation index which is kept by a program-specific `sequence`.
* Adds a copy of the program's `c_program_reference` in `t_observation`.  This is maintained in-sync explicitly via triggers.
* Adds a `GENERATED ALWAYS` observation reference label that combines the program reference and the index.

I'm hesitant to make the program reference part of a compound primary key for `t_program`, and then simply reference it as a foreign key in `t_observation`.  Maybe that is a more straightforward approach.  I don't have a ton of experience with this kind of thing but it just _feels wrong_ to make the primary key have a mutable, nullable and redundant component.  I can be easily persuaded to damn the torpedos and change this though.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/data/ObservationReference.scala
+++ b/modules/service/src/main/scala/lucuma/odb/data/ObservationReference.scala
@@ -1,0 +1,59 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.odb.data
+
+import cats.Order
+import cats.parse.Parser
+import cats.parse.Parser.*
+import eu.timepit.refined.types.numeric.PosInt
+import io.circe.Decoder
+import io.circe.Encoder
+import io.circe.syntax.*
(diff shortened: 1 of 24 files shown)


modules/service/src/main/scala/lucuma/odb/data/ObservationReference.scala
```diff
@@ -0,0 +1,59 @@
...
+import io.circe.Decoder
+import io.circe.Encoder
+import io.circe.syntax.*
+import lucuma.core.model.ProgramReference
+import lucuma.core.model.parser.ReferenceParsers
+import lucuma.core.optics.Format
+
+case class ObservationReference(
+  programReference: ProgramReference,
+  observationIndex: PosInt
```
swalker2m, line 19: I would move this to `core` in the future, just as `ProposalReference` and `ProgramReference` started their careers in `lucuma.odb.data` and went on to `lucuma.core.model`.

tpolecat (approved):
👍🏻 This looks great.

PR #1046, opened 2024-03-11
dynamic sequence creation update

swalker2m:
A small migration to

* remove an unused next index function
* handle unique constraint violations like the one discovered in `V0847__observation_reference.sql`


tpolecat (approved):
👍🏻

PR #1048, opened 2024-03-12
Fix observation reference trigger comparison

swalker2m:
When the program reference changes, a trigger updates the observations that correspond to the program.  Unfortunately the comparison was failing because it wasn't handling `null` correctly.  :-/


toddburnside (approved):
LGTM

PR #1050, opened 2024-03-13
DatasetReference

swalker2m:
Initial support for `DatasetReference`, for example, `G-2024B-1234-Q-0001-0002-0003`. Adds the `DatasetReference` type, data migrations and the mapping from `Dataset` to its `reference`.

Next up, lookup by dataset reference and `WHERE` filtering.


modules/service/src/main/scala/lucuma/odb/data/DatasetReference.scala
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
```
toddburnside, line 1: Will this get moved to lucuma-core, too, so that it can be used by the clients?

toddburnside (approved):
LGTM

modules/service/src/main/scala/lucuma/odb/data/DatasetReference.scala
swalker2m: Yes.  I believe I'll do that today in fact.

swalker2m: https://github.com/gemini-hlsw/lucuma-core/pull/871

PR #1052, opened 2024-03-14
`DatasetReference` in lookup and `WHERE`

swalker2m:
Rounds out #1050, adding dataset lookup and filtering by `DatasetReference` to the schema.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/graphql/mapping/QueryMapping.scala
+++ b/modules/service/src/main/scala/lucuma/odb/graphql/mapping/QueryMapping.scala
@@ -19,6 +19,7 @@ import grackle.skunk.SkunkMapping
 import lucuma.core.model
 import lucuma.core.model.ProgramReference
 import lucuma.core.model.ProposalReference
+import lucuma.odb.data.DatasetReference
 import lucuma.odb.data.ObservationReference
 import lucuma.odb.data.Tag
 import lucuma.odb.data.TargetRole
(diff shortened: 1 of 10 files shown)


modules/service/src/main/scala/lucuma/odb/service/DatasetService.scala
```diff
@@ -30,6 +31,14 @@ import Services.Syntax.*
 
 sealed trait DatasetService[F[_]] {
 
+  /**
+   * Finds the dataset id consistent with the given ids (if any).
+   */
+  def resolveDid(
+    did: Option[Dataset.Id],
+    ref: Option[DatasetReference]
+  ): F[Result[Dataset.Id]]
```
swalker2m, line 40: I thought I'd need this but then it turned out I didn't.  Still, it seems inevitable and it prompted generalizing the id resolver which is now shared with the `ObservationService` so I decided I'd leave it.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/QueryMapping.scala
```diff
@@ -140,18 +142,31 @@ trait QueryMapping[F[_]] extends Predicates[F] {
...
+    rDid: Result[Option[model.sequence.Dataset.Id]],
+    rRef: Result[Option[DatasetReference]],
+    pred: DatasetPredicates
+  ): Result[Predicate] =
+    (rDid, rRef).parTupled.map { (did, ref) =>
+      and(List(
+        did.map(pred.id.eql).toList,
+        ref.map(r => pred.referenceLabel.eql(r)).toList
+      ).flatten) match {
+        case True => False // neither did nor ref was supplied
```
toddburnside, line 155: I like this. 😆

toddburnside (approved):
LGTM

modules/service/src/main/scala/lucuma/odb/graphql/mapping/QueryMapping.scala
swalker2m: Yeah Grackle collapses an empty list down to `Predicate.True` but in this case it shouldn't match anything.

PR #1056, opened 2024-03-15
Update core, use core references

swalker2m:
Updates to `lucuma-core.0.93.1`, using its `ObservationReference` and `DatasetReference` instead of the `lucuma.odb.data` versions.


toddburnside (approved):
LGTM

PR #1061, opened 2024-03-17
Fix GCAL Mapping Issue

swalker2m:
Fixes a mapping issue for GCAL continuum, which should be optional.

```
type Gcal implements StepConfig {
  """
  GCAL continuum, present if no arcs are used
  """
  continuum: GcalContinuum
```

```scala
    object Gcal {
      val Continuum: ColumnRef = col("c_gcal_continuum", gcal_continuum.opt)
```


rpiaggio (approved):
LGTM. Thank you!

PR #1065, opened 2024-03-18
Update Core, add SlewEvent

swalker2m:
Updates to core 94, adding slew events and step-level events for abort, continue, pause and stop.


cquiroz (approved):
Thanks, LGTM

PR #1070, opened 2024-03-19
Library updates 10

swalker2m:
Random library updates.  Mostly I want to confirm that the migration validation still works after I snubbed it last night.


swalker2m:
All the non-build changes are due to the Scala fix update.

cquiroz (approved):

Issue #253, opened 2023-02-08
Changes to Proposal Schema


toddburnside:
This issue is the result of discussion in https://github.com/gemini-hlsw/lucuma-core/pull/666
The `proposed` changes are:

- Make `toOActiviation` `!` required in the schema.
- Allow for `partnerSplits` to be either an empty array, or one whose split percents add up to 100. The empty array is required because there is no sensible default.
- Fix the creation validation. The code is supposed to check `proposalClass`, `toOActivation` and `partnerSplits` on creation. But, a proposal can be created with only a `proposalClass`. In this case the `toOActivation` is returned as a null, and the `partnerSplits` as an empty array.

toddburnside:
Completed by #1071

PR #1071, opened 2024-03-19
Add createProposal, updateProposal and setProposalStatus mutations

toddburnside:
This removes the ability to create proposals, update proposals and set proposal status from the updatePrograms mutation and makes separate mutations for them.

The updatePrograms mutation still retains the ability to set the semester, but this will be removed when we get a mutation for setting Call for Proposal.


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -537,6 +537,29 @@ type CreateProgramResult {
...
+"""
+Input for creating a proposal.  Identify the program to update with one
+of `programId`, `proposalReference` or `programReference`.  If more than one of
+these is specified, all must match.  Use `SET` to specify the new program
+reference properties.
+"""
+input CreateProposalInput {
+  programId:         ProgramId
+  proposalReference: ProposalReferenceLabel
+  programReference:  ProgramReferenceLabel
```
swalker2m, line 549: Seems like it would be odd to already have a proposal or program reference when creating a new proposal.

modules/service/src/main/scala/lucuma/odb/graphql/input/SetProposalStatusInput.scala
```diff
@@ -0,0 +1,30 @@
...
+import lucuma.core.model.ProgramReference
+import lucuma.core.model.ProposalReference
+import lucuma.odb.data.Tag
+import lucuma.odb.graphql.binding.*
+
+final case class SetProposalStatusInput(
+  programId: Option[Program.Id],
+  proposalReference: Option[ProposalReference],
+  programReference: Option[ProgramReference],
+  status:           Tag
```
swalker2m, line 17: I hope we can switch up our enum handling somehow.

modules/service/src/test/scala/lucuma/odb/graphql/feature/reference.scala
```diff
@@ -267,154 +252,22 @@ class reference extends OdbSuite {
...
-      _   <- addProposal(pi, pid)
-      ref <- submitProposal(pi, pid, none) // no semester
-    } yield ref
-
-    assertIO(res, ref2010A1)
-  }
-
-  // TODO: proposal status mutations
-  // I think we want to remove semester as an independently settable property.
-  test("TEMP: cannot unset semester after submit") {
```
swalker2m, line 287: Is it now not possible to unset the semester via the API after a science program reference has been established?

modules/service/src/main/scala/lucuma/odb/service/ProgramService.scala
```diff
@@ -313,72 +286,10 @@ object ProgramService {
...
-          session.prepareR(Statements.getTempTableData).use(
-            _.stream(Void, chunkSize = 1024)
-             .fold(Result.unit) { case (acc, (pid, oldSemester, psTag, hasProposal)) =>
-               val check: Result[Unit] =
-                 for {
-                   oldStatus <- tagToProposalStatus(psTag)
-                   _         <- validateStatusUpdate(pid, oldStatus, newStatusUpdate, hasProposal)
-                   finalStatus   = newStatusUpdate.getOrElse(oldStatus)
-                   finalSemester = SET.semester.fold(none, oldSemester, _.some)
-                   _         <- UpdateProgramsError.NoSemesterForSubmittedProposal(pid)
```
swalker2m, line 365: I think (hope) we would get unhandled constraint violations if we allow people to unset the semester.  Is this handled somewhere else so that it is caught before it gets that far?

modules/service/src/test/scala/lucuma/odb/graphql/DatabaseOperations.scala
```diff
@@ -163,117 +188,89 @@ trait DatabaseOperations { this: OdbSuite =>
...
+        prog <- programCursor.downFields("reference", "label").success.traverse(_.as[ProgramReference])
+        prop <- programCursor.downFields("proposal", "reference", "label").success.traverse(_.as[ProposalReference])
+      } yield (prog, prop)).leftMap(f => new RuntimeException(f.message)).liftTo[IO]
     }
 
+  def submitProposal(user: User, pid: Program.Id): IO[ProposalReference] =
+    setProposalStatus(user, pid, "SUBMITTED").map(_._2.get) // should have a proposal reference now.
+
+  def unsubmitProposal(user: User, pid: Program.Id): IO[Option[ProposalReference]] =
+    setProposalStatus(user, pid, "SUBMITTED").map(_._2)
```
swalker2m, line 264: Should this be `"UNSUBMITTED"` here?

modules/service/src/main/scala/lucuma/odb/graphql/input/SetProposalStatusInput.scala
toddburnside: Yeah, me too.

modules/service/src/test/scala/lucuma/odb/graphql/DatabaseOperations.scala
toddburnside: Oops. Good catch. Darn copy and paste.

modules/service/src/test/scala/lucuma/odb/graphql/feature/reference.scala
```diff
@@ -147,15 +127,18 @@ class reference extends OdbSuite {
     for {
       pid0 <- createProgramAs(pi)
       _    <- addProposal(pi, pid0)
-      ref0 <- submitProposal(pi, pid0, sem2024B.some)
+      _    <- setSemester(pi, pid0, sem2024B)
```
swalker2m, line 130: Do we have, or could we add, a test somewhere in which we try to submit a proposal without first having set the semester to make sure it is handled in some reasonable-ish way?

toddburnside: I thought I had that, but I guess not. I'll add it.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
toddburnside: Good point. I'll remove the references. I guess I got carried away by the boilerplate.

swalker2m: Raises a question: what would happen if you create a proposal for a program that already has one?  Does it just reset it and become un-submitted / unaccepted if submitted / accepted?

modules/service/src/test/scala/lucuma/odb/graphql/feature/reference.scala
toddburnside: As discussed in the comment in `ProgramService`, the ability to change/unset the semester has been left in `updatePrograms` until we get a way to set the Call for Proposal, which will also set the semester. We could make it so the API does not allow unsetting the semester for now to avoid the problem until we have another method of setting the semester.

swalker2m: Right, I think that seems reasonable.

modules/service/src/test/scala/lucuma/odb/graphql/feature/reference.scala
toddburnside: I was wrong. There is one. 4th test in `setProposalStatus`

modules/service/src/test/scala/lucuma/odb/graphql/feature/reference.scala
toddburnside: OK. I'll do that.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
toddburnside: If you try to create a proposal for a program that already has one, you get an error. There is a test for that in `createProposal`

modules/service/src/test/scala/lucuma/odb/graphql/feature/reference.scala
toddburnside: Done

modules/service/src/main/scala/lucuma/odb/service/ProgramService.scala
toddburnside: It is now not possible to unset the semester.

swalker2m (approved):
Looks great to me. :+1:

toddburnside:
Latest commit allows empty or missing partner splits on creation. This addresses issue #253. 

It seems that when proposal creation was part of `updatePrograms`, the validation was messed up, maybe because it could be either an proposal update or creation, and missing (but not empty) were allowed to pass. With the new `createProposal`, the validation worked. But, the UI requires the ability to create a proposal without partner splits.

PR #1075, opened 2024-03-20
Reference Updates

swalker2m:
Adds `COM` (commissioning) and `MON` (monitoring) program types in the latest reference spec.  These are formatted just like `CAL` and `ENG` and have their own semester indices (e.g., `G-2025A-COM-GMOSN-01` and `G-2025A-MON-GMOSN-01` can both coexist in the database along with the `01` `CAL` and `ENG` programs).


modules/service/src/main/resources/db/migration/V0854__program_reference_update.sql
```diff
@@ -0,0 +1,3 @@
+-- Add the new types to the program type enum
+ALTER TYPE e_program_type ADD VALUE 'commissioning' BEFORE 'engineering';
+ALTER TYPE e_program_type ADD VALUE 'monitoring'    BEFORE 'science';
```
swalker2m, line 3: I've learned my lesson and put these in a separate migration (i.e., separate transaction).

modules/service/src/main/resources/db/migration/V0855__program_reference_update2.sql
```diff
@@ -0,0 +1,187 @@
...
+    abbr text;
+    prefix text;
+    sequence_name VARCHAR;
+BEGIN
+    SELECT c_abbr INTO abbr FROM t_program_type WHERE c_type = program_type;
+    prefix := CONCAT('s_', semester, '_', abbr);
+
+    IF instrument IS NOT NULL THEN
+        sequence_name := CONCAT(prefix, '_', instrument);
+    ELSE
```
swalker2m, line 64: Updating a function requires, as far as I know, completely rewriting it.  Old migrations cannot be touched.  The relevant change is that now it looks for whether `instrument` is `NULL` when deciding to include the instrument in the sequence name.  For `CAL`, `COM`, `ENG` and `MON` it will be supplied.  For science it will be `NULL`.

swalker2m:
The corresponding core updates are in so this is now ready for review.

toddburnside (approved):
LGTM

tpolecat (approved):
Looks good. :+1:

PR #1077, opened 2024-03-21
access control in some service method types

tpolecat:
Some service methods are restricted entirely based on things we know without first consulting the database. In these cases we can push access control out of the methods and require that callers pass evidence that the requirement is met. So for example we can say:

```scala
trait SomeService[F[_]]:
  def someMethod(args)(using Transaction[F], Services.AdminAccess): F[Result[Something]]
```

And the caller needs to provide a `Services.AdminAccess`. We can do this with the `requireAdminAccess` extension method:

```scala
services.useTransactionally:
  requireAdminAccess:
    someService.someMethod(args)
```

This will yield a `Failure` with `NotAuthorized` if the current user isn't an Admin or better, otherwise it calls the passed thunk normally.

The intent here is that we can capture a subset of access control and make it evident in the types. It also makes access control transitive for services calling one another, at least in these cases.


modules/service/src/main/scala/lucuma/odb/service/Services.scala
```diff
@@ -259,6 +274,14 @@ object Services:
...
     def guideService[F[_]](httpClient: Client[F], itcClient: ItcClient[F], commitHash: CommitHash, ptc: TimeEstimateCalculator.ForInstrumentMode)(using Services[F]): GuideService[F] = summon[Services[F]].guideService(httpClient, itcClient, commitHash, ptc)
     def userInvitationService[F[_]](using Services[F]): UserInvitationService[F] = summon[Services[F]].userInvitationService
+    
+    def requireStaffAccess[F[_], A](fa: Services.StaffAccess ?=> F[Result[A]])(using Services[F], Applicative[F]): F[Result[A]] =
+      if user.role.access >= Access.Staff then fa(using ())
+      else OdbError.NotAuthorized(user.id).asFailureF
+
+    def requireServiceAccess[F[_], A](fa: Services.ServiceAccess ?=> F[Result[A]])(using Services[F], Applicative[F]): F[Result[A]] =
+      if user.role.access >= Access.Service then fa(using ())
+      else OdbError.NotAuthorized(user.id).asFailureF
```
tpolecat, line 284: It's probably possible to abstract this pattern into a helper method but I couldn't figure it out.

swalker2m (commented):
This looks amazing.  :+1: I really like how it makes the access requirement explicit at the mapping level and in the signature of the service method.  Seems like a win.

tpolecat:
Ok this is ready. There were a few more cases I missed the first time around, but I think I got them all. Here's a list of all the service methods, with check marks where access is limited based on user role, if you want to double-check that I didn't miss any.

- https://gist.github.com/tpolecat/0fdd7e69aadfd3f28960469e7e048263

modules/service/src/main/scala/lucuma/odb/graphql/mapping/MutationMapping.scala
```diff
@@ -512,19 +508,21 @@ trait MutationMapping[F[_]] extends Predicates[F] {
   private lazy val UpdateDatasets: MutationField =
     MutationField("updateDatasets", UpdateDatasetsInput.binding(Path.from(DatasetType))) { (input, child) =>
       services.useTransactionally {
-        // Our predicate for selecting datasets to update
-        val filterPredicate = and(List(
-          Predicates.dataset.observation.program.isWritableBy(user),
-          input.WHERE.getOrElse(True)
-        ))
+        requireStaffAccess {
```
swalker2m, line 511: I wonder if we'll need, at some point, _either_ staff or service.  We don't need it now, but can something like that be expressed?  I suppose that's just `Staff` since `Service` is more powerful than `Staff`?

swalker2m (approved):
I looked at this before.  I'm very happy about how much easier and transparent this is.  Seems like a huge win.  :+1:

modules/service/src/main/scala/lucuma/odb/graphql/mapping/MutationMapping.scala
tpolecat: `Guest` < `Pi` < `Ngo` < `Staff` < `Admin` < `Service` so yeah you just ask for the bigger one. We could get rid of the transitive aspect and ask for a union of roles if we need it to be more fine-grained.

PR #1081, opened 2024-03-25
invitation recipient

tpolecat:
This adds `recipientEmail: EmailAddress!` to `type UserInvitation`, the idea being that once we have email hooked up we can email the key rather than returning it to the user to deliver by hand. The email address can also be used to disambiguate when revoking invitations.


cquiroz (approved):
LGTM
Thanks a lot, should it be enforced the email to be unique per invitation? Do we need that?

tpolecat:
> should it be enforced the email to be unique per invitation? Do we need that?

We need to support cases where people decline an invitation and then later accept another one for example, and there's no harm in redeeming more than one invitation; it's a no-op if you're already associated with the program. So I think at least for now let's leave it as-is.

PR #1086, opened 2024-03-28
Spectroscopy Config Options

swalker2m:
Defines spectroscopy instrument configuration options (see the table of options in Explore) and a query for obtaining them. The idea is to move the definition of options from a client-side `.csv` file to the database so that we can match options to the selected call for proposals.  As of this PR, there is no call for proposals or availability filtering.  It basically offers all the GMOS long slit options along with filtering.  

An example query might be:

```graphql
          spectroscopyConfigOptions(
            WHERE: {
              wavelengthOptimal: {
                AND: [
                  { nanometers: { GT: 760.0 } }
                  { nanometers: { LT: 770.0 } }
                ]
              }
              resolution: { GT: 1000 }
            }
          ) {
            name
          }
```

This query is open to anybody in the world, since it is just about exploring Gemini's capabilities.


tpolecat (approved):
This looks great. 👍🏻

PR #1093, opened 2024-03-29
Update regex for emails

cquiroz:
The regex used for emails is not working if the email name has a dot.
I updated the regex based on this:
https://stackoverflow.com/questions/201323/how-can-i-validate-an-email-address-using-a-regular-expression


rpiaggio (approved):
LGTM

tpolecat (approved):
Yikes! 👍🏻

PR #1096, opened 2024-03-30
Observe Client Data

swalker2m:
This PR is the answer to a request from @rpiaggio.  The database generates sequences with atom and step ids.  The Observe application records the atoms and steps it will execute before executing them.  In response it receives a new atom or step id.  For nefarious UI purposes, there is a need to tie the generated id to the recorded id when possible.  The PR introduces what is essentially, to the database, opaque optional client data in the form of atom and step ids.  Observe can supply these IDs and later read them back for its own purposes.

Hopefully this will satisfy the requirement, though keep in mind the original generated atom and step ids are not guaranteed to remain stable for all time.  Any code update for example will cause all the unexecuted generated atom and step IDs to change.


tpolecat (approved):
This looks fine. 👍🏻

PR #1099, opened 2024-04-01
introspection test

tpolecat:
This adds an introspection test. It's failing now but should pass once https://github.com/gemini-hlsw/lucuma-odb/pull/1098 is merged.


modules/service/src/test/scala/lucuma/odb/graphql/query/introspection.scala
```diff
@@ -0,0 +1,119 @@
...
+                    }
+                  }
+                }
+              }
+            }
+          }
+        }      
+      """
+    ).onError: t =>
+      fail("\n🐞🐞🐞\n🐞🐞🐞 Schema introspection failed!\n🐞🐞🐞\n", t)
```
tpolecat, line 119: Introspection failure results in a massive stacktrace, so this message gets lost unless it's over-emphasized.

tpolecat:
(I'll wait to make sure it fails here, then merge with main and make sure it passes.)

PR #1103, opened 2024-04-02
Require PiAccess for the proposal mutations

toddburnside:
Andy requested that guest users not be allowed to create proposals. This has been enforced in Explore for some time, but not by the API. I believe the reasoning is that if a user is going to create a proposal, they must be getting serious. And, it is too easy for guest users to loose everything if they clear cookies, have computer issues, etc.
 
Because a PI must be non-guest to create a proposal, and any users linked to a program must be non-guest, the `requirePiAccess` probably isn't strictly necessary for `setProposalStatus` and `updateProposal`, but it seemed to be consistent. It can be removed if there are objections.


tpolecat (approved):
Looks great!

PR #1107, opened 2024-04-03
StepExecutionState

swalker2m:
This PR is an answer to a request for more information for the sequence UI in Explore and Observe.  It invents a new `StepExecutionState` enumeration which is computed based on the last step event received from Observe (if any).

### DIFF
--- a/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -3615,6 +3615,27 @@ input StepConfigSmartGcalInput {
 }
 
 """
+Execution state of a recorded step.
+"""
+enum StepExecutionState {
+
+  "No events have been received since the step was recorded."
+  NOT_STARTED
+
+  "Events have been received, but no terminal `Abort`, `EndStep` or `Stop` events."
+  ONGOING
+
+  "An event with an 'ABORT' 'StepStage' was received."
+  ABORTED
+
+  "An event with an 'END_STEP' 'StepStage' was received."
+  COMPLETED
+
+  "An event with a 'STOP' 'StepStage' was received."
+  STOPPED
+}
+
+"""
 Step type
 """
 enum StepType {
@@ -9418,6 +9439,12 @@ type StepRecord {
 
   "The step was created by Observe at this time."
   created: Timestamp!
+
+  """
+  The execution state of this step, according to events received (if any) from
+  Observe.
+  """
+  executionState: StepExecutionState!
 
   """
   Time interval during which this step executed.  This measures the range of
--- a/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -9419,7 +9419,12 @@ type StepRecord {
   "The step was created by Observe at this time."
   created: Timestamp!
 
-  "Time interval during which this step executed."
+  """
+  Time interval during which this step executed.  This measures the range of
+  time from the first event to the last, whether or not the step ever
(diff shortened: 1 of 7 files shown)


modules/service/src/main/scala/lucuma/odb/graphql/mapping/StepRecordMapping.scala
```diff
@@ -54,6 +54,7 @@ trait StepRecordMapping[F[_]] extends StepRecordView[F]
         SqlField("instrument",   StepRecordView.Instrument, discriminator = true),
         SqlObject("atom",        Join(StepRecordView.AtomId, AtomRecordTable.Id)),
         SqlField("created",      StepRecordView.Created),
+        EffectField("executionState", executionStateHandler, List("id")),
```
swalker2m, line 57: This is accomplished via a query in an `EffectHandler`.  Another alternative would be to update the step record as events are received, storing the updated execution state in the `t_step_record` table as we go and avoiding a more costly query.  That requires a couple more moving parts though.

modules/service/src/main/scala/lucuma/odb/data/StepExecutionState.scala
```diff
@@ -0,0 +1,30 @@
...
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.odb.data
+
+import lucuma.core.util.Enumerated
+
+/**
+ * The execution state of a recorded step.
+ *
+ * @param tag database tag
```
swalker2m, line 11: There isn't actually an enumeration in the database for this, at least not yet.  It is used though in Json codecs.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -3614,6 +3614,27 @@ input StepConfigSmartGcalInput {
...
+  ONGOING
+
+  "An event with an 'ABORT' 'StepStage' was received."
+  ABORTED
+
+  "An event with an 'END_STEP' 'StepStage' was received."
+  COMPLETED
+
+  "An event with a 'STOP' 'StepStage' was received."
+  STOPPED
```
swalker2m, line 3635: An `ABANDONED` state might make sense as well, but it would require something other than just step events to determine.

tpolecat: Would we need to do a sweep after the fact and mark steps as abandoned if we never got a final event?

modules/service/src/main/scala/lucuma/odb/graphql/mapping/StepRecordMapping.scala
tpolecat: :+1: Seems reasonable to do it the "easy" way and add more caching if it's a problem.

tpolecat (approved):
This looks good. :+1:

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
swalker2m: Yeah I think that's right.  A daily sweep of steps that are still `ONGOING`.  I'll leave it for now unless @rpiaggio thinks it is needed?

rpiaggio: I plan to just hide the `ONGOING` step in observe, and the atom where an `ONGOING` step is located in explore.

So, without a daily sweep we would be hiding executed steps from abandoned atoms. Otherwise, we can check the starting instant and assume that an `ONGOING` atom from the current night is executing, and if it's from a previous night then it is abandoned.

rpiaggio (approved):
LGTM, thank you!

PR #1108, opened 2024-04-03
Filter out GAIA bands on ITC calls

swalker2m:
This is a requested hack to remove GAIA bands from source profiles until they are supported in the ITC.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/GeneratorParamsService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/GeneratorParamsService.scala
@@ -233,8 +233,8 @@ object GeneratorParamsService {
         mode:       InstrumentMode,
         wavelength: Wavelength
       ): ValidatedNel[Error, (Target.Id, (ImagingIntegrationTimeInput, SpectroscopyIntegrationTimeInput))] = {
-        val targetBand = params.sourceProfile.flatMap(_.nearestBand(wavelength)).map(_._1)
-
+        val sourceProf = params.sourceProfile.map(_.gaiaFree)
+        val targetBand = sourceProf.flatMap(_.nearestBand(wavelength)).map(_._1)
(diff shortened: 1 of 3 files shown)


modules/service/src/main/scala/lucuma/odb/service/extensions.scala
```diff
@@ -23,3 +25,18 @@ class FromResultPartiallyApplied[F[_]]:
   def apply[A](result: Result[A])(using F: Applicative[F]): ResultT[F, A] =
     ResultT(F.pure(result))
 
+extension (self: SourceProfile)
+  def gaiaBands: Set[Band] =
```
rpiaggio, line 29: This is a constant. It can be confusing to have it as an instance method.

If it's meant to be accessed only from here maybe it can be a `private val GaiaBands` ?

rpiaggio (approved):
LGTM.

Does this affect the UI?

cquiroz:
It should be mostly transparent to explore but we need to reset the cache

modules/service/src/main/scala/lucuma/odb/service/GeneratorParamsService.scala
```diff
@@ -233,16 +233,16 @@ object GeneratorParamsService {
         mode:       InstrumentMode,
         wavelength: Wavelength
       ): ValidatedNel[Error, (Target.Id, (ImagingIntegrationTimeInput, SpectroscopyIntegrationTimeInput))] = {
-        val targetBand = params.sourceProfile.flatMap(_.nearestBand(wavelength)).map(_._1)
-
+        val sourceProf = params.sourceProfile.map(_.gaiaFree)
+        val targetBand = sourceProf.flatMap(_.nearestBand(wavelength)).map(_._1)
```
cquiroz, line 237: This  is key. wee need to recalculate the nearest band removing gaia

cquiroz (approved):
LGTM, thanks

PR #1114, opened 2024-04-05
Call for Proposals

swalker2m:
This is an initial stab at defining the Call for Proposals.  It supports only creation and simple lookup by ID so far.  They are not used for anything yet.

## Future Work

* `WHERE` clause lookup?  This would allow searches by semester and/or type and/or active period etc.
* Editing.  At the moment you have to create one in its final state.
* Use in proposal submission.
* Use in filtering configuration lookups.
* Validation of observations to make sure they meet the RA/Dec limits?
* Remove/integrate `ProposalClassEnum` and `ProposalClass`?


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -282,6 +282,189 @@ enum Breakpoint {
...
+  """
+  The unique Call for Proposals id associated with this Call.
+  """
+  id: CallForProposalsId!
+
+  """
+  Open/closed status of this Call. Submissions are only accepted when the Call
+  is OPEN.
+  """
+  status: CallForProposalsStatus!
```
cquiroz, line 299: I wonder if this should rather be dates start/end of the CfP

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -282,6 +282,189 @@ enum Breakpoint {
...
+  CLOSED
+}
+
+enum CallForProposalsType {
+  DEMO_SCIENCE
+  DIRECTORS_TIME
+  FAST_TURNAROUND
+  LARGE_PROGRAM
+  POOR_WEATHER
+  REGULAR_SEMESTER
```
cquiroz, line 464: Are Exchanges inside Regular?

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
swalker2m: Each partner has a deadline for submission, and there's a start and end time for when observing will happen associated with this Call.  This was just meant as a way of leaving it off until and whenever we aren't ready for them.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
swalker2m: Good question.  It's unclear what we'll need to do for exchange time and they are missing from the CfP document that I based this on.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -282,6 +282,189 @@ enum Breakpoint {
...
+  "DELETED or PRESENT.  On create defaults to PRESENT."
+  existence: Existence
+}
+
+"""
+Groups a partner with its submission deadline.
+"""
+type CallForProposalsPartner {
+  partner:  Partner!
+  deadline: Timestamp!
```
tpolecat, line 436: I think `submissionDeadline` might be clearer.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/MutationMapping.scala
```diff
@@ -260,6 +262,17 @@ trait MutationMapping[F[_]] extends Predicates[F] {
             Predicates.cloneTargetResult.newTarget.id.eql(newTargetId)
           ), child)
 
+  private lazy val CreateCallForProposals: MutationField =
+    MutationField("createCallForProposals", CreateCallForProposalsInput.Binding) { (input, child) =>
+      services.useTransactionally {
+        requireStaffAccess {
```
tpolecat, line 268: 👍🏻

tpolecat (approved):
This looks good. 👍🏻 

Will a CFP be implicitly closed when all deadlines have passed? How does that work?

Did you solve your list of leaf values problem by adding the deadline to the partner object?

swalker2m:
> Did you solve your list of leaf values problem by adding the deadline to the partner object?

It was the instruments list that has that issue. I had to make a view that represents those values as an array type.

swalker2m:
> I wonder if this should rather be dates start/end of the CfP

> Will a CFP be implicitly closed when all deadlines have passed? How does that work?

Maybe too influenced by the way I think this works today, I added an explicit open vs. closed parameter.  The idea being that we could control the options displayed in Observe by querying CfPs that are set to `OPEN`.  It would give Gemini the ability to turn off submissions for a call arbitrarily and temporarily when necessary.  Maybe this isn't needed though and we can make it implicitly open as soon as created and closed when all the partner's deadlines have passed.  Or maybe we need explicit start/end proposal acceptance dates.  I added a question to the CfP doc.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -8043,6 +8238,26 @@ type Query {
...
+  """
+  callsForProposals(
+    OFFSET: CallForProposalsId
+
+    "Limits the result to at most this number of matches (but never more than 1000)."
+    LIMIT: NonNegInt
+
+    "Set to true to include deleted values"
+    includeDeleted: Boolean! = false
+  ): CallsForProposalsSelectResult!
```
toddburnside, line 8259: What about including an `includeClosed` option as well. For display in explore we'll only care about open CfPs.

modules/service/src/main/scala/lucuma/odb/data/CallForProposalsType.scala
```diff
@@ -0,0 +1,22 @@
...
+  case FastTurnaround     extends CallForProposalsType("fast_turnaround")
+  case LargeProgram       extends CallForProposalsType("large_program")
+  case PoorWeather        extends CallForProposalsType("poor_weather")
+  case RegularSemester    extends CallForProposalsType("regular_semester")
+  case SystemVerification extends CallForProposalsType("system_verification")
+}
+
+object CallForProposalsType {
+
+  given Enumerated[CallForProposalsType] = Enumerated.derived
```
toddburnside, line 20: I think you can also do this as `enum CallForProposalsType(val tag: String) derives Enumerated:` and skip the object?

toddburnside:
> * Use in proposal submission.

I think we probably want a `setCfP` mutation and then validate it on proposal submission. That seems like the best workflow for the UI.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
swalker2m: Yes, the plan for the next related PR is to add a `WHERE` clause with options like that and more.

modules/service/src/main/scala/lucuma/odb/data/CallForProposalsType.scala
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.odb.data
+
+import lucuma.core.util.Enumerated
+
+enum CallForProposalsType(val tag: String) {
```
toddburnside, line 8: It seems we'll be wanting these enums available in Explore. We have been moving them to `core`, but another possibility would be to move them to the odb schemas module?

swalker2m: I was planning on putting it in `core` since that's where everything else goes.  Putting it in `odb-schemas` would be easier though if satisfactory.  All schemas has at the moment are codecs and the `OdbSchema.graphql` itself.

toddburnside (approved):
LGTM

modules/service/src/main/scala/lucuma/odb/data/CallForProposalsType.scala
toddburnside: Just leave it for now. Since the enum doesn't contain any information than the tag, we can start by using the enum generated by clue.

swalker2m:
> Maybe too influenced by the way I think this works today, I added an explicit open vs. closed parameter. 

Bryan voted for having calls automatically open while there is at least one partner with a pending deadline.  Accordingly I've removed the explicit `CallForProposalsStatus`.

PR #1124, opened 2024-04-11
Acquisition Sequence Correction

swalker2m:
The PR reworks acquisition sequence generation.  I was treating acquisition sequences just like science sequences, which are just a list of atoms which all must eventually be executed.  Acquisitions are different though because they are

* __infinite__:  whenever `nextAtom` is called for an acquisition sequence there should always be another value
* __repeating__: we start over from the beginning when returning to an acquisition in a new visit or when switching back from a science sequence

Mostly the changes are to the completion state tracking, which gets acquisition-specific information.  This was previously just a `CompletedAtomMap` (renamed `Completion.AtomMap`).  Now there is a map for acquisition and a map for science.  The acquisition map is wiped clean whenever the acquisition sequence needs to repeat from scratch.  The science map continues building up as before.


cquiroz (approved):
LGTM

PR #1125, opened 2024-04-11
ShortCut 2772 + lucuma.core 0.96.1

swalker2m:
Adds a test case that shows that the API returns the expected binning, when defaulted, for the parameters in ShortCut 2772.


modules/service/src/test/scala/lucuma/odb/graphql/issue/shortcut/2772.scala
```diff
@@ -0,0 +1,275 @@
...
+                   },
+                 },
+                 radialVelocity: {
+                   kilometersPerSecond: 27.58
+                 },
+                 parallax: {
+                   milliarcseconds: 2.422
+                 }
+               },
+               sourceProfile: {
```
cquiroz, line 121: Did you say the bug  was related  to the source profiile

cquiroz (approved):
LGTM

modules/service/src/test/scala/lucuma/odb/graphql/issue/shortcut/2772.scala
swalker2m: That's correct.  For a point source, we use the smaller of the IQ and the slit width to determine the object size.  For uniform, it is effectively just the slit width.  For Gaussian, it is the explicit FWHM.  The issue was that the conversion of the IQ to the corresponding angle was incorrect.  For example an IQ of `ONE_POINT_ZERO` was turned into an angle of 10 microarcseconds instead of 1 arcsecond.  This [was fixed in core](https://github.com/gemini-hlsw/lucuma-core/pull/885).

PR #1133, opened 2024-04-17
Where CFP

swalker2m:
A few initial `WHERE` clause options for `CallForProposals` searches.

It's missing the main option, which I think would be to obtain all the open calls.  That requires checking the current time vs all the partner deadlines.  I'm not sure how to do that.


modules/service/src/main/resources/db/migration/V0863__call_for_proposals_update.sql
```diff
@@ -0,0 +1,33 @@
...
+ALTER TABLE t_cfp
+  ADD c_active_start timestamp,
+  ADD c_active_end   timestamp;
+
+UPDATE t_cfp
+  SET c_active_start = COALESCE(lower(c_active), '2000-01-01 00:00:00'::timestamp),
+      c_active_end   = COALESCE(upper(c_active), '2000-01-01 00:00:00'::timestamp);
+
+ALTER TABLE t_cfp
+  DROP c_active,
```
swalker2m, line 12: I switched this from a `tsrange` to two `timestamp` columns because I couldn't figure out how to deal with it in Grackle.

swalker2m:
Superseded by #1136

PR #1136, opened 2024-04-18
Where CFP

swalker2m:
A second attempt at an initial `WHERE` clause for `CallForProposals` searches.  Changes the active period schema from a start / end pair to a single `TimestampInterval`.

It's missing the main option, which I think would be to obtain all the open calls. That requires checking the current time vs all the partner deadlines. I'm not sure how to do that.


swalker2m:
Superseded by #1138

PR #1138, opened 2024-04-18
WHERE CFP

swalker2m:
Adds a `WHERE` clause to the Call for Proposals query.  It allows matching on

* id
* proposal type
* semester
* active period
* whether the call is currently open


modules/service/src/main/resources/db/migration/V0863__call_for_proposals_update.sql
```diff
@@ -0,0 +1,29 @@
...
+    array_remove(array_agg(i.c_instrument ORDER BY i.c_instrument), NULL) AS c_instruments,
+    (SELECT COUNT(*)
+      FROM t_cfp_partner
+     WHERE t_cfp_partner.c_cfp_id = c.c_cfp_id
+       AND t_cfp_partner.c_deadline > thyme.c_now
+    ) > 0 AS c_is_open
+  FROM
+    t_cfp c
+  CROSS JOIN
+    (SELECT CURRENT_TIMESTAMP AS c_now) AS thyme
```
swalker2m, line 24: The goal here is to compute the current time just once instead of once per row, and I think (?) this accomplishes it.  The name `thyme` is a bit cringe but I wanted to avoid using anything that seemed like a function name.  If someone has better ideas for how to structure this let me know.

modules/service/src/main/resources/db/migration/V0863__call_for_proposals_update.sql
```diff
@@ -0,0 +1,29 @@
...
+  LEFT JOIN
+    t_cfp_instrument i ON c.c_cfp_id = i.c_cfp_id
+  GROUP BY
+    c.c_cfp_id, thyme.c_now;
```
swalker2m, line 28: Postgres informed me that I had to add this to the grouping.  It _should_ be the same value for all rows so it _shouldn't_ matter ...

modules/service/src/main/resources/db/migration/V0863__call_for_proposals_update.sql
tpolecat: `CURRENT_TIMESTAMP` is constant within a transaction (it's the start time) so you don't need to worry about it changing … was that your concern?

tpolecat (approved):
One comment/question about the timestamp, but otherwise :+1:

modules/service/src/main/resources/db/migration/V0863__call_for_proposals_update.sql
swalker2m: Oh, excellent thanks.  I think I can simplify this a bit then ...

PR #1142, opened 2024-04-22
unlink user

tpolecat:
Here is `unlinkUser`. Simplified program user roles (pending) will make the access control simpler but for now there are a lot of cases to consider.


modules/service/src/main/scala/lucuma/odb/service/ProgramService.scala
```diff
@@ -262,6 +271,68 @@ object ProgramService {
           case LinkUserResponse.InvalidUser(uid)        => OdbError.InvalidUser(uid, Some(s"User $uid does not exist or is of a nonstandard type.")).asFailure
           case LinkUserResponse.Success(pid, user)      => Result((pid, user))
 
+      // delete a link unconditionally
+      private def unlinkUnconditionally(input: UnlinkUserInput): F[Result[Boolean]] =
+        session.prepareR(Statements.UnlinkUnconditionally).use: pc =>
+          pc.execute(input.programId, input.userId).map:
+            case Completion.Delete(1) => Result(true)
```
swalker2m, line 278: I'll admit I've tended to ignore the `Completion` but I suppose I should be checking it in cases like this.

swalker2m (approved):
Looks very good. :+1:

modules/service/src/main/scala/lucuma/odb/service/ProgramService.scala
tpolecat: Yeah, most of the time it doesn't matter. This is a rare case where it makes sense to look.

PR #1145, opened 2024-04-23
Add invitation emails

toddburnside:
Sends emails for invitations and tracks them by providing an endpoint for the mailgun webhooks.


modules/service/src/main/resources/db/migration/V0864__emails.sql
```diff
@@ -0,0 +1,107 @@
...
+  BEFORE UPDATE on t_email
+  FOR EACH ROW
+  EXECUTE PROCEDURE ch_update_email_status_time();
+
+-- Add the email id to the invitation
+
+ALTER TABLE t_invitation
+  -- Needs to be nullable because we need to add a row to get the hash before we can send the email.
+  -- Unfortunately, CHECK constraints cannot be deferred.
+  ADD COLUMN c_email_id TEXT NULL REFERENCES t_email (c_email_id);
```
toddburnside, line 47: So, really, we'll always have an email id for an invitation because I roll back the transaction with the initial add if sending the email fails. But, I couldn't see a way to make it NOT NULL.

modules/service/src/main/scala/lucuma/odb/data/EmailStatus.scala
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.odb.data
+
+import lucuma.core.util.Enumerated
+
+// TODO: Move to lucuma-core
```
toddburnside, line 8: I will move this to core and update in another PR.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/EmailMapping.scala
```diff
@@ -0,0 +1,33 @@
...
+        SqlField("id", EmailTable.EmailId, key = true, hidden = true),
+        SqlField("senderEmail", EmailTable.SenderEmail),
+        SqlField("recipientEmail", EmailTable.RecipientEmail),
+        SqlField("subject", EmailTable.Subject),
+        SqlField("textMessage", EmailTable.TextMessage),
+        SqlField("htmlMessage", EmailTable.HtmlMessage),
+        SqlField("originalTime", EmailTable.OriginalTime),
+        SqlField("status", EmailTable.Status),
+        SqlField("statusTime", EmailTable.StatusTime),
+        // SqlObject("program", Join(EmailTable.ProgramId, ProgramTable.Id)) // do we want this here?
```
toddburnside, line 29: Really, I only have the program id in t_email to facilitate sending programEdit subscription messages. It doesn't seem necessary to provide the program off of the Email, IMO.

modules/service/src/main/scala/lucuma/odb/service/UserInvitationService.scala
```diff
@@ -53,8 +58,35 @@ object UserInvitationService:
     def apply(id: UserInvitation.Id, detail: Option[String]): OdbError.InvitationError =
       OdbError.InvitationError(UserInvitation.Id.fromString.reverseGet(id), detail)
 
-  def instantiate[F[_]: MonadCancelThrow](using Services[F]): UserInvitationService[F] =
+  def instantiate[F[_]: MonadCancelThrow](emailConfig: Config.Email, httpClient: Client[F])(using Services[F]): UserInvitationService[F] =
     new UserInvitationService[F]:
+      
+      // TODO: move to UserInvitation class?
+      extension(invitation: UserInvitation)
+        def token: String = s"${invitation.id}.${invitation.body}"
```
toddburnside, line 66: Unless someone sees a reason not to, I'll move this to the UserInvitation in lucuma-core.

modules/service/src/main/scala/lucuma/odb/service/UserInvitationService.scala
```diff
@@ -53,8 +58,35 @@ object UserInvitationService:
...
+
+      def sendInvitation(input: CreateUserInvitationInput, invitation: UserInvitation)(
```
cquiroz, line 68: Really nice. I wonder if we should (eventually) add  the ability to easily change the contents of the email. I can foresee changes being requested often

toddburnside: Right. I should have commented on that. These messages are `VERY` temporary. We need a better mechanism for creating and editing them.

modules/service/src/main/resources/db/migration/V0864__emails.sql
```diff
@@ -0,0 +1,107 @@
...
+  'rejected',
+  'accepted',
+  'delivered',
+  'permanent_failure',
+  'temporary_failure'
+);
+COMMENT ON TYPE e_email_status IS 'Status of emails.';
+
+CREATE TABLE t_email (
+  c_email_id          TEXT           NOT NULL PRIMARY KEY CHECK(length(c_email_id) > 0),
```
swalker2m, line 13: How is the id structured?  It surprised me to see it can be arbitrary text.  I suppose it comes from the mail service provider.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/EmailMapping.scala
swalker2m: One reason, maybe, would be if we provide queries in the API to return say all the emails associated with programs matching a given `WherePrograms`.  Perhaps we don't plan on that though and in any case there's no reason to add it until we do.

swalker2m (approved):
This looks very good.  Seems like it must have been difficult and tedious so I commend you for tackling it.  :+1:

modules/service/src/main/resources/db/migration/V0864__emails.sql
toddburnside: Yes, it comes from the email provider. It has a structure, of sorts, but it seems imprudent to rely on that when we have no control over it and it isn't documented. It looks like `20240411142701.584cf35193edf2d5@gpp.bernmohr.com`. So, currently it is a valid email address.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/EmailMapping.scala
toddburnside: We might want to do something like that in the future. But, it is unclear to me at the moment. I foresee us sending out other types of emails in the future. We may want to provide a view of all emails, regardless of type, or maybe not. I'm guessing we will probably want to add an `EmailType` to the table in the future. At that point we can decide if/how we want to provide additional access to the emails.

toddburnside: Also, we can add emails to the Programs in the API without providing the Program for an Email. I guess I never see emails having their own top level query, but who knows. Maybe we'll send emails that aren't program specific and we'll have to make the program id optional in the table?

PR #1147, opened 2024-04-24
DO NOT MERGE!! Error swallowed in websocket connection

toddburnside:
This adds an intentional exception to the timeCharge effect handler for Programs. 

If you query the timeCharge for a program via the playground, the exception is printed to the ODB log. 

However, if you run explore and try to load any program, the exception is not printed to the log and the connection is repeatedly dropped.


tpolecat:
I will have a look.

toddburnside:
This issue has been fixed!

PR #1150, opened 2024-04-24
Surround `existsUserAccess` with () if has multiple clauses

toddburnside:
When the user is a PI, `existsUserAccess` returns a query with an `OR`. When the user is a COI, this led to some problems, such as this query:
image

One query in the code was smart enough to surround the the return value in parens, but others weren't. Rather than requiring all call sites to know enough to use parens, it seemed better to give that responsibility to `existsUserAccess`.


tpolecat (approved):
Thanks. This stuff is treacherous. 👍🏻

PR #1151, opened 2024-04-24
improve error handling

tpolecat:
This updates graphql-routes and adds an extra check for unmoored exceptions in the GraphQL handler. It also logs any internal errors it sees.

**PS** - it also changes the tests to run on a WS connection rather than http, so it acts more like Explore


swalker2m (approved):
Looks great.  This should help spot my divide by zeros much quicker.  :+1:

PR #1156, opened 2024-04-26
Change message for invitation emails.

toddburnside:


modules/service/src/main/scala/lucuma/odb/service/UserInvitationService.scala
```diff
@@ -68,14 +68,16 @@ object UserInvitationService:
         )
       
         val textMessage: NonEmptyString = NonEmptyString.unsafeFrom(
-          s"Here is your token: ${invitation.token}"
+          s"""${user.displayName} has invited you to collaborate on a Gemini proposal. To accept this invitation go to ${emailConfig.exploreUrl} to log in, then from the upper right menu select "Redeem Invitations" and enter the following key:
```
cquiroz, line 71: Someday  we'll send a link

cquiroz (approved):
LGTM

PR #1164, opened 2024-04-29
CFP Titles

swalker2m:
Adds a Call for Proposals title to the CFP table and view.  This is kept in a generated column which is based solely on other fields of the call.

This is an opening bid for how the titles should be formatted.  I'll ask @andrewwstephens for input.  I suspect SV might need the instrument(s) or something.

### DIFF
--- a/modules/service/src/main/resources/db/migration/V0867__cfp_titles.sql
+++ b/modules/service/src/main/resources/db/migration/V0867__cfp_titles.sql
@@ -4,32 +4,35 @@
 CREATE FUNCTION format_cfp_title(
   cfp_type e_cfp_type,
   semester d_semester,
-  active   tsrange
+  active   tsrange,
+  instruments d_tag[]
 ) RETURNS text AS $$
+DECLARE
+  instrument_list text;
 BEGIN
+  instrument_list := array_to_string(instruments, ', ');
   RETURN CASE
-    WHEN cfp_type = 'demo_science'        THEN concat(semester, ' ', 'Demo Science')
-    WHEN cfp_type = 'directors_time'      THEN concat(semester, ' ', 'Director''s Time')
+    WHEN cfp_type = 'demo_science'        THEN concat(semester, ' Demo Science')
+    WHEN cfp_type = 'directors_time'      THEN concat(semester, ' Director''s Time')
     WHEN cfp_type = 'fast_turnaround'     THEN
       concat(
-        left(semester, length(semester) - 1),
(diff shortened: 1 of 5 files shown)


modules/service/src/main/resources/db/migration/V0867__cfp_titles.sql
```diff
@@ -0,0 +1,53 @@
...
+      concat(
+        left(semester, length(semester) - 1),
+        ' ',
+        to_char(lower(active) + (upper(active) - lower(active))/2, 'FMMonth'),
+        ' Fast Turnaround'
+      )
+    WHEN cfp_type = 'large_program'       THEN concat(semester, ' ', 'Large Program')
+    WHEN cfp_type = 'poor_weather'        THEN concat(semester, ' ', 'Poor Weather')
+    WHEN cfp_type = 'regular_semester'    THEN concat(semester, ' ', 'Regular Semester')
+    WHEN cfp_type = 'system_verification' THEN concat(semester, ' ', 'System Verification')
```
swalker2m, line 23: Probably needs the instrument restrictions?  Maybe those should be included (when present) in general?

tpolecat (approved):
👍🏻

modules/service/src/main/resources/db/migration/V0867__cfp_titles.sql
```diff
@@ -0,0 +1,59 @@
...
+        ' ',
+        to_char(lower(active) + (upper(active) - lower(active))/2, 'FMMonth'),
+        ' ',
+        'Fast Turnaround'
+      )
+    WHEN cfp_type = 'large_program'       THEN concat(semester, ' Large Program')
+    WHEN cfp_type = 'poor_weather'        THEN concat(semester, ' Poor Weather')
+    WHEN cfp_type = 'regular_semester'    THEN concat(semester, ' Regular Semester')
+    WHEN cfp_type = 'system_verification' THEN
+          concat(semester, nullif(' ' || instrument_list, ' '), ' System Verification')
```
swalker2m, line 29: I had envisioned the title as a `GENERATED` column in `t_cfp` (the call for proposals table).  Unfortunately though I need to format the instrument names into the SV title.  The instruments are kept in a separate table (gathered together into an array column in the corresponding view `v_cfp`).  This makes it impossible (?) to use in a generated column I think so I moved the `c_title` field down to the view as seen below.

toddburnside (approved):
LGTM

PR #1167, opened 2024-04-30
Trigger observationEdit events when targets in the asterism change.

toddburnside:


swalker2m (approved):
The code looks great, and I think it makes sense.  I just have a nagging doubt about whether changing a target is the same as changing the observations that use it.  I suppose it is.  Given this, is there any point in having target edit events?

toddburnside:
> The code looks great, and I think it makes sense. I just have a nagging doubt about whether changing a target is the same as changing the observations that use it. I suppose it is. Given this, is there any point in having target edit events?

Target edit events are useful for other areas in Explore. At one point we had talked about finer grained events. We could add another event type - say `asterismEdit` - but this seemed easier. We are already doing something similar for attachments, but that could be changed, too.

What we could really use, but it seems much more difficult to write and maintain, is events for when times/sequences will be changed for observations, groups and programs. What we are trying to do with this PR is just to update the planned times for observations. So, we listen to observationEdit and when we get one, we ask for the times again. However, there are a lot of changes that can be made that don't affect the times, so we are over requesting.

PR #1178, opened 2024-05-06
Poor Weather Subtype

swalker2m:
Adds a poor weather science subtype (along with classical, queue, fast turnaround, etc.).  Introducing a poor weather subtype makes a 1:1 correspondence with the CfP poor weather type just as for fast turnaround, demo science etc.  (Queue and classical are still combined in the "regular semester" CfP type.)


cquiroz (approved):
LGTM

toddburnside (approved):
LGTM

PR #1183, opened 2024-05-16
Delete Group

### LINKED ISSUE #1182: Deleted obs in allGroupElements
This query returns deleted observations. Should this be changed to only return present observations?

```graphql
query {
  program(programId: "p-100") {
    allGroupElements {
        observation {
          id
          existence
        }
    }
  }
}
```

tpolecat:
Ok this adds basic support for group deletion via an `existence` field, as with other model objects. Attempting to delete a group that contains a non-deleted element results in an error, *as does undeleting an element in a deleted group*. This means deletion needs to happen bottom-up, and undeletion needs to happen top-down. This may end up being bad for usability, but then again it may be fine because it's easy to understand (and it's easy to implement "undo").

I believe this also fixes #1182


tpolecat:
cc @hugo-vrijswijk

modules/service/src/test/scala/lucuma/odb/graphql/OdbSuite.scala
```diff
@@ -375,7 +375,7 @@ abstract class OdbSuite(debug: Boolean = false) extends CatsEffectSuite with Tes
     query:     String,
     expected:  Either[List[String], Json],
     variables: Option[JsonObject] = None,
-    client:    ClientOption = ClientOption.Ws,
+    client:    ClientOption = ClientOption.Http,
```
swalker2m, line 378: Are we going back here?  We might need to while `Ws` hides the problems.  In my PR I added a `ClientOption.Default` which is used as the default value in these methods to make it easy to switch.

swalker2m (approved):
This looks great and I think the approach makes sense.  I suspect it will be fine.  :+1:

PR #1186, opened 2024-05-22
simplify program user model

tpolecat:
The current model associates collaborators with programs in 4 possible roles:

- CoI
- Observer (poorly named; the meaning here is that it's a read-only user)
- Support (staff)
- Support (NGO, for a specific partner)

After discussing with Andy it turns out we can simplify to three roles:

- Coi (same as before)
- Coi/Read-Only (formerly Observer)
- Support (no staff/NGO distinction)

So this PR makes these changes. There are more to come but this was a good place to pause.


swalker2m (approved):
This does look more simple.  :+1:

Just in case, I've had issues renaming enums in the past unless they are put in a separate migration (i.e., separate transaction).

tpolecat:
> Just in case, I've had issues renaming enums in the past unless they are put in a separate migration (i.e., separate transaction).

:+1: Ok, I will test the migration from dev and see what happens.

tpolecat:
I blew away my local database and populated it from dev, then I ran the server and it came up ok. I checked the enum and the value had been renamed as expected. So I think we're good.

PR #1187, opened 2024-05-27
CfP Deadline

swalker2m:
Currently there is no way to set the submission deadline for calls not limited to specific partners (for example Demo Science).  This PR reworks partners and submission deadlines to address this issue as follows:

* Removes time accounting categories like "Guaranteed Time" from the partners enum.  We'll need to create a separate time accounting category enum.
* When not specified, _all_ partners are now assumed when a new CfP is created.  For example, a new Demo Science CfP without a `partners` array in the input is created with all partners instead of no partners.
* There is now a top-level submission deadline default.
* Each partner may supply a submission deadline override.
* A partner's submission deadline becomes its override (if specified) and otherwise the top-level default (if specified).  When no deadline is available for a partner, it will be assumed closed.

This makes it easy to create a new CfP and specify a single submission deadline to apply to all partners by default while still permitting individual partners to have distinct deadlines.


toddburnside (approved):
LGTM

PR #1188, opened 2024-05-27
migration test procedure

swalker2m:
The procedure for testing migrations is a bit tricky and I always struggle to remember the steps.   This PR expands on the current process.  It's as easy as 1, 2, 3 .... 4, 5, 6, 7.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,10 +69,60 @@ There is also a required environment variable that sets the maximum file upload 
 ## Testing Migrations
 
 If you're adding migration scripts you should ensure that they will run correctly on a populated
-database. You can do this by bringing the database up via `docker-compose up` and then overwriting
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -69,10 +69,60 @@ There is also a required environment variable that sets the maximum file upload
...
+```
+psql -h localhost -U jimmy -d postgres
+```
+
+then
+
+```
+# drop database "lucuma-odb";
+# create database "lucuma-odb";
+^C
```
swalker2m, line 110: I suppose it might be that dropping and recreating the database here obviates the need to toggle `down` / `up` in step 4.  I'm too demoralized to try at the moment, but will the next time.

toddburnside (approved):
LGTM. Thanks for taking the time to document this.

tpolecat:
I have a script that does this, but it doesn't work in CI because computers.

PR #1189, opened 2024-05-29
Abandon Atoms / Steps

swalker2m:
This PR is a response to a request from @rpiaggio to know when an atom has been "abandoned" in terms of completed step/atom matching.  In order to count as successfully executed, all the steps in an atom have to complete without interruption from foreign atom events.  If a step from another atom appears in the event stream, the current atom is considered "abandoned".  To make this visible to the API and clients, I added `ABANDONED` step and atom execution states.  I also updated the logic to not update execution states once a terminal state (such as `ABORTED` or `COMPLETED`) been reached.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/ExecutionEventService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/ExecutionEventService.scala
@@ -118,6 +118,7 @@ object ExecutionEventService {
           e <- ResultT(insert)
           (eid, time, oid, vid) = e
           _ <- ResultT.liftF(services.sequenceService.setAtomExecutionState(atomId, atomStage))
+          _ <- ResultT.liftF(services.sequenceService.abandonOngoing(oid, atomId))
           _ <- ResultT.liftF(timeAccountingService.update(vid))
         } yield AtomEvent(eid, time, oid, vid, atomId, atomStage)).value
       }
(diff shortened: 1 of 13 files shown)


modules/service/src/main/scala/lucuma/odb/service/ExecutionEventService.scala
```diff
@@ -171,6 +172,16 @@ object ExecutionEventService {
...
+        extension (s: SequenceCommand)
+          def isTerminal: Boolean =
+            s match {
+              case SequenceCommand.Abort    |
+                   SequenceCommand.Stop       => true
+              case SequenceCommand.Continue |
+                   SequenceCommand.Pause    |
+                   SequenceCommand.Start      => false
+            }
+
```
tpolecat, line 184: I suppose this should move into core at some point.

tpolecat (approved):
This looks good. Note that you will need to renumber `V0875__xxx` in this PR or in #1096, depending on which is merged first. 👍🏻

modules/service/src/main/scala/lucuma/odb/service/ExecutionEventService.scala
swalker2m: Yes thanks.  I'll do it now and wait for the next core update to take advantage of it.

PR #1198, opened 2024-05-29
ShortCut 2887: Limit Forever Sequences

swalker2m:
Puts an arbitrary 1,000 atom limit on sequences (cleared with Andy). I'd started with 100,000 but even with a short 0.5 second exposure time and one step per atom, that is a multi-night observation.


toddburnside (approved):
LGTM

PR #1206, opened 2024-05-30
collaborator partners

tpolecat:
This adds a `partner` field to `ProgramUser` and `UserInvitation`, which is required for `Coi` and `CoiRO` users (and illegal for `Support` users).


modules/service/src/main/resources/db/migration/V0875__collaborator_partner.sql
```diff
@@ -0,0 +1,22 @@
+
+
+-- add partner to t_program_user, defaulting existing non-support program users to 'us'
```
swalker2m, line 3: Just a heads up that this'll need to be renumbered to 877.

modules/service/src/main/scala/lucuma/odb/service/UserInvitationService.scala
```diff
@@ -164,47 +166,51 @@ object UserInvitationService:
...
 
-    val createPiInvitation: Query[(User, Program.Id, EmailAddress, ProgramUserRole), UserInvitation] =
+    val createInvitationAsPi: Query[(User, Program.Id, EmailAddress, ProgramUserRole, Partner), UserInvitation] =
       sql"""
         select insert_invitation(
           $user_id,
           $program_id,
           $email_address,
           $program_user_role,
+          $tag,
```
swalker2m, line 195: The partner enum in the database turned into a `t_partner` table and I removed the time accounting categories that were mixed in.  It now matches the `lucuma.core.enum.Partner` (which was moved from `core.model`) and there's a `partner` codec.  It may need to go back to being a dynamic enum at some point, but at the moment we can avoid the `Tag` stuff.

swalker2m (approved):
Looks great, pending rebase.  :+1:

PR #1208, opened 2024-05-30
max connections

tpolecat:
This adds a config variable `ODB_MAX_CONNECTIONS` with a reasonable default. We still need to be sure we don't scale over the connection limit but I think we'll be fine for a while.


swalker2m (approved):
Looks great.  :+1:

PR #1215, opened 2024-06-05
grackle-0.20

tpolecat:
Ok this updates Grackle to accommodate fixes in 0.20
- Interface field mappings defined on the supertype don't need to be (and can't be) redefined on subtypes
- Json results are now "opaque" in the sense that Grackle won't try to hop into other mappings while traversing a json structure. We had implemented this with a hack that also added null fields that exist in the schema but not in the json result, but the built-in mapping doesn't do this. So to remove our hack I also needed to add explicit null fields to the encoders.
- I'm leaving the tests on Http because Ws just hangs on many kinds of errors.


modules/service/src/main/scala/lucuma/odb/graphql/mapping/ProposalTypeMapping.scala
```diff
@@ -35,7 +35,7 @@ trait ProposalTypeMapping[F[_]] extends BaseMapping[F]
       discriminator = proposalTypeDiscriminator,
       fieldMappings = List(
         SqlField("id", ProposalView.ProgramId, key = true, hidden = true),
-        SqlField("scienceSubtype", ProposalView.ScienceSubtype, discriminator = true),
+        SqlField("scienceSubtype", ProposalView.ScienceSubtype, discriminator = true, hidden = true),
```
swalker2m, line 38: Do discriminators need to be marked `hidden` even if they appear in the GraphQL schema?

modules/service/src/main/scala/lucuma/odb/graphql/mapping/ExecutionEventMapping.scala
```diff
@@ -28,7 +28,7 @@ trait ExecutionEventMapping[F[_]] extends ExecutionEventTable[F]
 
   lazy val ExecutionEventMapping: ObjectMapping =
     SqlInterfaceMapping(ExecutionEventType, executionEventTypeDiscriminator)(
-      SqlField("id",           ExecutionEventTable.Id, key = true),
+      SqlField("id",           ExecutionEventTable.Id, key = true, hidden = true),
```
swalker2m, line 31: Same here.  `id` appears in the `interface` definition.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/ProposalTypeMapping.scala
tpolecat: The validator isn't complaining anymore so I think maybe this was a bug that was inadvertently fixed? I'll switch them back where I changed them.

swalker2m: :+1: I wonder if it is a bug that the validator _doesn't_ complain about a hidden mapping for something that appears in the schema.

tpolecat: Ah good point.

swalker2m:
> Json results are now "opaque" in the sense that Grackle won't try to hop into other mappings while traversing a json structure. We had implemented this with a hack that also added null fields that exist in the schema but not in the json result, but the built-in mapping doesn't do this. So to remove our hack I also needed to add explicit null fields to the encoders.

It's great news that Grackle won't switch to database mappings in the middle handling a Json result.  It seems to me that Grackle should handle missing fields itself though, in a future improvement perhaps.

tpolecat:
> It's great news that Grackle won't switch to database mappings in the middle handling a Json result. It seems to me that Grackle should handle missing fields itself though, in a future improvement perhaps.

Miles argued that Json results should match the schema, which makes sense to me. Magic is bad so I'm ok doing a little extra work to stick in nulls.

swalker2m:
> Miles argued that Json results should match the schema

If the schema says a field is optional and it is not present in the results, why is that considered not matching?

tpolecat:
> If the schema says a field is optional and it is not present in the results, why is that considered not matching?

For input types we can distinguish absent from null, but I'm not sure about output types. Absent may not make sense. Maybe @milessabin could comment.

milessabin:
> I guess if Grackle is looking for optional field foo and it's not there at all, we don't know if it's (a) null or (b) missing because we forgot to include it.

That's the case here.

A couple of clarifications ...

> Interface field mappings defined on the supertype don't need to be (and can't be) redefined on subtypes

In fact they _can_ be defined on subtypes, however, if _all_ subtypes define a mapping the interface level mapping will never be used (because now the mapping is properly dynamically resolved) and Grackle will detect this and report it as a validation error. So, if you have a default case which works for some, but not all, subtypes, you can define that at the interface level and then redefine for the subtypes which need special treatment.

> If the schema says a field is optional and it is not present in the results, why is that considered not matching?

The approach I took here is that Grackle persists with the `CirceField` until the cursor drops off the end of the Json blob it's indexing into, at which point it would resume looking for other field mappings. It has dropped off the end of the json in this case, and there's no other mapping to be found, and so a missing mapping is reported.

I could arrange things so that if the search for an alternative mapping fails that's treated as if there were a nulled field in the Json blob. My only concern here is that we're than back with a certain amount of indeterminism about which mapping ends up being used. If the the json blob always contains all fields, with null values as appropriate, then it's completely deterministic as to which mapping will be used in these cases. If sometimes the null field is omitted then sometimes an alternative mapping will be used.

So this is really an ergonomics issue rather than a technical feasibility question. I would be happy to change the behaviour if you would prefer that.

swalker2m:
> The approach I took here is that Grackle persists with the CirceField until the cursor drops off the end of the Json blob it's indexing into, at which point it would resume looking for other field mappings

I am likely confused but I think I would assume that if the field is missing in the Json blob and it is optional, that just means it is simply not present in the result and so there is no need to look for any mapping.

At any rate, adding some placeholder `null`s to the result isn't really that onerous so this is fine.

PR #1221, opened 2024-06-10
rewrite target source profile blobs

tpolecat:
This adds a migration that rewrites the json blobs for source profiles, to add null fields required by Grackle. It also updates the populate-db-from-dev script to be a little bit better.


swalker2m (approved):
Excellent, and thanks for the populate db script.  :+1:

PR #1224, opened 2024-06-11
GMOS FPU nulls

swalker2m:
For latest Grackle we need to explicitly write `null` values for missing optional fields.  We had no test case, and therefore missed, an instance where `null` is required.  The GMOS FPU is either `builtin` or `customMask`.  The  odd man out needs to be explicitly `null`.

### DIFF
--- a/modules/schema/src/main/scala/lucuma/odb/json/gmos.scala
+++ b/modules/schema/src/main/scala/lucuma/odb/json/gmos.scala
@@ -208,10 +208,12 @@ trait GmosCodec {
     Encoder.instance[GmosFpuMask[A]] {
       case GmosFpuMask.Builtin(v)       =>
         Json.obj(
-          "builtin"    -> v.asJson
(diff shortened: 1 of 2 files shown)


modules/schema/src/main/scala/lucuma/odb/json/gmos.scala
```diff
@@ -208,10 +208,12 @@ trait GmosCodec {
     Encoder.instance[GmosFpuMask[A]] {
       case GmosFpuMask.Builtin(v)       =>
         Json.obj(
-          "builtin"    -> v.asJson
+          "builtin"    -> v.asJson,
+          "customMask" -> Json.Null
```
swalker2m, line 212: I remain unconvinced that Grackle should attempt to find a database mapping when processing JSON and encountering missing optional fields.

tpolecat (approved):

PR #1225, opened 2024-06-11
Add validations to observations

toddburnside:
Validates that we can create a sequence for an observation. And, if a CfP is selected, that it meets the CfP criteria for instruments and ra/dec limits.

In the future this will be expanded such the for accepted proposals, observations are validated against the ordained configuration/targets/etc.

### DIFF
--- a/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -7930,6 +7930,11 @@ type Observation {
   itc: Itc!
 
   """
+  A list of observation validation problems
+  """
+  validations: [ObservationValidation!]!
+
+  """
   Enclosing group, if any.
   """
   groupId: GroupId
@@ -7995,6 +8000,46 @@ observation index. For example, G-2024B-1234-Q-0001 where G-2024B-1234-Q is
 the program reference label and the observation index is 1.
 """
 scalar ObservationReferenceLabel
+
+"""
+An observation validation problem
+"""
+type ObservationValidation {
+  """
+  The type of validation problem
+  """
+  code: ObservationValidationCode!
+
+  """
+  Particular errors for this validation type
+  """
+  messages: [String!]!
+}
+
+"""
+Types of observation validations
+"""
+enum ObservationValidationCode {
+  """
+  The observation or targets are missing required information
+  """
+  MISSING_DATA,
+
+  """
+  The data in the observation or targets is inconsistent.
+  """
+  CONFLICTING_DATA,
+
+  """
+  The observation does not meet the requirements of the Call for Proposals
+  """
+  CFP_ERROR
+
+  """
+  The observation is missing or is not accessible
+  """
+  MISSING_OBSERVATION
+}
 
 """
 Each step in a sequence is tagged with an observe class which identifies its
--- a/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -8021,24 +8021,14 @@ Types of observation validations
 """
 enum ObservationValidationCode {
   """
-  The observation or targets are missing required information
+  The observation is not configured correctly and cannot be executed
   """
-  MISSING_DATA,
-
-  """
-  The data in the observation or targets is inconsistent.
-  """
-  CONFLICTING_DATA,
+  CONFIGURATION_ERROR,
 
   """
   The observation does not meet the requirements of the Call for Proposals
   """
   CFP_ERROR
-
-  """
-  The observation is missing or is not accessible
-  """
-  MISSING_OBSERVATION
 }
 
 """
--- a/modules/schema/src/main/scala/lucuma/odb/data/ObservationValidation.scala
+++ b/modules/schema/src/main/scala/lucuma/odb/data/ObservationValidation.scala
@@ -1,0 +1,38 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.odb.data
+
+import cats.data.NonEmptyChain
+import io.circe.Codec
+import io.circe.Decoder
+import io.circe.Encoder
+import lucuma.core.util.Enumerated
+
+enum ObservationValidationCode(
+  val tag: String,
+  val name: String,
+  val description: String
+) derives Enumerated:
+  case MissingData extends ObservationValidationCode("missing_data", "Missing Data", "Some required data is missing")
+  case ConflictingData extends ObservationValidationCode("conflicting_data", "Conflicting Data", "Conflicting data")
+  case CallForProposalsError extends ObservationValidationCode("cfp_error", "CfP Error", "Not valid for the selected Call for Proposals")
+  // we should never get this one, unless we change how the validation method is called.
+  case MissingObservation extends ObservationValidationCode("missing_observation", "Missing Observation", "The observation is missing or not accessible")
+
+case class ObservationValidation(
+  code: ObservationValidationCode,
+  messages: NonEmptyChain[String]
+) derives Codec
+
+object ObservationValidation:
+  def fromMsgs(code: ObservationValidationCode, msg: String, moreMsgs: String*): ObservationValidation =
+    ObservationValidation(code, NonEmptyChain.of(msg, moreMsgs*))
+  def missingData(msg: String, moreMsgs: String*): ObservationValidation =
+    fromMsgs(ObservationValidationCode.MissingData, msg, moreMsgs*)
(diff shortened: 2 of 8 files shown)


modules/schema/src/main/scala/lucuma/odb/data/ObservationValidation.scala
```diff
@@ -0,0 +1,38 @@
...
+
+package lucuma.odb.data
+
+import cats.data.NonEmptyChain
+import io.circe.Codec
+import io.circe.Decoder
+import io.circe.Encoder
+import lucuma.core.util.Enumerated
+
+enum ObservationValidationCode(
```
toddburnside, line 12: I'm not sure if we care to move these to lucuma-core or not. 🤷

modules/service/src/main/scala/lucuma/odb/graphql/mapping/ObservationMapping.scala
```diff
@@ -117,4 +119,18 @@ trait ObservationMapping[F[_]]
     effectHandler(readEnv, calculate)
   }
 
+  def validationsQueryHandler: EffectHandler[F] = {
```
toddburnside, line 122: I think having validations be part of the observation is very nice from a usability standpoint. The only thing I don't like is that we need to look up CfP information for each of the observations.

modules/service/src/main/scala/lucuma/odb/service/AsterismService.scala
```diff
@@ -89,7 +89,7 @@ trait AsterismService[F[_]] {
   def getAsterism(
     programId: Program.Id,
     observationId: Observation.Id
-  )(using NoTransaction[F]): F[List[(Target.Id, Target)]]
```
toddburnside, line 92: This is now called from an area that needs to have a transaction. Since this only does a query, it seems like it shouldn't matter if there is one or not, so I removed the requirement.

modules/service/src/main/scala/lucuma/odb/service/ObservationService.scala
```diff
@@ -428,6 +447,128 @@ object ObservationService {
           _             <- ResultT(asterismService.setAsterism(pid, NonEmptyList.of(newOid), input.asterism))
         yield CloneIds(origOid, newOid)).value
         
+      // ***********************
```
toddburnside, line 450: I toyed with the idea of creating a validation service instead of adding it to the observation service. I decided to go this way, but could be convinced to move it.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -7996,6 +8001,46 @@ the program reference label and the observation index is 1.
...
 scalar ObservationReferenceLabel
 
+"""
+An observation validation problem
+"""
+type ObservationValidation {
+  """
+  The type of validation problem
+  """
+  code: ObservationValidationCode!
```
cquiroz, line 8011: Do we need a separate discriminator for importance of the validation? like warning/errorr?

modules/service/src/main/scala/lucuma/odb/service/ObservationService.scala
```diff
@@ -428,6 +447,128 @@ object ObservationService {
...
+          else raStart <= ra && ra <= raEnd
+
+      extension (dec: Declination)
+        def isInInterval(decStart: Declination, decEnd: Declination): Boolean =
+          decStart <= dec && dec <= decEnd
+
+      extension (tsi: TimestampInterval)
+        def centerUnsafe: Timestamp =
+          if (tsi.isEmpty) tsi.start
+          else tsi.start.plusMicrosOption(tsi.boundedTimeSpan.toMicroseconds).get
```
cquiroz, line 468: Is it unsafe because of the `get``?

cquiroz (commented):
Seems fine we can use that on the UI

modules/service/src/main/scala/lucuma/odb/service/ObservationService.scala
toddburnside: Yes. But in practice it should always be OK.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
toddburnside: I think we may want that in the future, but everything now would be an error I think. I figured we could add another discriminator when we get things that are other than errors. It would be a non breaking change.

modules/service/src/main/scala/lucuma/odb/service/ObservationService.scala
swalker2m: Wouldn't the center be half of the interval?  I'm probably missing the obvious.

toddburnside: Oh. 🤦 I forgot to divide by 2. I got too involved in figuring out where to get the data from...

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
swalker2m: What do we plan on doing with the code in the UI, if anything?  Feels like we'll be adding new codes frequently.

modules/service/src/main/scala/lucuma/odb/service/ObservationService.scala
```diff
@@ -428,6 +447,128 @@ object ObservationService {
...
+      extension (dec: Declination)
+        def isInInterval(decStart: Declination, decEnd: Declination): Boolean =
+          decStart <= dec && dec <= decEnd
+
+      extension (tsi: TimestampInterval)
+        def centerUnsafe: Timestamp =
+          if (tsi.isEmpty) tsi.start
+          else tsi.start.plusMicrosOption(tsi.boundedTimeSpan.toMicroseconds / 2).get
+
+      extension (ge: GeneratorParamsService.Error)
```
swalker2m, line 470: If I'd written `GeneratorParamsService` today, I probably would have just had it return a `Result` instead of a specific `Error`.  In fact, I've been converting some of the older ones that return service-specific `Error` messages to just return `Result`.  I wonder if the `Error`s are needed after all.

toddburnside: The errors are useful here for reporting to the user. But, if it returned a Result we could just extract the error messages from there. We would just have to change the ObservationValidationCodes. Maybe just `Configuration Error`.

modules/service/src/main/scala/lucuma/odb/service/ObservationService.scala
```diff
@@ -916,6 +1057,42 @@ object ObservationService {
         FROM t_observation
         WHERE c_observation_id IN (
       """.apply(gid, index) |+| which |+| void")"
+
+    def getProgramCfpId(pid: Program.Id): AppliedFragment =
+      sql"""
+        SELECT c_cfp_id
+        FROM t_proposal
+        WHERE c_program_id = $program_id
+      """.apply(pid)
```
swalker2m, line 1066: A minor style thing but what is your opinion on making this, say:

```scala
val SelectCfp: Query[Program.Id, CallForProposals.Id] =
  sql"""
    SELECT c_cfp_id
    FROM t_proposal
    WHERE c_program_id = $program_id
  """.query(cfp_id)
```

Then you can call it with a one-liner like `session.option(Statements.SelectCfp)(pid)`.  I've been trying to do it that way when possible since `AppliedFragment` doesn't really signal what the parameters and output are expected to be.

toddburnside: That does seem clearer. I'll switch.

swalker2m: I guess some of these could be moved to the `CallForProposalsService` to keep them together with other CfP stuff.

swalker2m (approved):
This is great, I'm glad you took this on.  I think making in an on-demand effect field makes sense to start.  I wonder if we'll ultimately need to cache these along with all the headaches that brings.  But, if so we can switch later.  :+1:

modules/service/src/main/scala/lucuma/odb/service/ObservationService.scala
toddburnside: I thought about that, but the queries seem pretty specific to the task at hand. If we move them, we'd probably want to make them more general, which would require a CallForProposals class, which we currently don't have. That said, I'm not opposed to doing that.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
toddburnside: For now, we'll just use it to display the message for the validation types. Eventually, we may be able to offer "Quick Fixes". A suggestion Rob made was, when we have, in the future, unapproved changes for accepted proposals we could offer "Request Approval". I'm hoping we don't often add new codes, just new messages within existing codes, but we'll see...

PR #1228, opened 2024-06-12
Science Sequence Reset

swalker2m:
When moving to a new visit, the nighttime arc policy prescribes new arc datasets for each unique science configuration.  This will necessitate the generation of new atom and step ids for the science sequence in each distinct visit.  This PR makes that small adjustment.


toddburnside (approved):
LGTM

PR #1229, opened 2024-06-12
Move CallForProposalsType to schemas project

toddburnside:


modules/schema/src/main/scala/lucuma/odb/data/CallForProposalsType.scala
```diff
@@ -5,18 +5,11 @@ package lucuma.odb.data
```
swalker2m, line 5: This will split the `lucuma.odb.data` package across two modules.  I wonder if that is something to concern us?

tpolecat: If the goal here is to make it available to clients maybe it should move to core?

toddburnside: It already IS split across 2 modules. It started with OdbError, but I have been adding to it... I could move this, as well as ObservationValidation and ObservationValidationCode to lucuma-core instead.

toddburnside:
Will move to lucuma-core instead.

PR #1230, opened 2024-06-12
Allow multiple cors domains

cquiroz:
This is to allow navigate to run in `gemini.edu` while still in development


tpolecat (approved):
I don't understand any of this. 👍🏻

PR #1239, opened 2024-06-14
Target Calib role

cquiroz:
This is a first basic PR on the road to calibrations (and also to get acquainted with odb/grackle)

Targets get a extra optional column to indicate if they have a calibration role. This field is not editable in the API and can only be set directly in the API.

Additionally the Target field `role` was dropped


swalker2m (approved):
This code all looks great to me.  A few questions about future plans:

1.  Maybe for @toddburnside (?), I think you added the old target role.  Is it clear it is no longer needed?
2. Will there be API to set the calibration role?
3. We'll have to keep up with this flag in future operations, right?  Like when we copy a target we'll have to remember to reset the flag?
4. Do we need to mark the individual targets or would it be enough to see that it is in a telluric program?  If it weren't for the (understandable) desire to make them editable in Explore, we wouldn't need to expose the these targets at all?  I guess though we need to be able to treat them differently inside a normal program.

toddburnside:
> 1. Maybe for @toddburnside (?), I think you added the old target role.  Is it clear it is no longer needed?

Originally this was going to distinguish guide targets from science targets. But, we ended up not saving guide targets, so it isn't needed.

cquiroz:
>     2. Will there be API to set the calibration role?
As I see it no, Calibration programs would assign the role to its targets auitomatically

>     3. We'll have to keep up with this flag in future operations, right?  Like when we copy a target we'll have to remember to reset the flag?
cloning targets already carry the flag, I think the flag would not change as we need to remember the target is for calibrations

>     4. Do we need to mark the individual targets or would it be enough to see that it is in a telluric program?  If it weren't for the (understandable) desire to make them editable in Explore, we wouldn't need to expose the these targets at all?  I guess though we need to be able to treat them differently inside a normal program.
Yes we want to expose them in a program so we can edit them in the same UI, I think the UI would need to change a little
In normal programs we may not show them or show them in a limited way

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -10196,6 +10223,11 @@ type Target {
   """
   name: NonEmptyString!
 
+  """
+  calibration role
+  """
+  calibrationRole: CalibrationRole
```
toddburnside, line 10229: By specifying this as a `calibration role` rather than just a generic `role`, are we fairly certain that we won't have any other types of roles for targets in the future? I guess if we do, we could relax the name in the future.

toddburnside:
> 4\. Do we need to mark the individual targets or would it be enough to see that it is in a telluric program?  If it weren't for the (understandable) desire to make them editable in Explore, we wouldn't need to expose the these targets at all?  I guess though we need to be able to treat them differently inside a normal program.

That's an interesting idea. I'm trying to think how that would work. In explore, we need to see the targets, not only to edit them in calibration program, but also to be able to display them as part of the calibration observations. Since those observations won't be editable (I think/hope), the targets wouldn't be either. I think we want to avoid showing the calibration targets on the targets tab in explore. I'm not sure what the plans are for the API for these targets, but maybe the existing `targets` query would only provide the targets that are actually part of the program, and a different query endpoint could provide "calibration targets associated with the calibration observations of this program"?

The current API for editing might be OK, because in order to edit any target you need to have authorization to edit the program it is in. And PIs wouldn't have that. The API will also need to check if any observations using the calibration target have been executed, but that is the same as for regular targets. The more complex bit is in the UI for editing these programs. It needs to know ALL observations, regardless of program, to which these targets are assigned so it can check for executed status of any of them and clone the target to assign to the unexecuted observations (if necessary) regardless of program.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
cquiroz: It is possible I wanted to use this specifically for the `CalibrationRole` enumeration

modules/service/src/main/resources/db/migration/V0880__calibration_reference_data.sql
```diff
@@ -0,0 +1,30 @@
...
+
+alter table t_target
+add column c_calibration_role e_calibration_role null;
+
+-- Drop constraints/trigger that are no longer needed because c_role is dropped
+alter table t_asterism_target 
+drop constraint asterism_target_is_science;
+
+drop trigger t_target_role_immutable_trigger on t_target;
+
```
toddburnside, line 16: Should we also get rid of the old `e_target_role`?

cquiroz: Good observation, I'll remove it

toddburnside (approved):
LGTM

PR #1247, opened 2024-06-18
add allocations field on type Program

tpolecat:
An earlier oversight, which Bryan noticed.

I'm guessing time allocation will end up being much more complicated than what I implemented, but we'll see.


swalker2m (approved):
Looks great. :+1:

PR #1248, opened 2024-06-18
Fix internal error in observation validation when proposal has no cfp

toddburnside:
Andy was kind enough to find this stupid mistake for me on the first try.


cquiroz (approved):
LGTM

PR #1254, opened 2024-06-19
Configuration variable to allow CORS over plain http

cquiroz:
In some cases, local dev, it would be useful to support CORS over plain http
This PR adds support for that defaulting to requiring CORS

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/Config.scala
+++ b/modules/service/src/main/scala/lucuma/odb/Config.scala
@@ -37,16 +37,17 @@ import java.util.UUID
 import scala.concurrent.duration.*
 
 case class Config(
-  port:       Port,             // Our port, nothing fancy.
-  itc:        Config.Itc,       // ITC config
-  sso:        Config.Sso,       // SSO config
-  serviceJwt: String,           // Only service users can exchange API keys, so we need a service user JWT.
-  honeycomb:  Config.Honeycomb, // Honeycomb config
-  database:   Config.Database,  // Database config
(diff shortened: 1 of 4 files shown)


swalker2m:
Since all I know about Coors is that it is kind of a cheap beer, if this is merged and I do nothing, will everything continue to work for development the way it always has?

toddburnside:
What are you having trouble with in local dev, Carlos?

cquiroz:
playground doesn't work locally as we have plain http by default
Also @dngomez  is having some troubles with CORS for development of navigate

cquiroz:
> Since all I know about Coors is that it is kind of a cheap beer, if this is merged and I do nothing, will everything continue to work for development the way it always has?

The default is the same as now, but you have flexibility to change the way it works

toddburnside:
> playground doesn't work locally as we have plain http by default

Hmm. Interesting. Playground works for me at http://localhost:8082/playground.html

toddburnside (approved):
Seems fine to me.

swalker2m:
> Hmm. Interesting. Playground works for me at http://localhost:8082/playground.html

toddburnside:
And in explore, I use:

>    "odbURI": "ws://localhost:8082/ws",
    "odbRestURI": "http://localhost:8082",

Is Diego having trouble from server to server? Or navigate UI to ODB server?

cquiroz:
> Is Diego having trouble from server to server? Or navigate UI to ODB server?
Navigate UI to server
Note that ws does't use CORS that's why it works with explore

Issue #1237, opened 2024-06-13
Unique Constraint Violation on CfP Update


swalker2m:
![image](https://github.com/gemini-hlsw/lucuma-odb/assets/4906023/7f2e6710-ad4b-42e8-8d81-dfb1b4074665)

There is an `update_program_type` trigger fired when `t_program` is updated.  It doesn't update the semester index on unsubmitted proposals -- and in fact there is no semester index until a proposal is submitted.  We think this proposal was previously submitted, un-submitted, and then the CfP was updated.

swalker2m:
Fixed in #1260.

PR #1260, opened 2024-06-26
Fix GH issue 1237

swalker2m:
Fixes the problem reported in #1237.  It's a rare edge case.

When a proposal is successfully submitted it gains a sequential semester index.  A desired use case is to be able to un-submit the proposal, make an edit, and then resubmit it without changing the index.  For that reason, when un-submitting we don't change the index.

When a CfP update changes the semester _after_ a proposal has been submitted then un-submitted though, the previous index is no longer valid and must be removed.  The proposal will subsequently gain a new index when it is ultimately submitted.

### DIFF
--- a/modules/service/src/main/resources/db/migration/V0884__update_program_type.sql
+++ b/modules/service/src/main/resources/db/migration/V0884__update_program_type.sql
@@ -67,6 +67,18 @@ BEGIN
           NEW.c_instrument   := NULL;
           NEW.c_library_desc := NULL;
         END;
+
+      WHEN NEW.c_program_type = 'system' THEN
+        BEGIN
+          IF NEW.c_library_desc IS NULL THEN
+            RAISE EXCEPTION 'System programs must define a description';
+          END IF;
+          NEW.c_instrument      := NULL;
+          NEW.c_semester        := NULL;
+          NEW.c_semester_index  := NULL;
+          NEW.c_science_subtype := NULL;
+        END;
+
     END CASE;
 
     RETURN NEW;
--- a/modules/service/src/main/resources/db/migration/V0881__update_program_type.sql
+++ b/modules/service/src/main/resources/db/migration/V0881__update_program_type.sql
@@ -1,0 +1,74 @@
+-- Update the update_program_type function to handle COM and MON
+CREATE OR REPLACE FUNCTION update_program_type()
+RETURNS TRIGGER AS $$
+BEGIN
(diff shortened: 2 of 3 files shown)


modules/service/src/main/resources/db/migration/V0884__update_program_type.sql
```diff
@@ -0,0 +1,86 @@
...
+            IF NEW.c_semester IS NULL THEN
+              RAISE EXCEPTION 'Submitted science programs must define a semester';
+            ELSEIF NEW.c_science_subtype IS NULL THEN
+              RAISE EXCEPTION 'Submitted science programs must define a science subtype.';
+            ELSEIF (NEW.c_program_type != OLD.c_program_type) OR
+                   (NEW.c_semester IS DISTINCT FROM OLD.c_semester) OR
+                   (OLD.c_proposal_status = 'not_submitted' AND NEW.c_semester_index IS NULL) THEN
+              NEW.c_semester_index := next_semester_index('science', NEW.c_semester, NULL);
+            END IF;
+          ELSEIF (NEW.c_semester IS DISTINCT FROM OLD.c_semester) THEN
```
swalker2m, line 62: This branch is the new code.  It resets the index when a proposal is not submitted and yet the semester has changed.

modules/service/src/main/resources/db/migration/V0884__update_program_type.sql
```diff
@@ -0,0 +1,86 @@
...
+            -- Since it is not submitted and the semester has changed, we lose
+            -- any index we may have had while previously submitted
+            NEW.c_semester_index := NULL;
+          END IF;
+          NEW.c_instrument   := NULL;
+          NEW.c_library_desc := NULL;
+        END;
+
+      WHEN NEW.c_program_type = 'system' THEN
```
swalker2m, line 71: @cquiroz  I moved this down to keep the `WHEN` branches alphabetical and `null`-ed out the `c_instrument` field which doesn't apply to `SYS`.

cquiroz: Is it a new migration?

swalker2m: Yes, flyway will reject updates to existing migrations and the database startup will fail.

cquiroz (approved):
LGTM

PR #1263, opened 2024-06-27
Calib service scaffolding

cquiroz:
In this PR I add a new project that will run the calibration-service in a separate worker process in heroku

* A single instance can run at anytime
* It will share the same environment than odb thus able to use the same  db and have access to the service token
* At the moment it just listen for project events and  reads the observation configurations

It wont be enabled in heroku just yet but it was tested in a separate app


modules/calibrations/src/main/scala/lucuma/odb/calibrations/Main.scala
```diff
@@ -0,0 +1,177 @@
...
+  header = MainParams.Header
+) {
+
+  override def main: Opts[IO[ExitCode]] =
+    Opts(serve)
+
+  lazy val serve: IO[ExitCode] = {
+    given Logger[IO] = Slf4jLogger.getLoggerFromName[IO]("calibrations-service")
+
+    import natchez.Trace.Implicits.noop
```
cquiroz, line 67: I'm not sure if we'd need tracing here.

modules/calibrations/src/main/scala/lucuma/odb/calibrations/Main.scala
```diff
@@ -0,0 +1,177 @@
...
+              calibrationsService.recalculateCalibrations(elem.programId)
+            }.compile.drain.start.void)
+    } yield ()
+
+  def services[F[_]: Concurrent: UUIDGen: Trace: Logger](
+    user: Option[User],
+    enums: Enums
+  )(pool: Session[F]): F[Services[F]] =
+    user match {
+      case Some(u) if u.role.access === Access.Service =>
```
cquiroz, line 145: We either have a valid user or we quit

modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
```diff
@@ -0,0 +1,85 @@
...
+  )(using L: Logger[F]): F[Unit]
+
+}
+
+object CalibrationsService {
+  def instantiate[F[_]: FlatMap](using Services[F]): CalibrationsService[F] =
+    new CalibrationsService[F] {
+      def recalculateCalibrations(pid: Program.Id)(using L: Logger[F]): F[Unit] =
+        for {
+          gnls   <- session.execute(Statements.SelectGmosNorthLongSlitConfigurations)(pid)
```
cquiroz, line 35: For now we just need the configurations, but we need to add in the future the actual work

modules/calibrations/src/main/scala/lucuma/odb/calibrations/Main.scala
tpolecat: I think this is fine.

tpolecat (approved):
This looks reasonable. I will probably do something similar for approved configurations and may have more comments then. 👍🏻

cquiroz:
Would we want to run everything as a separate worker or one worker with multiple tasks (calibrations, approved configurations, checks, etc...)

tpolecat:
> Would we want to run everything as a separate worker or one worker with multiple tasks (calibrations, approved configurations, checks, etc...)

It makes sense to me to have multiple small workers, but I haven't thought about it that much. Seems simpler to me.

PR #1266, opened 2024-06-28
Add system attribute to groups

cquiroz:
This is a new attribute so we can create groups by the system, e.g. for calibrations

Regular users cannot update system calibrations


modules/service/src/main/scala/lucuma/odb/graphql/predicate/GroupPredicates.scala
```diff
@@ -7,8 +7,9 @@ import grackle.Path
...
 
 class GroupPredicates(path: Path) {
-  lazy val id       = LeafPredicates[Group.Id](path / "id")
-  lazy val parentId = LeafPredicates[Group.Id](path / "parentId")
-  lazy val program  = ProgramPredicates(path / "program")
+  lazy val id        = LeafPredicates[Group.Id](path / "id")
+  lazy val parentId  = LeafPredicates[Group.Id](path / "parentId")
+  lazy val program   = ProgramPredicates(path / "program")
   lazy val existence = ExistencePredicates(path / "existence")
+  lazy val system    = ExistencePredicates(path / "system")
```
swalker2m, line 14: `ExistencePredicates`?  I would have expected `LeafPredicates`, I think.  This raises the question of what we should do for these system / calibration role properties that we're adding wrt to queries.

modules/service/src/main/resources/db/migration/V0885__group_system.sql
```diff
@@ -0,0 +1,14 @@
...
+
+-- need to re-create this view to pick up the new column
+drop view v_group;
+create view v_group as
+  select *,
+  case when c_min_interval is not null then c_group_id end as c_min_interval_id,
+  case when c_max_interval is not null then c_group_id end as c_max_interval_id
+  from t_group;
+
+create unique index unique_system_group_name on t_group (c_program_id, c_name) where c_system is true;
```
swalker2m, line 13: Is the idea to look up system groups by specific names?  Will there be a lot of different kinds -- maybe one per calibration type?  I suppose we could add an enum with the options, along with a value for user-manipulated groups and convert `system` to the enum.  Maybe this was already discussed and decided, which is fine if so.

modules/service/src/main/scala/lucuma/odb/graphql/predicate/GroupPredicates.scala
cquiroz: You're right, it has to be a `LeafPredicates` and used in the context of `IsNull`

modules/service/src/main/resources/db/migration/V0885__group_system.sql
cquiroz: I expect a single calibration group with all of them combined. This `index` is to ensure that the first group we find with a specific name is the one we want.

modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
```diff
@@ -3,39 +3,82 @@
...
+
+      def recalculateCalibrations(pid: Program.Id)(using Transaction[F]): F[Unit] =
         for {
-          gnls   <- session.execute(Statements.SelectGmosNorthLongSlitConfigurations)(pid)
-          gsls   <- session.execute(Statements.SelectGmosSouthLongSlitConfigurations)(pid)
-          groups <- groupService.selectGroups(pid)
-          _      <- Logger[F].info(groups.toString)
+          gnls <- session.execute(Statements.SelectGmosNorthLongSlitConfigurations)(pid)
+          gsls <- session.execute(Statements.SelectGmosSouthLongSlitConfigurations)(pid)
+          _    <-  calibrationsGroup(pid, gnls.size + gsls.size)
```
swalker2m, line 81: These unique configurations will be used to create the actual calibrations in the future correct?

modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
```diff
@@ -3,39 +3,82 @@
...
+                case None      =>
+                  groupService.createGroup(
+                      input = CreateGroupInput(
+                        programId = pid.some,
+                        proposalReference = none,
+                        programReference = none,
+                        SET = GroupPropertiesInput.Create(
+                          name = CalibrationsGroupName.some,
+                          description = CalibrationsGroupName.some,
+                          minimumRequired = none,
```
swalker2m, line 61: Should the `minimumRequired` be the `size`?

modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
```diff
@@ -3,39 +3,82 @@
...
+                          ordered = false,
+                          minimumInterval = none,
+                          maximumInterval = none,
+                          parentGroupId = none,
+                          parentGroupIndex = none,
+                          existence = Existence.Present
+                        )
+                      ),
+                    system = true
+                  ).map(_.toOption)
```
swalker2m, line 71: This will return a failure result into a `None` of course.  I guess it shouldn't fail, but if it did it would be silent?  I don't know, would it make sense to pass the `Result` up to the top?

swalker2m (approved):
Seems like a good next step. :+1:

modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
cquiroz: Yes, that's correct we create a new obs for each unique configuration

modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
cquiroz: Not sure what `mimimumRequired` is

modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
cquiroz: Right, these run in a separate daemon thus there is nobody to report too, except log

PR #1272, opened 2024-07-02
CfP Default RA/Dec Limits

swalker2m:
This PR changes the `CallForProposals` schema to:

* make RA/Dec limits site-specific
* always store RA/Dec limits, either explicitly specified by the user or calculated by default
* change the active period from specific timestamps to dates (which can then be interpreted at each site to specific times)

It uses the algorithm suggested by Andy to compute the RA/Dec limits when not explicitly provided:

>  A simple suggestion would be to set the RA limits to half an hour after the LST at evening twilight on the first night of the semester and half an hour before the LST at morning twilight on the last night of the semester.

where in general "semester" refers to the active period for the call.


modules/schema/src/main/scala/lucuma/odb/data/DateInterval.scala
```diff
@@ -0,0 +1,35 @@
...
+import cats.Eq
+import cats.syntax.order.*
+import org.typelevel.cats.time.*
+
+import java.time.LocalDate
+
+/**
+ * Interval between a start date and an end date.
+ */
+sealed class DateInterval private (val start: LocalDate, val end: LocalDate):
```
swalker2m, line 15: I'll move this to `core` when this gets merged.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -453,40 +434,26 @@ input CallForProposalsPropertiesInput {
...
-  raLimitStart: RightAscensionInput
+  coordinateLimits: SiteCoordinateLimitsInput
 
   """
-  Optional RA limit end RA. Required only if raLimitStart is specified. Nullable
-  on edit if raLimitStart is also set to 'null'.
+  Active period start date (inclusive) for this call.  Required on create.  The
+  active period for this call will start on the night of the provided `Date`
+  value. Must be before the `activeEnd` date.  Not nullable.  Limited to dates
+  between 1900 and 2100 (exclusive).
```
swalker2m, line 446: The date range seems kind of arbitrary, but is necessary in order to use our LST calculation which is limited to those years.

modules/service/src/main/scala/lucuma/odb/syntax/instrument.scala
```diff
@@ -4,12 +4,37 @@
...
 import lucuma.core.enums.Instrument
+import lucuma.core.enums.Site
 
 trait ToInstrumentOps {
 
   extension (self: Instrument) {
     def fieldName: String =
       self.tag.updated(0, self.tag.charAt(0).toLower)
+
+    def site: Set[Site] =
```
swalker2m, line 15: This is a bit sketchy but should be fine for now.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
cquiroz: It will be close for release time

modules/service/src/main/resources/db/migration/V0885__cfp_limits.sql
```diff
@@ -0,0 +1,132 @@
...
+  ADD COLUMN c_north_ra_end    d_angle_µas,
+  ADD COLUMN c_north_dec_start d_angle_µas,
+  ADD COLUMN c_north_dec_end   d_angle_µas,
+  ADD COLUMN c_south_ra_start  d_angle_µas,
+  ADD COLUMN c_south_ra_end    d_angle_µas,
+  ADD COLUMN c_south_dec_start d_angle_µas,
+  ADD COLUMN c_south_dec_end   d_angle_µas;
+
+UPDATE t_cfp
+   SET c_north_ra_start  = COALESCE(c_ra_start,             0),
```
cquiroz, line 17: Didn't know about `COALESCE`, is it so that we basically see the old value but have a default if null?

modules/service/src/main/resources/db/migration/V0885__cfp_limits.sql
```diff
@@ -0,0 +1,132 @@
...
+  DROP COLUMN c_ra_start,
+  DROP COLUMN c_ra_end,
+  DROP COLUMN c_dec_start,
+  DROP COLUMN c_dec_end;
+
+-- Change the active period to dates rather than specific timestamps.
+ALTER TABLE t_cfp
+  ADD COLUMN c_active_start date,
+  ADD COLUMN c_active_end   date,
+  ADD CONSTRAINT active_dates_check CHECK (c_active_start <= c_active_end);
```
cquiroz, line 44: Can a Cfp start and end at the same time?

modules/service/src/main/resources/db/migration/V0885__cfp_limits.sql
swalker2m: Right, it chooses the first non-null value.

modules/service/src/main/scala/lucuma/odb/graphql/input/CoordinateLimitsInput.scala
```diff
@@ -0,0 +1,137 @@
...
+
+object CoordinateLimitsInput {
+
+  // From Andy: A simple suggestion would be to set the RA limits to half an
+  // hour after the LST at evening twilight on the first night of the semester
+  // and half an hour before the LST at morning twilight on the last night of
+  // the semester.
+  private val Buffer = Duration.ofMinutes(30L)
+
+  private def toRa(z: ZonedDateTime): RightAscension =
```
cquiroz, line 31: Why is it `ZonedDateTime` I thought we'd mostly use UTC

modules/service/src/main/scala/lucuma/odb/service/ObservationService.scala
```diff
@@ -490,16 +491,22 @@ object ObservationService {
...
+          active: DateInterval
         ): F[Option[ObservationValidation]] = 
           asterismService.getAsterism(pid,oid)
             .map { l =>
               val targets = l.map(_._2)
               // The lack of a target will get reported in the generator errors
+              val start    = ObservingNight.fromSiteAndLocalDate(site, active.start).start
+              val end      = ObservingNight.fromSiteAndLocalDate(site, active.end).end
+              val duration = Duration.between(start, end)
+              val center   = start.plus(duration.dividedBy(2L))
```
cquiroz, line 504: Are we calculating the position at the middle of the night? Is this just for some validations?

swalker2m: Yeah this is the conversion of the old absolute `TimestampInterval` active period check to use a `DateInterval` applied to a specific site.

tpolecat (approved):
This looks great. It continues to be amazing how complex this kind of thing ends up being once you get into the details. 👍🏻

PR #1275, opened 2024-07-03
Initial creation of spectrophotometric standard observations

cquiroz:
This PR is the initial implementation of adding calibration observations
They are based on the rules provided of creating observations for each distinct configuration.

Observations are created with a `calibrationRole` which can be queried but not modified. They are placed on the system calibrations group.

A new test was added to properly verify that the calibration service is doing what is expected


modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
```diff
@@ -31,13 +49,15 @@ trait CalibrationsService[F[_]] {
   def recalculateCalibrations(
     pid: Program.Id
   )(using Transaction[F]): F[Unit]
-
 }
 
 object CalibrationsService {
-  def instantiate[F[_]: Monad](using Services[F]): CalibrationsService[F] =
+  type GmosNConfigs = (GmosNorthGrating, GmosNorthFpu, Wavelength, Option[GmosXBinning], Option[GmosYBinning])
+  type GmosSConfigs = (GmosSouthGrating, GmosSouthFpu, Wavelength, Option[GmosXBinning], Option[GmosYBinning])
```
swalker2m, line 56: It feels like the binning may be problematic.  It's never really `None`, but rather defaults when not explicitly defined. `None` can actually mean any of the binning options.

cquiroz: From my tests i see this always defined, I think when you set the mode some binning will be set

However on the table is nullable thus the type needs to be `Option[A]`  I think

swalker2m: If the PI doesn't go out of their way to set the binning, it will be calculated from a variety of parameters in code.  It's a pretty convoluted calculation.  Maybe we can figure out how to add the result to a view, I'm not sure.

If you do a query to get all the unique configurations, it seems like we need the actual binning that will be used.

modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
```diff
@@ -74,17 +94,134 @@ object CalibrationsService {
...
+          o2          <- gmosSouthLSObservations(gsls.diff(currGmosSLS))
+        } yield (o1 ::: o2)).map(_.sequence)
+      }
+
+      // Set tthe calibration role of the observations in bulk
+      private def setCalibRoleAndGroup(oids: List[Observation.Id], calibrationRole: CalibrationRole)(using Transaction[F]): F[Unit] = {
+        val update = void"UPDATE t_observation " |+|
+          sql"SET c_calibration_role = $calibration_role "(calibrationRole) |+|
+          void"WHERE c_observation_id IN (" |+|
+            oids.map(sql"$observation_id").intercalate(void", ") |+| void")"
```
swalker2m, line 192: We put the SQL down in `Statements` typically.

modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
```diff
@@ -74,17 +94,134 @@ object CalibrationsService {
...
+                          mode = ScienceMode.Spectroscopy.some,
+                          spectroscopy = SpectroscopyScienceRequirementsInput.Default.copy(
+                            signalToNoise = Nullable.NonNull(SignalToNoise.unsafeFromBigDecimalExact(100.0))
+                        ).some
+                      ).some
+                    ).some
+            )
+          )
+        }
+        def gmosSouthLSObservations(gsls: List[GmosSConfigs]) = gsls.traverse { case (g, f, w, xb, yb) =>
```
swalker2m, line 138: Seems like most of this is in common with the `North` version.

modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
```diff
@@ -74,17 +94,134 @@ object CalibrationsService {
...
+      }
+
       def recalculateCalibrations(pid: Program.Id)(using Transaction[F]): F[Unit] =
         for {
-          gnls <- session.execute(Statements.SelectGmosNorthLongSlitConfigurations)(pid)
-          gsls <- session.execute(Statements.SelectGmosSouthLongSlitConfigurations)(pid)
-          _    <-  calibrationsGroup(pid, gnls.size + gsls.size)
+          gnls <- session.execute(Statements.selectGmosNorthLongSlitConfigurations(false))(pid)
+          gsls <- session.execute(Statements.selectGmosSouthLongSlitConfigurations(false))(pid)
+          _    <- generateCalibrations(pid, gnls, gsls).whenA(gnls.nonEmpty || gsls.nonEmpty)
```
swalker2m, line 214: What happens when there are already calibrations in the group?  I didn't see the code to remove existing ones.

swalker2m (approved):
Apart from the issue with binning which we'll have to figure out, this seems fine.  If you want to merge so you don't have to rebase a lot :+1:

modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
cquiroz: yes it is very similar. I'm not sure if it  can be parametrized as they are disjoint types

modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
cquiroz: at the moment it only creates if there is no exisiting one. I have a task also to prune them if needed

modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
cquiroz: As we discussed we'll tackle this next on a separate PR

PR #1278, opened 2024-07-04
Inclusive Active Period End

swalker2m:
It was requested in [Short Cut 2951](https://app.shortcut.com/lucuma/story/2951/calculate-ra-limit-according-to-cfp-active-period#activity-3095) to make the end date of the active period inclusive rather than exclusive. This cuts against the grain of intervals, at least in my mind and certainly relative to the `TimestampInterval`.  Exclusive end is more convenient in most contexts because then the end of one interval is equal to the start of the next abutting interval.  Not to mention that end minus start yields the length of the interval.  I realized I could handle this just by adding a day on input and subtracting a day on output.

The drawback is that what we store and what we model internally would still be an exclusive end.  I'm guessing this discrepancy will bite us or our descendants at some point, but I'm reluctant to make an interval with an inclusive end for dates while leaving timestamps with an exclusive end.  Converting `TimestampInterval` to inclusive end would be challenging since it is used extensively in time accounting.

What do we think?

### DIFF
--- a/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -390,7 +390,7 @@ type CallForProposals {
 
   """
   The active period during which accepted observations for this call may be
-  observed.
+  observed (inclusive start and end dates).
   """
   active: DateInterval!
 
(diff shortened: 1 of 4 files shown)


swalker2m:
Andy suggested that inclusive end dates might be sacrificable (to coin a new word).  I'm inclined to not do this for the reasons described above.

PR #1279, opened 2024-07-04
Add observerNotes field

cquiroz:


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -3104,6 +3104,10 @@ input ObservationPropertiesInput {
   """
   groupIndex: NonNegShort
 
+  """
+  Set the notes for  thhe observer
+  """
+  observerNotes: String
```
cquiroz, line 3110: I'm not sure if this should be `NonEmptyString`. Any opinions?

modules/service/src/main/scala/lucuma/odb/service/ObservationService.scala
```diff
@@ -901,6 +909,11 @@ object ObservationService {
           },
           SET.status.map(upStatus),
           SET.activeStatus.map(upActive),
+          SET.observerNotes match {
+            case Nullable.Null  => Some(upObserverNotes(None))
+            case Absent         => None
+            case NonNull(value) => Some(upObserverNotes(Some(value)))
+          },
```
swalker2m, line 916: A convenient way to do this is with `foldPresent` on `Nullable`:

```
SET.observerNotes.foldPresent(upObserverNotes)
```

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
swalker2m: I don't think it matters that much, but we made the subtitle `NonEmptyString` so I guess we could be more consistent with that.

swalker2m (approved):
This looks great.  :+1:

PR #1284, opened 2024-07-08
Banded Observations

swalker2m:
Adds a `scienceBand` to `Observation`, including the `WhereObservations` filter.  There will be a number of rules about which bands are legal to assign, when the band can and cannot be empty, what happens when the allocation is changed, etc.  I don't know what those rules will be so it is open-ended at the moment.


tpolecat (approved):
Calm before the storm. Looks good for now. :+1:

PR #1286, opened 2024-07-09
Add selector to pick targets based on calibration role

cquiroz:


modules/service/src/main/scala/lucuma/odb/graphql/input/WhereOptionEq.scala
```diff
@@ -25,8 +26,8 @@ object WhereOptionEq {
           (isNull, EQ, NEQ, IN, NIN) =>
             and(List(
               isNull.map(IsNull(path, _)),
-              EQ.map(a => Eql(path, Const(a))),
-              NEQ.map(a => NEql(path, Const(a))),
+              EQ.map(a => Eql(path, Const(a.some))),
+              NEQ.map(a => NEql(path, Const(a.some))),
               IN.map(as => In(path, as)),
```
swalker2m, line 31: Should that be `as.map(_.some)` for `IN` and `NIN`?  Maybe add a test or change the `EQ` test to `IN`?

modules/service/src/test/scala/lucuma/odb/graphql/DatabaseOperations.scala
```diff
@@ -1264,6 +1266,24 @@ trait DatabaseOperations { this: OdbSuite =>
...
+        c_library_desc
+      )
+      select
+        $calibration_role,
+        $text
+      returning c_program_id
+    """.apply(calibrationRole, description.value)
+    FMain.databasePoolResource[IO](databaseConfig).flatten
+      .use(_.prepareR(af.fragment.query(program_id)).use(_.unique(af.argument)))
+  }
```
swalker2m, line 1285: There isn't a service method that would do this?

swalker2m (approved):
Looks good, with a couple of observations.

modules/service/src/main/scala/lucuma/odb/graphql/input/WhereOptionEq.scala
cquiroz: I'll add a test for `IN`

modules/service/src/test/scala/lucuma/odb/graphql/DatabaseOperations.scala
cquiroz: Not really as it is not expected to create system programs programmatically. For example, it isn't allow in the API.

swalker2m: Wouldn't the programmatic update use the service?  If we can, I think we want to funnel all these database manipulations into service code.

swalker2m (approved):
This looks great to me.  Thanks for fixing the bug I left in `IN` and `NIN`! :+1:

PR #1288, opened 2024-07-10
`setAllocations`

swalker2m:
Replaces `setAllocation` with `setAllocations` (plural). There is a request to handle programs with a time allocation in a single band distinctly from programs with allocations in multiple bands.  In order to do this, we need to know the full set of allocations at once instead of adding them individually.

Inexplicably, I wasn't able to get the result of calling `setAllocations` (`SetAllocationsResult`) to have sorted `allocations`.  I've left the failed attempt commented out for now in case there's an opportunity to investigate further.


modules/service/src/main/scala/lucuma/odb/graphql/mapping/MutationMapping.scala
```diff
@@ -252,15 +251,6 @@ trait MutationMapping[F[_]] extends Predicates[F] {
...
     )
 
-  def proposalAttachmentResultSubquery(pid: Program.Id, aTypes: List[Tag], limit: Option[NonNegInt], child: Query) =
-    mutationResultSubquery(
-      predicate = And(Predicates.proposalAttachment.program.id.eql(pid), Predicates.proposalAttachment.attachmentType.in(aTypes)),
-      order = OrderSelection[Tag](ProposalAttachmentType / "attachmentType"),
-      limit = limit,
-      collectionField = "proposalAttachments",
-      child
-    )
```
swalker2m, line 262: This was unused and I suspected it was meant to be deleted some time ago, so I removed it.

modules/service/src/test/scala/lucuma/odb/graphql/DatabaseOperations.scala
```diff
@@ -552,57 +553,68 @@ trait DatabaseOperations { this: OdbSuite =>
...
+    json.hcursor.downFields(topLevelField, "allocations").values.toList.flatten.traverse { obj =>
+      val c = obj.hcursor
+      (for {
+        p <- c.downField("partner").as[Partner]
+        b <- c.downField("scienceBand").as[ScienceBand]
+        d <- c.downFields("duration", "hours").as[BigDecimal]
+      } yield (p, b, d))
+       .leftMap(f => new RuntimeException(f.getMessage))
+       .liftTo[IO]
+    }.map(_.toSet)
```
swalker2m, line 569: Since I cannot control the order of allocations coming out of the mutation, tests have to compare the set of allocations returned.

swalker2m:
> For the record can you say a little bit more about this?

When there is only one band in the time allocations, every observation will be automatically assigned that band.  When there are multiple bands, the PI will have choose (at least for new observations not given a band by the PES).

PR #1291, opened 2024-07-11
Add method to clone a target into a project

cquiroz:
This will be needed to copy calibration targets into programs

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/TargetService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/TargetService.scala
@@ -56,6 +56,7 @@ trait TargetService[F[_]] {
   def createTarget(input: CreateTargetInput)(using Transaction[F]): F[Result[Target.Id]]
   def updateTargets(input: TargetPropertiesInput.Edit, which: AppliedFragment)(using Transaction[F]): F[Result[List[Target.Id]]]
   def cloneTarget(input: CloneTargetInput)(using Transaction[F]): F[Result[(Target.Id, Target.Id)]]
+  def cloneTargetInto(targetId: Target.Id, programId: Program.Id)(using Transaction[F]): F[Result[(Target.Id, Target.Id)]]
 }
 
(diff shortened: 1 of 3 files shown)


modules/service/src/main/scala/lucuma/odb/service/TargetService.scala
```diff
@@ -201,14 +203,42 @@ object TargetService {
...
+        // Ensure the destination program exists
+        val pid: F[Option[Program.Id]] =
+          session.prepareR(sql"select c_program_id from t_program where c_program_id = $program_id".query(program_id)).use { ps =>
+            ps.option(programId)
+          }
+
+        pid.flatMap {
+          case None => NoSuchProgram(programId).pure[F]
+          case Some(pid) =>
+            clone(targetId, pid).flatMap {
```
swalker2m, line 220: It took me a beat to figure out what was going on here.  It looks like we first clone the target into its existing program (which would be the calibration program).  The `pid` being passed is not that calibration program though but instead the ultimate destination of the target.  The `pid` is used just for user access checks.  Is this correct?

I wonder if we could alter `clone` itself in order to do the clone directly into the target program.

```scala
    sql"""
        INSERT INTO t_target(
          c_program_id,
          c_name,
          c_type,
...
        )
        SELECT
          $program_id,
          c_name,
          c_type,
...
        FROM t_target
        WHERE c_target_id = $target_id
      """.apply(pid, tid) |+|
      ProgramService.Statements.existsUserAccess(user, pid).foldMap(void"AND " |+| _) |+|
      void"""
        RETURNING c_target_id
      """
```
Would that work?  Then we wouldn't need to do a separate statement and the `clone` method would be more direct.

cquiroz: I thought about doing that but didn't pursue it. I agree it would be better

swalker2m (approved):
Looks good to me. :+1:

PR #1292, opened 2024-07-11
Assign observation band on single band allocation

swalker2m:
This PR has two disparate updates

1. Sets the band of every observation in a program to the band of the time allocation (when there is only a single band)
2. Fixes a mapping issue when removing time allocations along with the allocation sorting issue in `setAllocations` results


modules/service/src/main/scala/lucuma/odb/service/AllocationService.scala
```diff
@@ -31,11 +30,14 @@ object AllocationService {
...
-      def setAllocations(input: SetAllocationsInput)(using Transaction[F], Services.StaffAccess): F[Result[Unit]] =
+      def setAllocations(input: SetAllocationsInput)(using Transaction[F], Services.StaffAccess): F[Unit] =
         deleteAllocations(input.programId) *>
         NonEmptyList.fromList(input.allocations).traverse_ { lst =>
           session.execute(Statements.setAllocations(lst))((input.programId, lst))
-        }.as(Result.unit)
+        }                                  *>
+        observationService
+          .setScienceBand(input.programId, input.bands.head)
+          .whenA(input.bands.sizeIs == 1)
```
tpolecat, line 40: I haven't seen `sizeIs` before ... ?

tpolecat (approved):
👍🏻 looks good

PR #1293, opened 2024-07-11
obs status part 1 - itc validation

tpolecat:
This adds a validation that ensures ITC results are present. The idea here is that an observation isn't fully defined until we have valid ITC results in the cache. This is kind of a weird case because Explore forces computation of ITC results eagerly, so this error may never be visible. But we need to handle this case (and need to cache ITC failures, which is the more important one, to follow).


modules/service/src/main/scala/lucuma/odb/service/ItcService.scala
```diff
@@ -188,8 +188,8 @@ object ItcService {
   ) extends TargetResult
 
   case class AsterismResult private (
-    acquisitionResult: Zipper[TargetResult],
-    scienceResult: Zipper[TargetResult]
+    acquisitionResult: Zipper[TargetImagingResult],
+    scienceResult: Zipper[TargetSpectroscopyResult]
```
tpolecat, line 192: The more precise type here is required in order to pull out the "input" values below.

modules/service/src/main/scala/lucuma/odb/service/ItcService.scala
```diff
@@ -422,8 +422,8 @@ object ItcService {
         resultSet: AsterismResult
       )(using Transaction[F]): F[Unit] = {
 
-        def insertOrUpdateSingleTarget(acquisition: TargetResult)(science: TargetResult): F[Unit] = {
-          val h = hash((acquisition, science))
+        def insertOrUpdateSingleTarget(acquisition: TargetImagingResult)(science: TargetSpectroscopyResult): F[Unit] = {
+          val h = hash(acquisition.input, science.input)
```
tpolecat, line 426: Here is the caching issue (it was hashing the entire result rather than just the inputs). Took a while to find but easy to fix.

modules/service/src/main/scala/lucuma/odb/service/ObservationService.scala
```diff
@@ -560,11 +565,17 @@ object ObservationService {
...
+            case Some(_) => ObservationValidationMap.empty
+            case None => ObservationValidationMap.empty.add(ObservationValidation.itc("ITC results are not present."))
 
         for {
-          genVals <- generatorValidations
-          cfpVals <- cfpValidations
-        } yield (genVals |+| cfpVals).toList
+          (genVals, op) <- generatorValidations
+          cfpVals       <- cfpValidations
+          itcVals       <- op.filter(_ => cfpVals.isEmpty).foldMapM(itcValidations) // only compute this if cfp and gen are ok
```
tpolecat, line 577: I'm only computing the ITC validation if everything else is ok; there's no point complaining about ITC if the asterism or config are invalid.

modules/service/src/main/scala/lucuma/odb/service/ItcService.scala
swalker2m: Thanks for handling this.

modules/service/src/main/scala/lucuma/odb/service/ObservationService.scala
```diff
@@ -560,11 +565,17 @@ object ObservationService {
           )
         }
 
+        def itcValidations(params: GeneratorParams): F[ObservationValidationMap] =
+          itcService(itcClient).selectOne(pid, oid, params).map:
+            // N.B. there will soon be more cases here
+            case Some(_) => ObservationValidationMap.empty
+            case None => ObservationValidationMap.empty.add(ObservationValidation.itc("ITC results are not present."))
```
swalker2m, line 572: I think we discussed this, but just in case there may be no results yet simply because we haven't requested them.

swalker2m (approved):
Looks great to me. :+1:

PR #1300, opened 2024-07-15
Edit proposals with deleted cfp

swalker2m:
Proposals may reference existing Calls for Proposals.  They draw the semester from the CFP and the type must agree as well.  This PR introduces an update trigger on `t_cfp` such that when critical properties are modified, if there are any existing referencing proposals we reject the change with an exception.


modules/service/src/main/resources/db/migration/V0892__stop_invalid_cfp_updates.sql
```diff
@@ -0,0 +1,41 @@
...
+      (NEW.c_semester  <> OLD.c_semester)  OR
+      (NEW.c_type      <> OLD.c_type))     THEN
+
+    -- Select program ids that use this CFP, if any.
+    SELECT json_agg(c_program_id) INTO program_ids
+    FROM t_proposal
+    WHERE c_cfp_id = NEW.c_cfp_id;
+
+    IF program_ids IS NOT NULL THEN
+      RAISE EXCEPTION 'Cannot delete this Call for Proposals, or change its type or semester, because dependent proposals reference it: %', program_ids::text
```
cquiroz, line 27: 👍

cquiroz (approved):
LGTM

PR #1306, opened 2024-07-17
Science Band Validations

swalker2m:
Performs a few validations related to setting observation science bands:

* Disallows creating an observation with a science band for which time has not been allocated
* Disallows updating an observation and setting a science band for which time has not been allocated
* Permits updating allocations, even if it would leave observations with an invalid science band, but generates a warning

I'll add a validation check for observations assigned a science band without a time allocation in the next (?) PR.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/graphql/mapping/MutationMapping.scala
+++ b/modules/service/src/main/scala/lucuma/odb/graphql/mapping/MutationMapping.scala
@@ -527,7 +527,7 @@ trait MutationMapping[F[_]] extends Predicates[F] {
     MutationField("setAllocations", SetAllocationsInput.Binding): (input, child) =>
       services.useTransactionally:
         requireStaffAccess:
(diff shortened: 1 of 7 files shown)


swalker2m:
> Permits updating allocations, even if it would leave observations with an invalid science band, but generates a warning

I'm particularly interested in feedback on this one.  It kind of cuts against the grain.  I checked with Andy and he felt that we shouldn't prevent an update by the staff to the allocations, even if it leaves some observations with an invalid band. It does make it easier to adjust the allocations, I suppose.

I have a new observation validation that checks for this situation (observation assigned invalid band) ready to go if we merge this.

tpolecat (approved):
This all looks fine, but I do wonder if this kind of thing might work better in some cases as validation errors rather than a database errors. I think they sometimes say "we need to prevent X from happening" when "tell the user they need to fix X" would also be ok.

In any case :+1:

PR #1307, opened 2024-07-18
Science Band Observation Validation

swalker2m:
Adds a simple observation validation that checks whether the science band is valid according to the time allocations.


toddburnside (approved):
LGTM

PR #1309, opened 2024-07-18
Create new observations with a band when appropriate

swalker2m:
When a new observation is created, if the program has time allocated in a single band, the desired behavior is that the observation be automatically assigned that band.  If time is allocated in multiple bands, new observations are not automatically assigned any band.


cquiroz:
Is there no concept of default band?

swalker2m:
> Is there no concept of default band?

No, in order for an observation to be considered ready its program has to be allocated time in one or more particular bands and the observation must be assigned one of those bands.

cquiroz (approved):
LGTM

PR #1311, opened 2024-07-19
Don't allow cloning of calibration observations

cquiroz:


swalker2m (approved):
Seems good.  We can't do it yet, but if we could maybe the right thing would be to allow a clone of a calibration observation but it would become a normal observation with a manual sequence.  At any rate, :+1:

PR #1313, opened 2024-07-19
Calibration conditions

cquiroz:
Calibrations shouldn't use the default conditions


swalker2m (approved):
Seems reasonable.  :+1:

PR #1321, opened 2024-07-24
Adds GMOS Long Slit Spectro Photometric Calibration Generation

swalker2m:
Refactors the sequence generation code to accommodate calibration observations.  Adds GMOS Long Slit Spectro Photometric sequence generation.


tpolecat (approved):
This looks good to me. At some point it might be worthwhile writing a README for sequence generation as an overview of how it works (maybe after we add more modes since the generalizations may change). In any case :+1:

PR #1328, opened 2024-07-30
Initial group contents

tpolecat:
This adds a parameter to `CreateGroupInput` allowing the user to specify the initial contents of the group. Elements specified will be moved into the new group in the order specified.

### DIFF
--- a/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -7306,6 +7306,12 @@ type GroupElement {
   existence: Existence!
 }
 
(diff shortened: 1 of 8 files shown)


swalker2m (approved):
Looks perfectly cromulent. :+1:

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -7347,6 +7353,9 @@ input CreateGroupInput {
   proposalReference: ProposalReferenceLabel
   programReference: ProgramReferenceLabel
   SET: GroupPropertiesInput
+  
+  "Group elements specified here, if any, will be moved into the created group in the specified order."
+  initialContents: [GroupElementInput]
```
hugo-vrijswijk, line 7358: This looks great 👍

PR #1331, opened 2024-08-02
ShortCut 3017: Phase 0 instrument configuration option update

swalker2m:
Updates the instrument configuration options, using a repeatable migration.  These are requested by explore and those that match the science requirements are presented in the UI for selection by the PI.

At the moment, the options are maintained by the science staff in a Google spreadsheet.  This can now be exported to a `.tsv` file and replaced in the codebase.  A redeployment will cause the repeatable migration to run, parse the `.tsv` file and update the tables.  This is how SmartGcal works for now as well.

We could automate this in the future to obviate the need for software involvement, but updates are anticipated only once per semester.  Ultimately, these could be computed directly from a query based on availability.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,7 @@ jobs:
         run: sbt -v -J-Xmx6g +update
 
       - name: Docker compose up
-        run: docker-compose up -d
+        run: docker compose up -d
 
       - name: Check that workflows are up to date
         run: sbt -v -J-Xmx6g githubWorkflowCheck
@@ -93,7 +93,7 @@ jobs:
         uses: codecov/codecov-action@v4
 
       - name: Docker compose down
-        run: docker-compose down
+        run: docker compose down
 
   publish:
     name: Publish Artifacts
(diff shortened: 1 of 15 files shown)


swalker2m:
CI fails with

```
Run docker-compose up -d
  docker-compose up -d
  shell: /usr/bin/bash -e {0}
  env:
    GITHUB_TOKEN: ***
    JAVA_HOME: /opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/17.0.12-7/x64
    JAVA_HOME_17_X64: /opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/17.0.12-7/x64
/home/runner/work/_temp/fbf58df0-c8fe-4d2d-95b9-e00dfd9876f1.sh: line 1: docker-compose: command not found
Error: Process completed with exit code 127.
```

I'm not sure what to make of that.

swalker2m:
> CI fails with
...
> I'm not sure what to make of that.

I updated `sbt-lucuma` to use `docker compose` without the hyphen and that seems to work.  Thanks for the suggestion @hugo-vrijswijk!

toddburnside (approved):
LGTM

PR #1333, opened 2024-08-05
CFP `allowsNonPartnerPi`

swalker2m:
From [ShortCut 3279](https://app.shortcut.com/lucuma/story/3279/allow-specifying-whether-a-call-for-proposals-allows-non-partners-to-be-pi), this PR adds an `allowsNonPartnerPi` flag to the `CallForProposals` type, along with a `nonPartnerDeadline` that may be used in that case.

Does not actually change anything about invitations and partner/pi or partner/coi links.


modules/service/src/main/resources/db/migration/V0899__cfp_non_partner.sql
```diff
@@ -0,0 +1,46 @@
...
+      GROUP BY
+        c.c_cfp_id
+    ),
+    cte_partner AS (
+      SELECT
+        c.c_cfp_id,
+        (COUNT(*) FILTER (WHERE p.c_deadline IS NOT NULL AND p.c_deadline > CURRENT_TIMESTAMP) > 0) AS c_is_open,
+        MAX(CASE WHEN p.c_partner = 'us' AND (c.c_type = 'regular_semester' OR c.c_type = 'directors_time') THEN 1 ELSE 0 END) AS c_non_partner,
+        MAX(CASE WHEN p.c_partner = 'us' THEN p.c_deadline ELSE NULL END) AS c_us_deadline
+      FROM
```
swalker2m, line 27: The new fields are computed in the `v_cfp` view here.  Only proposals that US PIs may apply to may be used by PIs without a partner.  By default, when a CFP is created if partners are not listed then all partners are assumed.  The US deadline (which may be the default deadline if not explicitly specified) is used as the non-partner deadline.

modules/service/src/main/resources/db/migration/V0899__cfp_non_partner.sql
```diff
@@ -0,0 +1,46 @@
...
+      GROUP BY
+        c.c_cfp_id
+    )
+    SELECT
+      c.*,
+      cte_instrument.c_instruments,
+      cte_partner.c_is_open,
+      format_cfp_title(c.c_type, c.c_semester, c.c_active_start, c.c_active_end, cte_instrument.c_instrument_names) AS c_title,
+      cte_partner.c_non_partner = 1 AS c_allows_non_partner,
+      (CASE WHEN cte_partner.c_non_partner = 1 THEN cte_partner.c_us_deadline ELSE NULL END) AS c_non_partner_deadline
```
swalker2m, line 40: Here we refer to the common table expression above with the new fields.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -414,6 +414,16 @@ type CallForProposals {
   """
   partners: [CallForProposalsPartner!]!
 
+  """
+  Whether this Call allows PIs without a partner to participate.
+  """
+  allowsNonPartnerPi: Boolean!
```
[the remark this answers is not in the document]
swalker2m: We'll need further changes to actually support non-partner PIs and COIs.  All this PR does is add some computed fields on the `CallForProposals` to make it easier to determine whether non-partners are allowed to apply and if so, what would their deadline be.

PR #1334, opened 2024-08-08
NonPartner (and unspecified partner) Partner Options

swalker2m:
This is a follow up for #1318, incorporating the discussion from the GPP Summit in La Serena (see [ShortCut 3165](https://app.shortcut.com/lucuma/story/3165/add-non-partner-partner)).  In particular, it introduces a `PartnerLink` interface which covers three cases:

* `HasPartner(Partner)` - a normal partner
* `HasNonPartner` - explicitly set the partner to non-partner
* `HasUnspecifiedPartner`- we haven't selected a partner, or there cannot be a partner for this role

For example:

```
partnerLink {
  linkType
  ... on HasPartner {
    partner
  }
}
```

I think this will help enable [ShortCut 219](https://app.shortcut.com/lucuma/story/219/add-ui-for-viewing-adding-and-editing-investigators) and [ShortCut 3359](https://app.shortcut.com/lucuma/story/3359/associate-pi-with-partner-or-nonpartner).

At the moment, there is no dedicated API call to change the partner for an existing user but I believe it can be accomplished by an `unlinkUser` followed by `linkUser`?


tpolecat (approved):
This looks great. 👍🏻

swalker2m:
As a heads up on this, I fear something might have to change when we associate a `PartnerLink` with the PI.  I think the `PartnerLink` cannot be moved to `User` because that would involve SSO and cookies, etc.?  So the next best thing might be to split the role from `ProgramUser` so that the PI could be a `ProgramUser` as well (as is, this isn't possible because `ProgramUserRole` doesn't include PI).  So maybe something like this?

```
type ProgramUserInRole {
  role: ProgramUserRole!
  user: ProgramUser!
}

type ProgramUser {
  partnerLink: PartnerLink!
  userId:      UserId!
  user:        User
}

type Program {
  pi: ProgramUser
  users: [ProgramUserInRole!]!
}
```

tpolecat:
I wonder if maybe it would make sense to move the PI to the ProgramUser table. I don't think there's any particular reason it needs to be on the program itself … the design kind of evolved without a lot of thought.

swalker2m:
> I wonder if maybe it would make sense to move the PI to the ProgramUser table.

That seems like a good idea.  I think we would need to add a PI role to the `ProgramUserRole` enum, and we'd need to validate that the users list has exactly one PI entry.

PR #1336, opened 2024-08-13
clone group

tpolecat:
This adds an operation to clone a [non-system] group.


modules/service/src/main/scala/lucuma/odb/service/GroupService.scala
```diff
@@ -249,6 +320,79 @@ object GroupService {
...
+          .contramap[Group.Id](a => (a, a))
+          .query(
+            (group_id.opt ~ observation_id.opt).emap:
+              case (None, Some(oid)) => Right(Right(oid))
+              case (Some(gid), None) => Right(Left(gid))
+              case (a, b)            => Left("SelectGroupElements: unpossible row: $a, $b")
+          )
+
+      /** Select a `CreateGroupInput` for a given `Group.Id` that can be used to create an empty clone. */
+      val SelectGroupAsInput: Query[Group.Id, (CreateGroupInput, Boolean)] =
```
cquiroz, line 345: Should this filter out system groups?

tpolecat: It selects the system bit and delivers an error if it's true. I could have filtered it out but that would also mask the case where the group id is invalid. Not sure if we need an explicit error in that case.

modules/service/src/main/scala/lucuma/odb/service/GroupService.scala
```diff
@@ -62,6 +71,68 @@ object GroupService {
...
+        ResultT:
+          session.prepareR(Statements.SelectGroupElements).use: pq =>
+            pq.stream(gid, 1024).compile.toList.map(Result.success)
+
+      // Clone `gid` as a sibling.
+      private def cloneGroupImpl(input: CloneGroupInput)(using Transaction[F]): ResultT[F, Group.Id] =
+        for
+          cgi   <- selectGroupAsInput(input.groupId)
+          cgiʹ   = cgi.copy(SET = input.SET.foldLeft(cgi.SET)(_.withEdit(_)))
+          clone <- ResultT(createGroup(cgiʹ, false))
```
swalker2m, line 128: Is the sibling added at the end or next door?

swalker2m (approved):
Looks great :+1:

modules/service/src/main/scala/lucuma/odb/service/GroupService.scala
tpolecat: The clone is inserted at the specified index, or at the source index if left unspecified (in which case the source will be pushed down to make room).

PR #1337, opened 2024-08-13
Make the PI a `ProgramUser`

swalker2m:
This PR is an initial step toward [making it possible to associate the PI with a partner](https://app.shortcut.com/lucuma/story/3359/associate-pi-with-partner-or-nonpartner).  It changes the way PIs are associated with programs to make them more consistent with CoIs and support users.  In particular, the explicit pi user reference on `t_program` is removed and instead we add the PI as a normal `ProgramUser`.  `ProgramUser`s in turn have a partner link, so we now have a place to tie PIs with a partner (or non-partner).

Currently there is no mechanism for setting the partner.  I think the next PR would add a new mutation to update program users, particularly the partner link.  This would be made to work for PIs and CoIs alike.


swalker2m:
I think this works fine, but I have two chief concerns:

1. Will it negatively impact authorization checks, since so much depends on users.
2. Is it kosher to make the program `UPDATED` event immediately followed the `CREATED` event (because the PI is linked)?

tpolecat:
> Will it negatively impact authorization checks, since so much depends on users.

We already have to check the linked users basically all the time, so it may actually simplify things.

> Is it kosher to make the program UPDATED event immediately followed the CREATED event (because the PI is linked)?

I think that's fine, although it might be worth asking the client chaps if it's going to cause problems on their end.

tpolecat (approved):
This was a fiddly refactor and it looks really good. Thanks for taking this on. :+1:

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -8466,7 +8467,7 @@ type Program {
   proposalStatus: ProposalStatus!
 
   "Principal Investigator"
-  pi: User
+  pi: ProgramUser
```
cquiroz, line 8470: Do we need it here? Wouldn't it be enough to have the `pi` in  `users`?

swalker2m: That's a good point and I wondered about doing that.  The PI is kind of "special" and it seemed like it might be handy to have a direct reference to it instead of having to pick it out of the `users` list.  That said, I thought I'd end up needing it in a predicate but it turns out I didn't.

cquiroz (approved):
LGTM

PR #1338, opened 2024-08-14
Update observation time

cquiroz:
Adds a new mutation to update the observation time, this could be used by the scheduler to set the expected execution time and eventually the duration

### DIFF
--- a/modules/service/src/main/resources/db/migration/V0904__observation_time.sql
+++ b/modules/service/src/main/resources/db/migration/V0904__observation_time.sql
@@ -1,0 +1,17 @@
+-- Rename visualization time
+ALTER TABLE t_observation
+RENAME COLUMN c_visualization_time TO c_observation_time;
+
+-- Re-create v_observation to include the science band
+DROP VIEW v_observation;
+CREATE OR REPLACE VIEW v_observation AS
+  SELECT o.*,
+  CASE WHEN c_explicit_ra              IS NOT NULL THEN c_observation_id END AS c_explicit_base_id,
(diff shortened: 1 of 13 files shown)


modules/service/src/test/scala/lucuma/odb/graphql/mutation/updateObservationsTimes.scala
```diff
@@ -0,0 +1,136 @@
...
+        }
+      }
+    """.asRight
+
+    val expected1 = json"""
+      {
+        "updateObservationsTimes": {
+          "observations": [
+            {
+              "observationTime": null
```
swalker2m, line 124: It occurred to me that setting the observation time to the same value for multiple observations might be something that is never really done, but it does seem handy to be able to wipe them all out in one mutation like this.

cquiroz: That's right. it tests the machinery of multiple observations though

modules/service/src/main/resources/db/migration/V0904__observation_time.sql
```diff
@@ -0,0 +1,17 @@
+-- Rename visualization time
+ALTER TABLE t_observation
+RENAME COLUMN c_visualization_time TO c_observation_time;
+
+-- Re-create v_observation to include the science band
```
toddburnside, line 5: I don't see the science band in the view...

toddburnside (approved):
LGTM

PR #1339, opened 2024-08-14
Add WhereProgramUser

swalker2m:
Adds a `WhereProgramUser` and uses it to match program PIs.  This is part of a new mutation under development that will allow a user's partner to be updated.  It seems useful in its own right though and marks a reasonable size chunk for review. It is pretty much mechanical, but I haven't tested all the available options. In particular, there's no way to set the PI's partner yet so matching on that cannot be tested.

### DIFF
--- a/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -11529,6 +11529,60 @@ input WhereEqPartner {
 }
 
 """
+Filters on equality (or not) of the partner link type. All supplied criteria
+must match, but usually only one is selected.
+"""
+input WhereEqPartnerLinkType {
+  """
+  Matches if the partner link type is exactly the supplied value.
+  """
+  EQ: PartnerLinkType
+
+  """
+  Matches if the partner link type is not the supplied value.
+  """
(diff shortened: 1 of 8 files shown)


cquiroz (approved):
LGTM

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -13233,6 +13411,11 @@ input WhereProgram {
   """
   reference: WhereProgramReference
 
+  """
+  Matches the PI.
+  """
+  pi: WhereProgramUser
```
toddburnside, line 13417: WhereProgramUser allows specifying a role. I guess it's up to the user to not specify that for the PI, or to make sure it is `PI`? I think that is reasonable.

toddburnside (approved):
LGTM

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
swalker2m: Right, for this case matching on the program pi role and setting it to coi would be a complicated way to get an empty program list result.  The idea though is to use this same `WhereProgramUser` in the partner update mutation though.

PR #1340, opened 2024-08-14
`programUsers` query

swalker2m:
Adds a `programUsers` query, which will be used (I hope) to select program users for updating the partner link.


tpolecat (approved):
Looks great 👍🏻

PR #1341, opened 2024-08-19
Update a user's partner

swalker2m:
Adds a mutation that allows updating a `ProgramUser`'s partner.  It follows the fairly worn path taken by all the other update mutations, but there are a couple of points worthy of discussion:

* The `ProgramUserPropertiesInput` has just one property at the moment, the `partnerLink`. It seems like this could grow other fields in the future (e.g., changing the user's role?) but for now it is an extra level in the input given that `partnerLink` is the only option.
```
input: {
  SET: {
    partnerLink: { partner: US }
  }
}
```

* I believe that any user who can edit the program can also set other users' partner links. I think this can be accomplished by the standard `ProgramPredicates.isWritableBy(user)` filter to the `WHERE` clause id resolution in `MutationMapping` as we do elsewhere.  It seemed wrong though for `ProgramService` to accept an `AppliedFragment` of `ProgramUser` ids and just allow updates to be done to them without checking so I added a `correlated` variant of the `existsUserAccess` subquery that references the outer query program id.  Maybe this is not necessary?  Now it does check twice I suppose.


modules/service/src/main/scala/lucuma/odb/graphql/mapping/MutationMapping.scala
```diff
@@ -739,6 +760,33 @@ trait MutationMapping[F[_]] extends Predicates[F] {
       }
     }
 
+  private lazy val UpdateProgramUsers =
+    MutationField("updateProgramUsers", UpdateProgramUsersInput.binding(Path.from(ProgramUserType))) { (input, child) =>
+      services.useTransactionally {
+        val filterPredicate = and(List(
+          Predicates.programUser.program.isWritableBy(user),
```
swalker2m, line 767: Here we check the first time that the user has write access, when we figure out which program users we will attempt to update.  The `isWritableBy` predicate needs to be updated to work for read-only CoIs and for NGO users.

cquiroz:
I'm sure we want to edit more fields like student status, etc

modules/service/src/main/scala/lucuma/odb/graphql/mapping/MutationMapping.scala
tpolecat: Yeah all of these things are on my very long list of things to review once the obs status stuff is working.

tpolecat (approved):
This all makes sense. I will re-evaluate the access control stuff eventually. 👍🏻

PR #1351, opened 2024-08-20
Add calibration time change and adjust the target when it changes

cquiroz:
On this PR we add a new stream of events whenever a calibration observation time changes.
The calibration daemon listens to those changes and will update the target depending on the date

There is one missing part which is avoiding the creation of unneeded targets and deleting stale one but will be addressed in another PR

### DIFF
--- a/modules/calibrations/src/main/scala/lucuma/odb/calibrations/Main.scala
+++ b/modules/calibrations/src/main/scala/lucuma/odb/calibrations/Main.scala
@@ -153,11 +153,8 @@ object CMain extends MainParams {
             }.compile.drain.start.void)
       _  <- Resource.eval(Logger[F].info("Start listening for calibration time changes"))
       _  <- Resource.eval(calibTopic.subscribe(100).evalMap { elem =>
-              services.useTransactionally{
-                for {
-                  t <- Sync[F].delay(LocalDate.now(ZoneOffset.UTC))
-                  _ <- calibrationsService.recalculateCalibrationTarget(elem.programId, elem.observationId)
-                } yield ()
+              services.useTransactionally {
+                calibrationsService.recalculateCalibrationTarget(elem.programId, elem.observationId)
               }
             }.compile.drain.start.void)
     } yield ()
--- a/modules/calibrations/src/main/scala/lucuma/odb/calibrations/Main.scala
+++ b/modules/calibrations/src/main/scala/lucuma/odb/calibrations/Main.scala
@@ -18,6 +18,7 @@ import lucuma.core.model.Access
 import lucuma.core.model.User
 import lucuma.odb.Config
 import lucuma.odb.graphql.enums.Enums
+import lucuma.odb.graphql.topic.CalibTimeTopic
 import lucuma.odb.graphql.topic.ObservationTopic
 import lucuma.odb.sequence.util.CommitHash
 import lucuma.odb.service.Services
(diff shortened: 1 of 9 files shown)


modules/calibrations/src/main/scala/lucuma/odb/calibrations/Main.scala
```diff
@@ -122,23 +126,37 @@ object CMain extends MainParams {
...
+                          LocalDateTime.of(t, LocalTime.MIDNIGHT).toInstant(ZoneOffset.UTC)
+                        )
+                } yield ()
+              }
+            }.compile.drain.start.void)
+      _  <- Resource.eval(Logger[F].info("Start listening for calibration time changes"))
+      _  <- Resource.eval(calibTopic.subscribe(100).evalMap { elem =>
+              services.useTransactionally{
+                for {
+                  t <- Sync[F].delay(LocalDate.now(ZoneOffset.UTC))
```
swalker2m, line 158: `t` isn't used here?  Do we need it?

modules/calibrations/src/main/scala/lucuma/odb/calibrations/Main.scala
```diff
@@ -122,23 +126,37 @@ object CMain extends MainParams {
...
+                  _ <- calibrationsService.recalculateCalibrationTarget(elem.programId, elem.observationId)
```
swalker2m, line 159: At first glance I thought that there wasn't a time element to this, but I believe it is taking the time from the observation time in the observation.  Maybe a comment?  It says "Start listening for calibration time changes" which. I think means the observation time of the calibration observation?

modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
```diff
@@ -369,6 +419,17 @@ object CalibrationsService {
...
+
+    val selectCalibrationTimeAndConf: Query[Observation.Id, (Observation.Id, Option[Timestamp], Option[ObservingModeType])] =
+      sql"""
+        SELECT
+            c_observation_id,
+            c_observation_time,
+            c_observing_mode_type
+          FROM v_generator_params
+          WHERE c_observation_id = $observation_id AND c_calibration_role IS NOT NULL
+          """.query(observation_id *: core_timestamp.opt *: observing_mode_type.opt)
```
swalker2m, line 431: I'm wondering about adding observation time to the generator params view.  It isn't really needed to generate the sequence, at least not yet.  All these things come from the observation itself in `t_observation`.  Is there any reason to not just query `t_observation` directly?

swalker2m (commented):
This looks great, with a few questions.

modules/calibrations/src/main/scala/lucuma/odb/calibrations/Main.scala
cquiroz: you're right, copy-paste error

modules/calibrations/src/main/scala/lucuma/odb/calibrations/Main.scala
cquiroz: Right it listens for events every time a the observation time of a calibration observation changes

swalker2m (approved):
Looks great!  :+1:  Did you want to remove that extra local date calculation?

PR #1362, opened 2024-08-28
Update scala to 3.5.0

cquiroz:


toddburnside (approved):
Wow. A lot of unused imports that weren't caught before.

PR #1363, opened 2024-08-28
Remove unneeded calibration observations

cquiroz:
This PR deletes calibration observations that are no longer needed (e.g. if the original configuration is not in use anymore)

Observations are fully deleted in the DB unlike regular observations which are marked as deleted. Also dependencies are removed, like targets.

A new `EditType.Deleted` is added to notify clients when these observations are deleted.

This change will break explore

This change ended up being a lot more complicated than I expected. many thanks to @tpolecat and @swalker2m for helping to get this over the finishing line


modules/service/src/main/scala/lucuma/odb/service/ObservationService.scala
```diff
@@ -264,6 +269,52 @@ object ObservationService {
...
+                    // set the existence to deleted, so it gets removed from groups too
+                    updateObservations(existenceOff, sql"$observation_id"(o))
+                  )
+                  // Look for the linked targets
+          tids <- session.execute(Statements.linkedTargets(oids))(oids.toList)
+                  // Delete asterisms and targets
+          _    <- NonEmptyList
+                    .fromList(tids)
+                    .map(tids => session.executeCommand(Statements.deleteLinkedAsterisms(tids)).void)
+                    .getOrElse(Applicative[F].unit)
```
tpolecat, line 307: I think you can do these as `NonEmptyList.fromList(...).traverse(...)`

tpolecat (approved):
A few comments but this looks fine. I would like @swalker2m to have a look as well before merging.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -3084,11 +3089,13 @@ type ObservationEdit { #implements Event {
   """
   editType: EditType!
 
+  observationId: ObservationId!
+
   """
   Edited object
```
swalker2m, line 3095: maybe add `(if not deleted)`?  It is the case that when deleted this is `null` correct?

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -988,6 +988,11 @@ enum EditType {
   EditType Updated
   """
   UPDATED
+
+  """
+  EditType Deleted
+  """
+  DELETED
```
swalker2m, line 995: Seems like someone might be forgiven for thinking that they'd get a `DELETED` event when setting the existence flag to `DELETED` but they'd be wrong?

modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
```diff
@@ -147,39 +154,41 @@ object CalibrationsService {
           }
         } else none.pure[F]
 
-      private def uniqueConfiguration(pid: Program.Id, selection: ObservationSelection)(using Transaction[F]): F[(List[GmosNConfigs], List[GmosSConfigs])] = {
-        val all: F[List[Config.GmosNorth | Config.GmosSouth]] = services.generatorParamsService.selectAll(pid, selection = selection)
-          .map(_.values.toList.map(_.map(_.observingMode)).collect {
-            case Right(mode: Config.GmosNorth) => mode
-            case Right(mode: Config.GmosSouth) => mode
+      private def uniqueConfiguration(pid: Program.Id, selection: ObservationSelection)(using Transaction[F]): F[(List[(Observation.Id, GmosNConfigs)], List[(Observation.Id, GmosSConfigs)])] = {
+        val all: F[List[(Observation.Id, Config.GmosNorth) | (Observation.Id, Config.GmosSouth)]] = services.generatorParamsService.selectAll(pid, selection = selection)
```
swalker2m, line 158: I think I asked this before, but will this keep growing as we add more instruments?

swalker2m (approved):
Congratulations on figuring this out. I probably would have given up.  :+1:

modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
cquiroz: yes it will

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
cquiroz: Yes, they'd be wrong

cquiroz: Do you think another word would be more appropriate?

PR #1366, opened 2024-08-29
Fix best calibration target selection

cquiroz:


swalker2m (approved):
Looks reasonable :+1:

PR #1369, opened 2024-08-30
Delete orphan calibration targets

cquiroz:
When the observation time changes a new target may be assigned to calibration observations. When that happens the previous target remains though it is unused

This PR adds a query to delete unused targets and it is run an a program basis every time the obs time changes


swalker2m (approved):
This looks good to me.

PR #1379, opened 2024-09-02
Add ability to save guide target, change guideEnvironment

toddburnside:
This adds a `guideEnvironment` on the in the `targetEnvironment` of the observation. Instead of returning the ordered list of potential guide stars (as `guideEnvironments` does) only a single guide star, or an error, will be returned. If a guide star name has been set, that guide star will be returned. If not set, or the observation has changed since the name was set, gaia will be queried and the "best" available guide star will be returned. There is a `lookupIfUndefined` parameter on `guideEnvironment` that explore can set to false, such that a null can be returned when the guide star is not set. And, if only the name is requested in the query gaia will not be called unless the name is not set (or is invalid) AND lookupIfUndefined is true.

I left the old `guideEnvironments` query to make it easier for navigate to convert. Once navigate is on the new `guideEnvironment`, guideEnvironments will be removed.

### DIFF
--- a/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -4186,17 +4186,13 @@ type SetupTime {
 
 """
 Input parameters for setting the guide star name for an observation.
+Identify the observation to clone by specifying either its id or reference.  If
+both are specified, they must refer to the same observation.  If neither is
+specified, an error will be returned.
 """
 input SetGuideTargetNameInput {
-  """
-  The program id for the observation being set.
-  """
-  programId: ProgramId!
-
-  """
-  The id of the observation to set.
-  """
-  observationId: ObservationId!
+  observationId: ObservationId
+  observationReference: ObservationReferenceLabel
 
   """
   The name of the guide star. This must satisfy the regular expression "^Gaia DR3 (-?\d+)$" where the
--- a/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -2928,6 +2928,11 @@ type Mutation {
   """
   setAllocations(input: SetAllocationsInput!): SetAllocationsResult!
 
+  """
+  Set the name of the guide target for an observation.
+  """
+  setGuideTargetName(input: SetGuideTargetNameInput!): SetGuideTargetNameResult!
+
   "Set the program reference."
   setProgramReference(input: SetProgramReferenceInput!): SetProgramReferenceResult!
 
@@ -3187,9 +3192,14 @@ Observation times properties
 """
 input ObservationTimesInput {
   """
-  Expected execution time used for time-dependent calculations such as average parallactic angle
+  Expected execution time used for time-dependent calculations such as average parallactic angle and guide star selection.
   """
   observationTime: Timestamp
+
+  """
+  Expected observation duration used in conjunction with observationTime. If not set, remaining observation time is used.
+  """
+  observationDuration: TimeSpanInput
 }
 
 type Offset {
@@ -4175,6 +4185,34 @@ type SetupTime {
 }
 
 """
+Input parameters for setting the guide star name for an observation.
+"""
+input SetGuideTargetNameInput {
+  """
+  The program id for the observation being set.
+  """
+  programId: ProgramId!
+
+  """
+  The id of the observation to set.
+  """
+  observationId: ObservationId!
+
+  """
+  The name of the guide star. This must satisfy the regular expression "^Gaia DR3 (-?\d+)$" where the
+  numeric part is the Gaia source_id. Omit or set to null to delete.
+  """
+  targetName: NonEmptyString
+}
+
+"""
+The result of setting the guide target name for an observation.
+"""
+type SetGuideTargetNameResult {
+  observation: Observation
+}
+
+"""
 Input for setting the program reference.  Identify the program to update with one
 of `programId`, `proposalReference` or `programReference`.  If more than one of
 these is specified, all must match.  Use `SET` to specify the new program
@@ -4973,7 +5011,6 @@ type UpdateGroupsResult {
   """
   hasMore: Boolean!
 }
-
 
 """
 Observation selection and update description.  Use `SET` to specify the changes, `WHERE` to select the observations to update, and `LIMIT` to control the size of the return value.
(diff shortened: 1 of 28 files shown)


modules/sequence/src/main/scala/lucuma/odb/sequence/util/HashBytes.scala
```diff
@@ -71,9 +72,18 @@ object HashBytes {
...
+  }
+
   given HashBytes[TimeSpan] with {
     def hashBytes(a: TimeSpan): Array[Byte] =
       HashBytes[Long].hashBytes(a.toMicroseconds)
   }
 
-}
\ No newline at end of file
+  given [A](using HashBytes[A]): HashBytes[Option[A]] with {
```
toddburnside, line 85: These behavior for the hash of Option made sense in GuideService. If it doesn't make sense globally, it could be localized there.

modules/service/src/test/scala/lucuma/odb/graphql/query/guideEnvironment.scala
toddburnside, line 1: This is actually a new file, but since the tests for guideEnvironments was improperly named this, it is being diffed with the guideEnvironments tests.

modules/service/src/test/scala/lucuma/odb/graphql/query/guideEnvironments.scala
toddburnside, line 1: These are the old guideEnvironments tests. The file was just renamed.

swalker2m:
> I left the old guideEnvironments query to make it easier for navigate to convert. Once navigate is on the new guideEnvironment, guideEnvironments will be removed.

Does Explore call GAIA directly in order to provide the `<` and `>` buttons for cycling through options?

cquiroz:
> > I left the old guideEnvironments query to make it easier for navigate to convert. Once navigate is on the new guideEnvironment, guideEnvironments will be removed.
> 
> Does Explore call GAIA directly in order to provide the `<` and `>` buttons for cycling through options?

Yes, it does the guidestar calculation on the browser

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -4174,6 +4184,34 @@ type SetupTime {
...
+input SetGuideTargetNameInput {
+  """
+  The program id for the observation being set.
+  """
+  programId: ProgramId!
+
+  """
+  The id of the observation to set.
+  """
+  observationId: ObservationId!
```
swalker2m, line 4199: Given the `observationId`, do we need the `programId` as well?  Shouldn't we allow the observation reference as an alternative as well (for example as in `CloneObservationInput`).

modules/sequence/src/main/scala/lucuma/odb/sequence/util/HashBytes.scala
swalker2m: Seems good to me.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
toddburnside: Yeah. You're right. I guess I was just being lazy while I developed this. I'll go back and make those changes.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/TargetEnvironmentMapping.scala
```diff
@@ -113,23 +121,50 @@ trait TargetEnvironmentMapping[F[_]: Temporal]
...
         }
 
     effectHandler(readEnv, calculate)
   }
 
+  def guideEnvironmentQueryHandler: EffectHandler[F] = {
+    val readQueryAndCursor: (Query, Cursor) => Result[GuideService.GuideEnvironmentOptions] = (query, cursor) =>
+      // Determine if the query only contains a request for the target name(s). If so, we can avoid
+      // calling gaia under some circumstances. Explore will only be requesting the name, but
+      // Navigate will want everything.
```
swalker2m, line 141: So if we specify `lookupIfUndefined: true`, and it is undefined, we call GAIA and calculate the result but don't store it for future reference?  I guess maybe I would have expected it to be cached.

swalker2m (approved):
The code looks great to me.  I just wonder if having done the remote call and calculation we should be saving the result?

modules/service/src/main/scala/lucuma/odb/graphql/mapping/TargetEnvironmentMapping.scala
toddburnside: Right. Nothing is cached at the moment. Probably only navigate will be looking for the full guide star information with lookupIfUndefined === true, and since it should only do this just before it is observed I'm told the short delay is not a problem. And if it only asks for the guide star once, caching wouldn't help. That said, with the new way of doing things, we could more easily cache the result.

toddburnside: We could also have explore send us the entire guide star info instead of just the name. This could change frequently while the user is choosing guide stars, so maybe we could just save it as json instead of dealing with actually saving and (especially) deleting targets from the db?

swalker2m: Every time they click on the `>` button it sets the value in the database or do they have to click on "Accept" or something?

What would we do if we save it as JSON and then need to update the target JSON?  Seems like we'd need to write a Scala data migration in that case.  I guess the goal of this task was just to keep up with the selection which it seems like you accomplished.  We can consider whether/what to cache at another opportunity (i.e., maybe never :smiley:) if you like.

toddburnside: > Every time they click on the > button it sets the value in the database or do they have to click on "Accept" or something?

Every time they click the > or < button. Although we can do some debouncing.

> What would we do if we save it as JSON and then need to update the target JSON? Seems like we'd need to write a Scala data migration in that case

Yeah. That wouldn't be fun. However, wouldn't any change we make to the targets invalidate the generator hash and thereby invalidate the guide target?

> We can consider whether/what to cache at another opportunity (i.e., maybe never 😃) if you like.
That sounds like great plan to me.

swalker2m: > However, wouldn't any change we make to the targets invalidate the generator hash and thereby invalidate the guide target?

That's true and in fact the commit hash is folded into the calculation so any change whatsoever to the code would invalidate it.  Is that going to be a problem?  If they selected a particular target and we do some unrelated update will it disappear?  Should we just keep up with the selection regardless and do validation apart from that?

toddburnside: The target selection needs to be done after the observation time/duration is set. So, for the actual observation the observer will need to select the guide target immediately before the observation is observed. I'm hoping we don't do updates to production in the middle of an observing night. 

The PI or other user can make guide star selections while they "play around", but it is the observer who will make the final selection. At the moment, at least, we're not trying to preserve what the PI selected. I'm hoping we never have to.

PR #1380, opened 2024-09-03
Twilight calibrations

cquiroz:
This PR adds twilight calibrations. They are similar to spectphoto, thus the code has been reorganized to avoid duplication.

Right now there is no sequence for twilights, as a side effect of this the twilights are not deleted when the configuration changes


swalker2m:
> Right now there is no sequence for twilights, as a side effect of this the twilights are not deleted when the configuration changes

Could you expand on this?  I'm not sure I see the relationship.

modules/service/src/test/scala/lucuma/odb/graphql/mutation/updateObservations.scala
```diff
@@ -530,7 +530,7 @@ class updateObservations extends OdbSuite
 
   }
 
-  test("target environment: fail to set an asterism across programs") {
+  test("target environment: fail to set an asterism across programs".ignore) {
```
cquiroz, line 533: @tpolecat This is the failing test, note that if I comment out the 0911 migration this test passes

tpolecat (approved):
I don't have any insight about the logic, but the code looks fine. 👍🏻

PR #1382, opened 2024-09-06
Make separate guide target name query for Explore

toddburnside:
I was trying to use the same query for Explore and Navigate, but their needs are quite different. This adds a query so Explore can get the target name without worrying about some of the errors that Navigate expects. It modifies the guideEnvironment to have the return value to be required.

### DIFF
--- a/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -10828,17 +10828,13 @@ type TargetEnvironment {
   ): [GuideEnvironment!]!
 
   """
-  The guide star(s) and related information
+  The guide target(s) and related information.
+  If a guide target has been set via `guideTargetName`, that target will be
+  returned. If it not found or not usable, an error will be returned.
+  If no guide target has been set, or it has been invalidated by observation/target
+  changes, Gaia will be searched for the best guide target available.
   """
-  guideEnvironment(
-    """
-    If this is set to false, a null will be returned if a guide star is not
-    already set or the current guide star has been invalidated by changes.
-    If this is set to true, gaia will be searched to find the best guide star
-    in the above scenario.
-    """
-    lookupIfUndefined: Boolean! = true
-  ): GuideEnvironment
+  guideEnvironment: GuideEnvironment!
 
   """
   Availability of guide stars during a specified time range.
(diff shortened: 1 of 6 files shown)


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -10828,17 +10828,13 @@ type TargetEnvironment {
...
-  guideEnvironment(
-    """
-    If this is set to false, a null will be returned if a guide star is not
-    already set or the current guide star has been invalidated by changes.
-    If this is set to true, gaia will be searched to find the best guide star
-    in the above scenario.
-    """
-    lookupIfUndefined: Boolean! = true
-  ): GuideEnvironment
+  guideEnvironment: GuideEnvironment!
```
cquiroz, line 10837: Why was the parm removed?

toddburnside: Because this is used by Navigate, and it always wants the odb to find a default value if one isn't set. The parameter would only have been set to false by Explore.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -10857,6 +10853,13 @@ type TargetEnvironment {
   When set, overrides the default base position of the target group
   """
   explicitBase: Coordinates
+
+  """
+  The name of the guide target, if any, set by `setGuideTargetName`.
+  If the name is no longer valid or a sequence cannot be generated, null will
+  be returned.
+  """
```
swalker2m, line 10861: That it can disappear after having been set could be disquieting to a sensitive person.  I wonder if we should keep it and just issue warnings / errors when the guide environment is sought.

swalker2m (approved):
Code looks great, just a question about the semantics.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
toddburnside: I think Explore needs to know when it is invalid. So that it can choose the best one of the (possibly) new choices.

toddburnside: Navigate will get the errors.

PR #1386, opened 2024-09-07
(WIP) New GMOS Sequence 

swalker2m:
## Summary 
Reimplements GMOS Long Slit sequence generation.  In a nutshell we stream step records from the database (as in `fs2.Stream`) and fold state over them.  When we run out of recorded steps then we use the resulting state to generate the remainder of the sequence.  This makes it possible to take into account failed steps, delays that invalidate previous calibrations, etc.

From the client perspective, the big difference is that Observe needs to refetch the `nextAtom` after every step, and be prepared that the atom will change.

Keep in mind that the sequence generation is now time-dependent.  Calibrations are only valid for so long (1.5 hours) so if enough time passes they may need to be regenerated.  Science datasets without timely calibrations may need to be repeated.  **This means that the sequence can "change" even if there are no updates to anything in the database.**

## Notes:
* All required smartgcal configurations are resolved _before_ the sequence generation, instead of during.  That means that the generated sequence can be `Pure` instead of effectful.
* Time estimation is done as the steps are generated instead of at a higher level.
* The API hasn't changed (much). I deprecated the `stepCount` parameter that was used when recording an atom since it is no longer used.

## Science Sequence
The science sequence is now a series of "atoms" where each atom contains an arc, a flat, and then how ever many science exposures we can fit into an hour (ignoring overheads).  For example, if the exposure time is 20 minutes 3 science exposures fit in the atom.  Each atom has a distinct wavelength dither and spatial offset, and we attempt to spread the science exposures across them somewhat evenly.

In this implementation, there is no penalty for executing the steps in an atom in any order or for only performing a subset of the science steps.  Failed datasets are repeated during the execution of an atom if time permits, or otherwise grouped with subsequent atoms for the same wavelength dither and spatial offset.


swalker2m:
An example of how an atom could change in mid-flight: for a 20 minute exposure the atoms are of the shape:

```
1. Arc
2. Flat
3. Science
4. Science
5. Science
```

This is followed by a wavelength dither and a spatial offset and the process repeats.  While working on this atom, say during step 4, the flat and arc are determined to be failures.  After completing Step 4, if the atom is fetched again, the remainder would now be:

```
1. Science (former step 5)
2. Arc (former step 1)
3. Flat (former step 2)
```

So, the Arc and Flat are repeated but not until the end of the atom to avoid moving the science fold.

swalker2m:
More examples and discussion in [ShortCut 3374](https://app.shortcut.com/lucuma/story/3374/update-gmos-sequence-generator).

modules/sequence/src/main/scala/lucuma/odb/sequence/SequenceGenerator.scala
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Sequence generator.
+ *
+ * @tparam D dynamic instrument config type
+ */
+trait SequenceGenerator[D]:
+
+  def generate(t: Timestamp): Stream[Pure, (ProtoAtom[(ProtoStep[D], Int)], Int)]
+
+  def record(step: StepRecord[D])(using Eq[D]): SequenceGenerator[D]
```
swalker2m, line 23: `record` updates the state with the next step from the database.  When all the existing step records have been processed, `generate` is called to obtain the sequence.  The `Int` part of the tuple has to do with maintaining the atom and step ids.  It is annoying.

modules/service/src/main/scala/lucuma/odb/service/GmosSequenceService.scala
```diff
@@ -26,6 +27,14 @@ import Services.Syntax.*
 
 trait GmosSequenceService[F[_]] {
 
+  def selectGmosNorthStepRecords(
+    observationId: Observation.Id
+  ): Stream[F, StepRecord[DynamicConfig.GmosNorth]]
```
swalker2m, line 32: @tpolecat Dumb question: does the creation of the `Stream` of steps need to be done with a `Transaction`, or is it only needed when the `Stream` is actually executed?  In other words, should this be `using Transaction`?

tpolecat: It's only needed on execution; the transaction resource, when used, conceptually just puts `begin` and `commit` around whatever you're doing. So as long as you compose this with a transaction eventually you'll be fine.

modules/sequence/src/main/scala/lucuma/odb/sequence/gmos/longslit/Science.scala
swalker2m, line 1: I think if one wanted to limit their review to just one file and ignore the rest, this would be the one to look at.

tpolecat (approved):
Man this was a ton of work. I'm glad you were able to get it to the finish line. I don't see anything obviously amiss, so let's ship it and see how things go. 👍🏻

PR #1387, opened 2024-09-09
Fix trigger on t_target_asterism to exclude deleted targets from observation title (sc-3036)

cquiroz:


modules/service/src/test/scala/lucuma/odb/graphql/issue/shortcut/3062.scala
```diff
@@ -0,0 +1,56 @@
...
+
+import io.circe.literal.*
+import lucuma.core.model.User
+
+class ShortCut_3062 extends OdbSuite {
+  val pi: User    = TestUsers.Standard.pi(nextId, nextId)
+
+  override lazy val validUsers: List[User] = List(pi)
+
+  test("Obs name should be updated when a target is deleted".ignore) {
```
cquiroz, line 15: This test case fails but should pass once sc-3062 is fixed

tpolecat (approved):
👍🏻 cool, thanks!

PR #1393, opened 2024-09-12
add "for review" bit

tpolecat:
This adds a `forReview` bit to `type Observation`, which will be set to `true` after ITAC and must be set to false prior to execution. A validation error appears on observations where `forReview = true`.


cquiroz (approved):
LGTM

PR #1399, opened 2024-09-16
split support roles

tpolecat:
This changes the ODB to accommodate https://github.com/gemini-hlsw/lucuma-core/pull/945, which splits the support role into primary and secondary.


rpiaggio (approved):
LGTM

PR #1407, opened 2024-09-25
Don't create validations for calibration observations

toddburnside:


toddburnside:
@tpolecat If you are using the validations in the observation status process, you might want to review this and make sure it doesn't affect you.

modules/service/src/main/scala/lucuma/odb/service/ObservationService.scala
```diff
@@ -713,21 +721,24 @@ object ObservationService {
...
 
-        val initialMap = 
-          for {
-            (genVals, op) <- generatorValidations
-            cfpVals       <- cfpValidations
-            itcVals       <- op.filter(_ => cfpVals.isEmpty).foldMapM(itcValidations) // only compute this if cfp and gen are ok
-            bandVals      <- validateScienceBand
-          } yield (genVals |+| itcVals |+| cfpVals |+| bandVals)
+        obsInfo.flatMap: info =>
+          if (info.role.isDefined) List.empty.pure // it's a calibration
```
cquiroz, line 725: I still wonder about no warnings. Calibrations can be ill defined, e.g. no guide star

modules/service/src/test/scala/lucuma/odb/graphql/DatabaseOperations.scala
```diff
@@ -1484,6 +1484,12 @@ trait DatabaseOperations { this: OdbSuite =>
       .use(_.prepareR(command).use(_.execute(role, tid).void))
   }
 
+  def setObservationCalibratioRole(oid: Observation.Id, role: CalibrationRole): IO[Unit] = {
```
cquiroz, line 1487: I think this was defined in `CalibrationService` already

toddburnside: Ah. Can I call a service method from the tests?

modules/service/src/main/scala/lucuma/odb/service/ObservationService.scala
toddburnside: Hmm. Even if they are automatically generated? And if so, what can the user do about that? Who should be notified and how?

tpolecat:
In terms of workflow it seems like calibrations shouldn't really count, so this makes sense to me.

modules/service/src/test/scala/lucuma/odb/graphql/DatabaseOperations.scala
cquiroz: Sure, though I think it ought to be a static method

modules/service/src/main/scala/lucuma/odb/service/ObservationService.scala
cquiroz: Maybe a question for user?

toddburnside: Andy commented on the shortcut ticket:

> That's a good question. I am optimistically hopeful that all of our automatically generated calibration observations will be perfect. I could imagine that if all targets were accidentally deleted from the Specphot standard storage program, or a target was added without an SED or brightness information those could cause trouble, but we probably don't want to bother users about that.

> I don't know if we need to solve this now, but we may need a way to give a warning to ourselves if an automatic calibration fails to generate

cquiroz (approved):
LGTM

PR #1408, opened 2024-09-27
ShortCut 22260: Error when updating between bandNormalized and emissionLines

toddburnside:
This turned out to be a general problem when attempting to update from `bandNormalized` to `emissionLines` or vice versa for both `point` and `uniform` source profiles.


swalker2m (approved):
I think this looks great.  :+1:

PR #1410, opened 2024-10-01
see if this fixes the deadlock problem

tpolecat:
`Services` has a `GraphQLService` that can be used to make recursive GraphQL queries from the service layer, which is sometimes more reasonable than doing complex SQL queries. Unfortunately doing so allocates another connection, which can lead to deadlocks if the pool is empty. This works around the issue by using another mapping that is bound to the `Services` database session rather than to the connection pool.

I feel shame, obvs.


swalker2m (approved):
This is subtle.  I'd never have found it.  :+1:

PR #1413, opened 2024-10-07
Shortcut 3660: distinguish remote ITC call errors from observation definition errors

swalker2m:
This PR has two purposes:

1. When a sequence is requested, the ITC must be consulted for the integration time and exposure counts (or have been consulted previously and cached).  The sequence generation can fail if the observation isn't fully defined and the ITC remote call itself can fail for various reasons.  Previously any ITC error produced when generating the sequence would be mapped to the `OdbError` tagged `sequence_unavailable`.  Now when the problem with the ITC call is truly an issue in the ITC it is distinguished as such and is returned as `itc_error`.
2. Adds an often failing test case (commented out) that illustrates the issue reported by @rpiaggio.  Namely, when a query requests _both_ the ITC results and the sequence and both fail, sometimes only one error is returned.  Other times both errors are returned.  This failure happens at a higher level, I suspect, and perhaps has something to do with the fact that both of these fields are handled by `EffectHandler`s.

### DIFF
--- a/modules/service/src/test/scala/lucuma/odb/graphql/query/executionFailures.scala
+++ b/modules/service/src/test/scala/lucuma/odb/graphql/query/executionFailures.scala
@@ -114,54 +114,6 @@ class executionFailures extends ExecutionTestSupport {
     }
   }
 
-/**
-  In this test case, the query requests both the observation `execution` field
-  and the `itc` field.  Neither query should be successful because there is no
-  observing mode.  We expect two errors, one for `execution` and one for `itc`
-  and often actually get them.  Sometimes though only the `itc` error is
-  returned.
-
-  test("cannot generate, missing mode + failed itc query") {
-    val setup: IO[Observation.Id] =
-      for {
-        p <- createProgram
-        t <- createTargetWithProfileAs(pi, p)
-        o <- createObservationWithNoModeAs(pi, p, t)
-      } yield o
-
-    setup.flatMap { oid =>
-      expectOdbErrors(
-        user  = pi,
-        query =
-          s"""
-             query {
-               observation(observationId: "$oid") {
-                 itc {
-                   science {
-                     selected { signalToNoise }
-                   }
-                 }
-                 execution {
-                   config {
-                     gmosNorth {
-                       static {
-                         stageMode
-                       }
-                     }
-                   }
-                 }
-               }
-             }
-           """,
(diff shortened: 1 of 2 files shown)


modules/service/src/test/scala/lucuma/odb/graphql/query/executionFailures.scala
```diff
@@ -104,21 +107,60 @@ class executionFailures extends ExecutionTestSupport {
...
+      )
+    }
+  }
+
+/*
+  In this test case, the query requests both the observation `execution` field
+  and the `itc` field.  Neither query should be successful because there is no
+  observing mode.  We expect two errors, one for `execution` and one for `itc`
+  and often actually get them.  Sometimes though only the `itc` error is
+  returned.
```
swalker2m, line 122: This would seem to be an issue with Grackle?

tpolecat: This is weird because Grackle only demands `MonadThrow` for its effect type, which means it shouldn't be able to do anything concurrently and results should be deterministic. Could this test be interacting nondeterministically with others in the suite? Maybe move it to its own suite so it gets its own database?

swalker2m: Thanks for the suggestion.  I isolated the test case into `lucuma.odb.graphql.issue.shortcut.3660`.  Upon attempting it I got 3 consecutive failures, followed by 5 success, followed by 1 failure.  All the failures mss the same error:

```
values are not the same
=> Obtained
Set(
  InvalidObservation(
    observationId = Id(
      value = 256
    ),
    detail = Some(
      value = "ITC cannot be queried until the following parameters are defined: observing mode."
    )
  )
)
 Set(
+  SequenceUnavailable(
+    detail = Some(
+      value = "Could not generate a sequence from the observation o-100: observing mode"
+    )
+  ),
```

swalker2m:
Regardless of the existing, underlying issue with nondeterministic error reporting, this PR separates and reports observation definition issues vs. remote ITC service errors.  I'd like to propose merging this for that alone. It doesn't introduce the problem with errors in multiple queries, it just exposes it in a test case (which is `@Ignore`ed).

tpolecat (approved):

PR #1415, opened 2024-10-08
configuration request bulk actions

tpolecat:
This adds two features for configuration requests.

- It adds deduplication, so only one request is generated per unique configuration per program.
- It adds bulk operations "add requests for all observations in this program" and "delete all requests for this program", which are executed when the proposal is submitted or retracted, respectively.

The "add requests for all observations in this program" operation accumulates errors and places them in the warnings; if a request can't be created for an observation due to an incomplete observation then the client will receive one or more warnings about it. This simplifies testing but should not happen in practice because incomplete observation definitions will lead to validation errors that should prevent submission.

The bulk operations are triggered by proposal state changes that already include access control checks, so there are not additional checks here.


swalker2m:
> The bulk operations are triggered by proposal state changes that already include access control checks, so there are not additional checks here.

Otherwise, for a one-off, there's a "button" in Explore that hits the `createConfigurationRequest` mutation?

swalker2m (commented):
What happens after a request is created?  Is there a trigger or something that ends up notifying someone?

swalker2m (approved):
Looks great to me. :+1:

toddburnside (approved):
LGTM - sorry for the delay, I was trying to get Explore ready for Shane's PR

tpolecat:
> > The bulk operations are triggered by proposal state changes that already include access control checks, so there are not additional checks here.
> 
> Otherwise, for a one-off, there's a "button" in Explore that hits the `createConfigurationRequest` mutation?

Yes, that's the plan.

tpolecat:
> What happens after a request is created? Is there a trigger or something that ends up notifying someone?

Yes, also the plan. There will probably be a daily email or something so they get batched up.

PR #1416, opened 2024-10-08
Time Accounting by Band

swalker2m:
Changes the `timeCharge` field at the `Program` level to a `BandedTime`, which groups a `ScienceBand` and a `CategorizedTime`.  In other words, it groups the time charges by bands at the program level.

__This will presumably break Explore.__


modules/schema/src/main/scala/lucuma/odb/data/BandedTime.scala
```diff
@@ -0,0 +1,19 @@
...
+
+case class BandedTime(
+  band: Option[ScienceBand],
+  time: CategorizedTime
+)
+
+object BandedTime:
+  val Empty: BandedTime = BandedTime(None, CategorizedTime.Zero)
+
+  given Eq[BandedTime] =
```
tpolecat, line 18: Might be useful to define `Order` here. Looks like it's defined for the member types.

tpolecat (approved):
Looks good. 👍🏻

toddburnside:
> This will presumably break Explore.

As far as I can see, Explore doesn't look at the `timeCharge`

modules/schema/src/main/scala/lucuma/odb/data/BandedTime.scala
swalker2m: :+1: Thanks, that's a good point.  I've updated it but will add property tests when I move this to core.

PR #1420, opened 2024-10-11
Proprietary Period (Months)

swalker2m:
The proprietary period is the amount of time (specified in months) that data taken for a program should remain unavailable to the public at large.  This value should be written into FITS headers of the datasets, as I understand it.

This PR adds a proprietary period property to:

* the definition of each call for proposals type (as a default value for each type)
* the call for proposals itself, since sometimes they want a different proprietary period default for a particular call
* the program properties, since sometimes individual programs need a distinct proprietary period


tpolecat (approved):
This looks good. 👍🏻

PR #1428, opened 2024-10-15
GoaProperties

swalker2m:
This creates a new `GoaProperties` (Gemini Observatory Archive) object and associated input and moves the `proprietaryMonths` into it along with a couple of new properties:

```
"Gemini Observatory Archive properties for a particular program."
type GoaProperties {

  """
  How many months to withhold public access to the data.  This property is
  applicable to science programs, defaults to the proprietary period associated
  with the Call for Proposals if any; 0 months otherwise.
  """
  proprietaryMonths: NonNegInt!

  """
  Whether the PI wishes to be notified when new data are received. This property
  is applicable to science programs and defaults to true.
  """
  shouldNotify: Boolean!

  """
  Whether the header (as well as the data itself) should remain private.  This
  property is applicable to science programs and defaults to false.
  """
  privateHeader: Boolean!

}
```
All of these go into the FITS header and are used by the archive.

I originally put `proprietaryMonths` at the top level in `Program` but, since there are these other GOA properties as well, decided to group them all before anybody has had a chance to use it. 🤞

### DIFF
--- a/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -3455,13 +3455,10 @@ input ProgramPropertiesInput {
   name: NonEmptyString
 
   """
-  Set the proprietaryMonths for this program.  This property is only applicable
-  to normal science programs (i.e., not CAL, ENG, MON programs etc.). It is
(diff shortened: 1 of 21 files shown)


tpolecat (approved):
Looks great 👍🏻

toddburnside (approved):
LGTM

PR #1431, opened 2024-10-16
Switch to UserInvitation.Id in OdbError.InvitationError

swalker2m:
A trivial improvement that was enabled when `UserInvitation` was moved to core.


cquiroz (approved):
LGTM

PR #1440, opened 2024-10-18
config req events

tpolecat:
This adds an edit event for configuration requests, per @toddburnside request.


toddburnside (approved):
LGTM. Thanks!

PR #1432, opened 2024-10-16
Support Twilight Calibration Observation Sequence

swalker2m:
Adds sequence generation for [twilight calibration observations](https://app.shortcut.com/lucuma/story/3410/create-twilight-calibration-observations).  These have a single 30 second exposure, no acquisition and no gcals.

PR #1443, opened 2024-10-21
Twilight Again

swalker2m:
#1432 added sequence generation for twilight calibration observations.  Unfortunately it failed to take into account that these observations use degenerate targets which don't pass observation validation. This PR updates `GeneratorParams` to carry deferred target errors instead of stopping immediately when the target isn't a valid science target.

### Before: ITC inputs required
```scala
case class GeneratorParams(
  itc:             GeneratorAsterismParams,
  observingMode:   ObservingMode,
  calibrationRole: Option[CalibrationRole]
)
```

### After: ITC inputs may be missing
```scala
case class GeneratorParams(
  itcInput:        Either[ItcInput.Missing, ItcInput.Defined],
  observingMode:   ObservingMode,
  calibrationRole: Option[CalibrationRole]
)
```

At sequence generation time, depending on the calibration role, the lack of a valid target may or may not be an actual issue.  In the case of a twilight, missing ITC parameters can be ignored.

### Disclaimer
_I'm not at all happy with the error handling. I want to circle back and clean this up if I can, but apparently this is needed for a demo on Wednesday so I'm offering it early._

### DIFF
--- a/modules/sequence/src/main/scala/lucuma/odb/sequence/data/GeneratorParams.scala
+++ b/modules/sequence/src/main/scala/lucuma/odb/sequence/data/GeneratorParams.scala
@@ -5,91 +5,33 @@ package lucuma.odb.sequence
 package data
 
 import cats.Eq
-import cats.Order.given
-import cats.data.NonEmptyList
-import cats.data.NonEmptyVector
+import cats.syntax.either.*
 import lucuma.core.enums.CalibrationRole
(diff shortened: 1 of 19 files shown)


cquiroz:
It seems this would be transparent to clients?

cquiroz:
we don't have validation errors for calibrations

cquiroz (approved):
LGTM

PR #1446, opened 2024-10-22
Support deleting unused twilight calibrations

cquiroz:


swalker2m (approved):
Looks good.

modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
```diff
@@ -150,8 +150,8 @@ object CalibrationsService extends CalibrationObservations {
           .generatorParamsService
           .selectAll(pid, selection = selection)
           .map(_.toList.collect {
-            case (oid, Right(GeneratorParams(Right(_), mode: Config.GmosNorth, _))) => (oid, mode)
-            case (oid, Right(GeneratorParams(Right(_), mode: Config.GmosSouth, _))) => (oid, mode)
+            case (oid, Right(GeneratorParams(_, mode: Config.GmosNorth, _))) => (oid, mode)
+            case (oid, Right(GeneratorParams(_, mode: Config.GmosSouth, _))) => (oid, mode)
```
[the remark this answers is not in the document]
cquiroz: Right, we need to consider all calibs in this case

modules/service/src/main/resources/logback.xml
```diff
@@ -10,14 +10,14 @@
   <logger name="org.http4s.blazecore.IdleTimeoutStage" level="OFF" />
   <logger name="org.http4s.blazecore.websocket.Http4sWSStage" level="OFF" />
 
-  <logger name="lucuma-odb" level="ALL" />
+  <logger name="lucuma-odb" level="WARN" />
```
toddburnside, line 13: Is this the level we want for this? I think it was INFO before Rob made it more verbose. I have no particular insight into what this should be, but maybe someone else does?

toddburnside (approved):
LGTM

modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
```diff
@@ -145,14 +146,22 @@ object CalibrationsService extends CalibrationObservations {
...
+        obs.toList.collect {
+          case (oid, Right(GeneratorParams(Right(_), mode: Config.GmosNorth, _))) => (oid, mode)
+          case (oid, Right(GeneratorParams(Right(_), mode: Config.GmosSouth, _))) => (oid, mode)
+        }
+
+      private def allCalibObservations(obs: Map[Observation.Id, EitherNel[GeneratorParamsService.Error, GeneratorParams]]): List[(Observation.Id, Config.GmosNorth | Config.GmosSouth)] =
+        obs.toList.collect {
+          case (oid, Right(GeneratorParams(_, mode: Config.GmosNorth, _))) => (oid, mode)
+          case (oid, Right(GeneratorParams(_, mode: Config.GmosSouth, _))) => (oid, mode)
+        }
```
swalker2m, line 159: `allCalibObservations` is misleading, I think, since it doesn't filter for calibration observations.  It will pick up twilight targets though as well as everything else in the map that it sent to it.

modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
```diff
@@ -279,21 +288,23 @@ object CalibrationsService extends CalibrationObservations {
...
           tgts          <- calibrationTargets(CalibrationTypes, referenceInstant)
           gsTgt         = CalibrationIdealTargets(Site.GS, referenceInstant, tgts)
           gnTgt         = CalibrationIdealTargets(Site.GN, referenceInstant, tgts)
-          uniqueSci     <- allObservations(pid, ObservationSelection.Science).map(uniqueConfiguration)
-          allCalibs     <- allObservations(pid, ObservationSelection.Calibration)
+          uniqueSci     <- allObservations(pid, ObservationSelection.Science)
+                              .map(allValidObservations)
+                              .map(uniqueConfiguration)
+          calibObs      <- allObservations(pid, ObservationSelection.Calibration)
+          allCalibs     = allValidObservations(calibObs)
```
swalker2m, line 300: `allCalibs` will be missing twilight targets since those without ITC params will be filtered out.  Sorry this is confusing.  Deferring the ITC errors introduces complexity and cognitive load.

modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
```diff
@@ -145,14 +146,22 @@ object CalibrationsService extends CalibrationObservations {
...
+        }
+
+      private def allObservations(pid: Program.Id, selection: ObservationSelection)(using Transaction[F]): F[Map[Observation.Id, EitherNel[GeneratorParamsService.Error, GeneratorParams]]] =
         services
           .generatorParamsService
           .selectAll(pid, selection = selection)
-          .map(_.toList.collect {
-            case (oid, Right(GeneratorParams(_, mode: Config.GmosNorth, _))) => (oid, mode)
-            case (oid, Right(GeneratorParams(_, mode: Config.GmosSouth, _))) => (oid, mode)
-          })
```
swalker2m, line 155: What if you used the `selection` to determine whether to match on `Right(itcInput)` or not?

modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
cquiroz: Right, we feed this a query with only calibrations and we are including both with and without itc, that's why it is `all`

modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
cquiroz: We also want to filter out science observations without itc

modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
cquiroz: Right, I need to fix this

swalker2m (approved):
:+1:

PR #1449, opened 2024-10-22
GeneratorParams missing data error handling

swalker2m:
Refactors `GeneratorParams` missing data errors a bit.  There is some deck chair shuffling but also some improvement on the margins.  Hopefully that's not overselling it.

Also updates the twilight sequence generation test to use a twilight observation directly produced by the calibration service itself instead of a faux twilight observation.


toddburnside (approved):
Looks great.

PR #1450, opened 2024-10-23
Add isComplete

swalker2m:
Adds an `isComplete` field to the observation `Execution` type.


tpolecat (approved):
Brilliant, thank you! :+1:

PR #1455, opened 2024-10-24
Workflow Status Part 2

tpolecat:
This updates the schema to provide a `workflow: ObservationWorkflow` field on `type Observation`. This type contains the existing validation error list (the top-level `validations` field will go away), as well as the current `ObservationWorkflowState` and a list of allowed transitions (for example, if the state is `DEFINED` and the proposal has been accepted then the user can set the state to `READY` or to `INACTIVE`). 

The mutation to allow the user to set a user-selectable `ObservationWorkflowState` will appear in a followup PR.

Also in a followup PR the fields `status`, `activeStatus`, `forReview`, and `validations` will be removed from `type Observation`; they are subsumed by the `workflow` struct. So the main goal here is to allow Explore to get ready for this.


swalker2m (approved):
Looks great to me. :+1:

PR #1458, opened 2024-10-25
Add GMOS read mode and gain to calibration configuration

swalker2m:
Calibration observations are required for distinct instrument configurations, but not all parameters matter for this purpose. For example, two identical GMOS long slit observations with distinct ROIs do not require distinct calibrations.  On the other hand, if the observations use different gratings then calibrations for each are required.

This PR adds the GMOS read mode and amp gain to the subset used to distinguish calibration instrument configurations.  It refactors the logic a bit as well to confine the configuration subset differences in one place.


modules/service/src/test/scala/lucuma/odb/graphql/feature/calibrations.scala
```diff
@@ -924,8 +999,8 @@ class calibrations extends OdbSuite with SubscriptionUtils {
                           "value" : null
                         }
                       }
-                    """,
-                  )})).mapN(_ ::: _)
+                    """
+                  })).mapN(_ ::: _)
```
swalker2m, line 1003: I think there may be a minor issue with the test case here.  The order that the events arrive doesn't seem to be guaranteed?  Changes elsewhere in the test case impact the ordering of events for this unrelated program.

cquiroz: That's right, order is not guaranteed and tests are a bit ad hoc

tpolecat: Ha, I was just looking at that. For operations that generate multiple events via a bulk update we might want to compare sets rather than lists.

tpolecat (approved):
"These observations are equivalent (from point of view X)" seems to be a common theme around here. Looks good. 👍🏻

PR #1459, opened 2024-10-25
remove old workflow state fields

tpolecat:
This removes `ObsStatus`, `ObsActiveStatus`, and the `forReview` field, all of which are subsumed by `ObservationWorkflowState`. Associated tests for these properties'

A followup PR will add an internal user state that allows the user to set the workflow state to `Inactive` and `Ready` when appropriate.

Explore should remove references to these things in preparation for the schema change.


modules/service/src/test/scala/lucuma/odb/graphql/query/observationValidations.scala
```diff
@@ -1,690 +0,0 @@
-// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
```
tpolecat, line 1: The `validations` field was removed and is now on `workflow.validationErrors`. The associated tests (in `observation_workflow`) include the tests here.

modules/service/src/test/scala/lucuma/odb/graphql/mutation/cloneObservation.scala
```diff
@@ -228,96 +222,96 @@ class cloneObservation extends OdbSuite {
 
   }
 
-  test("cloned observation should always be present/new/active") {
+  // test("cloned observation should always be present/new/active") {
```
toddburnside, line 225: Do we want to keep this code? If so, maybe comment on why it is commented out. Same for the commented out tests in the next few files.

toddburnside (approved):
LGTM. I'd almost approve it on the basis of this alone: 
image
Deleted lines could be even higher if you delete commented out tests. :)

modules/service/src/test/scala/lucuma/odb/graphql/mutation/cloneObservation.scala
tpolecat: Yeah I will go back over these this morning and see if they're still relevant.

PR #1464, opened 2024-10-29
Moves telescope config from step config science to the step itself.

swalker2m:
This is a fairly mechanical update to track https://github.com/gemini-hlsw/lucuma-core/pull/968 and update the API accordingly.  It moves the offset and guide state from the step config `Science` type to the step itself.  In other words, the telescope config applies to all step types now.


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -4549,7 +4548,7 @@ input StepConfigInput {
   """
   Science step creation option
   """
-  science: StepConfigScienceInput
+  science: Boolean
```
cquiroz, line 4551: Why boolean?

swalker2m: It marks that you want to create a science step.  Since there is no other state specific to a science step, there's nothing to configure but somehow you have to specify that you want a science step.

cquiroz (approved):
LGTM

PR #1465, opened 2024-10-30
Make targetEdit work like observationEdit

cquiroz:
`targetEdit` gets the same treatment as `observationEdit` supporting events for deletion of targets

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
@@ -266,6 +266,7 @@ object CalibrationsService extends CalibrationObservations {
           configs       = uniqueSci.map(_._2).diff(uniqueCalibs.map(_._2))
           addedOids    <- generateGMOSLSCalibrations(pid, configs, gnTgt, gsTgt)
           removedOids  <- removeUnnecessaryCalibrations(uniqueSci.map(_._2), calibs)
+          _            <- targetService.deleteOrphanCalibrationTargets(pid)
         yield (addedOids, removedOids)
 
       // Recalcula the target of a calibration observation


modules/service/src/main/scala/lucuma/odb/graphql/mapping/SubscriptionMapping.scala
```diff
@@ -182,8 +182,24 @@ trait SubscriptionMapping[F[_]] extends Predicates[F] {
...
+              ),
+              Filter(
+                Predicates.targetEdit.programId.eql(e.programId),
+                Query.mapSomeFields(child):
+                  case Select("value", a, c) =>
+                    Select("value", a,
+                      // This predicate needs to be down here
+                      Filter(
+                        if e.editType === EditType.DeletedCal
+                        then Predicates.target.id.isNull(true) // always false; Predicate.False doesn't work
```
swalker2m, line 197: Could you expand a bit on this? I'm easily confused.  

We are trying to support events for deletion of targets.  Those would be calibration targets because no others are deleted, correct?  Yet if we delete a calibration, we filter out the event here?

We've made it so target id is never null so this is always false but `Predicate.False` doesn't work?

cquiroz: TBH I copied this from obs events, if you recall this was a part that took quite a bit of effort by @tpolecat to get right.
the  edit type `DeletedCal` is a bit unfortunate as this are targets being deleted. I convinced myself it was ok  because targets are only deleted  when a target is deleted. I could create a separate `EditType` for ths particular case

swalker2m (approved):
Well, I can't say I understand how this works but I am convinced it matches observation deletion. :+1:

PR #1469, opened 2024-11-05
setObservationWorkflowState

tpolecat:
This adds a `setObservationWorkflowState` mutation, allowing the user to change the workflow state to one of the legal transitions as provided by the `workflow/validTransitions` field on Observations.


toddburnside (approved):
LGTM!

PR #1470, opened 2024-11-05
WIP: observing mode group

swalker2m:
This is a work-in-progress attempt to add observation grouping by observing mode in a way similar to the constraint set grouping.  I'd like feedback on whether this looks feasible.

Constraint set grouping is a much simpler case case because observing modes:

* are heterogeneous and live in separate tables (`t_gmos_north_long_slit` vs `t_gmos_south_long_slit`)
* have default values that are used when explicit values are not provided

In particular the last point is problematic because the default binning calculation is detailed.  Currently we compute the defaults in code, but in order to group observing modes regardless of whether the binning is explicitly provided or supplied by default, I need to get default binning into the long slit tables.  How to do that is the question. In this version the calculation is done in `plpgsql` but other alternatives might be:

* Carefully update all the mutations that touch relevant fields and explicitly redo the calculation and update via Scala code.
* Create a channel for updates to anything that would impact the default binning, respond to updates in Scala code.

What do we think best?


modules/service/src/main/resources/db/migration/V0927__grouping_modes.sql
```diff
@@ -0,0 +1,216 @@
...
+  CASE WHEN o.c_spec_signal_to_noise_at  IS NOT NULL THEN o.c_observation_id END AS c_spec_signal_to_noise_at_id,
+  CASE WHEN o.c_spec_wavelength_coverage IS NOT NULL THEN o.c_observation_id END AS c_spec_wavelength_coverage_id,
+  CASE WHEN o.c_spec_focal_plane_angle   IS NOT NULL THEN o.c_observation_id END AS c_spec_focal_plane_angle_id,
+  CASE WHEN o.c_observation_duration     IS NOT NULL THEN o.c_observation_id END AS c_observation_duration_id,
+  c.c_active_start::timestamp + (c.c_active_end::timestamp - c.c_active_start::timestamp) * 0.5 AS c_reference_time,
+  COALESCE(
+    CASE WHEN o.c_observing_mode_type = 'gmos_north_long_slit' THEN mode_gnls.c_mode_key END,
+    CASE WHEN o.c_observing_mode_type = 'gmos_south_long_slit' THEN mode_gsls.c_mode_key END,
+    NULL
+  ) AS c_mode_key
```
swalker2m, line 211: This `COALESCE` chunk would be the new bit on the view. It's probably dumb to add an explicit `NULL`.  Apart from that, the issue is that `c_mode_key` isn't on `t_observation` itself but rather in a view.  I might need to add it to `t_observation` so it can be indexed and then keep it up-to-date with triggers?  Advice?

swalker2m:
An example of the mode key:

```
lucuma-odb=# select c_observation_id, c_mode_key from v_observation where c_mode_key IS NOT NULL LIMIT 10;
 c_observation_id |                                               c_mode_key                                                
------------------+---------------------------------------------------------------------------------------------------------
 o-106            | gmos_south_long_slit:R150_G5326:LongSlit_2_00:695000:One:One:Slow:Low:FullFrame:0,20,-20:0,15,-15
 o-107            | gmos_north_long_slit:R400_G5305:LongSlit_1_50:700000:One:One:Slow:Low:FullFrame:0,8,-8:0,15,-15
 o-129            | gmos_north_long_slit:R831_G5302:LongSlit_1_00:890000:One:One:Slow:Low:FullFrame:0,5,-5:0,15,-15
 o-12a            | gmos_south_long_slit:R400_G5325:LongSlit_0_25:591000:One:One:Slow:Low:FullFrame:0,8,-8:0,15,-15
 o-12b            | gmos_south_long_slit:B480_G5327:LongSlit_1_00:555000:One:One:Slow:Low:FullFrame:0,5,-5:0,15,-15
 o-12e            | gmos_north_long_slit:R150_G5308:LongSlit_5_00:695000:One:One:Slow:Low:FullFrame:0,20,-20:0,15,-15
 o-14a            | gmos_north_long_slit:R831_G5302:LongSlit_1_00:600000:One:One:Slow:Low:FullFrame:0,5,-5:0,15,-15
 o-14d            | gmos_north_long_slit:B1200_G5301:LongSlit_5_00:600000:One:One:Slow:Low:FullFrame:0,5,-5:0,15,-15
 o-14e            | gmos_north_long_slit:R400_G5305:LongSlit_0_25:600000:One:One:Slow:Low:FullFrame:0,8,-8:0,15,-15
 o-151            | gmos_north_long_slit:R150_G5306:OG515:LongSlit_0_25:775000:One:One:Slow:Low:FullFrame:0,20,-20:0,15,-15
(10 rows)

```

modules/service/src/main/resources/db/migration/V0927__source_profile_summary.sql
```diff
@@ -0,0 +1,43 @@
...
+
+CREATE TYPE source_profile_summary AS (
+  c_profile_type e_source_profile_type,
+  c_fwhm         d_angle_µas
+);
+
+CREATE OR REPLACE FUNCTION extract_source_profile_summary(
+  src_profile  jsonb
+)
+RETURNS source_profile_summary AS $$
```
swalker2m, line 15: This needs testing but we would need to extract the source profile type (point, uniform, or gaussian) from the JSON blob along with the fwhm if gaussian.  This information is needed to compute the object size for the binning calculation.

modules/service/src/main/resources/db/migration/V0928__gmos_binning.sql
```diff
@@ -0,0 +1,224 @@
...
+--
+CREATE TYPE gmos_binning AS (
+  xbin d_tag,
+  ybin d_tag
+);
+
+-- Set the default binning.
+CREATE OR REPLACE PROCEDURE set_gmos_long_slit_default_binning(
+  oid d_observation_id
+) AS $$
```
swalker2m, line 142: We need to calculate the default binning in order to correct group observing modes.  Usually the binning is not explicitly set but we need to know what binning would be use.  As you can see, this calculation is rather involved. It makes me nervous to put so much logic into `plpgsql`.  Other options that occur to me:

* Carefully update all the mutations that touch relevant fields and explicitly redo the calculation and update via Scala code.
* Create a channel for updates to anything that would impact the default binning, respond to updates in Scala code.

What would be best?

modules/service/src/main/resources/db/migration/V0929__grouping_modes.sql
```diff
@@ -0,0 +1,216 @@
...
+  CASE WHEN o.c_spec_signal_to_noise_at  IS NOT NULL THEN o.c_observation_id END AS c_spec_signal_to_noise_at_id,
+  CASE WHEN o.c_spec_wavelength_coverage IS NOT NULL THEN o.c_observation_id END AS c_spec_wavelength_coverage_id,
+  CASE WHEN o.c_spec_focal_plane_angle   IS NOT NULL THEN o.c_observation_id END AS c_spec_focal_plane_angle_id,
+  CASE WHEN o.c_observation_duration     IS NOT NULL THEN o.c_observation_id END AS c_observation_duration_id,
+  c.c_active_start::timestamp + (c.c_active_end::timestamp - c.c_active_start::timestamp) * 0.5 AS c_reference_time,
+  COALESCE(
+    CASE WHEN o.c_observing_mode_type = 'gmos_north_long_slit' THEN mode_gnls.c_mode_key END,
+    CASE WHEN o.c_observing_mode_type = 'gmos_south_long_slit' THEN mode_gsls.c_mode_key END,
+    NULL
+  ) AS c_mode_key
```
swalker2m, line 211: This is the new bit in the view, but I suspect that doing this calculation in the view is not a good idea.  Instead, I think it probably would need to be done via triggers on updates to the `c_mode_key` in `t_gmos_north_long_slit` and `t_gmos_south_long_slit`.  That way we can index on this value for fast joins.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/QueryMapping.scala
```diff
@@ -413,6 +415,36 @@ trait QueryMapping[F[_]] extends Predicates[F] {
...
+        BooleanBinding("includeDeleted", rIncludeDeleted)
+      )) =>
+        Elab.transformChild: child =>
+          (programPredicate(rPid, rProp, rProg, Predicates.observingModeGroup.program), rWHERE, rLIMIT, rIncludeDeleted).parTupled.flatMap: (program, WHERE, LIMIT, includeDeleted) =>
+            val limit = LIMIT.foldLeft(ResultMapping.MaxLimit)(_ min _.value)
+            ResultMapping.selectResult(child, limit): q =>
+              FilterOrderByOffsetLimit(
+                pred = and(List(
+                  WHERE.getOrElse(True),
+                  program,
```
swalker2m, line 436: This predicate appears to be ignored.  If I add `program { id }` to the observing mode group query I can be told 

```
java.lang.Throwable: Expected single value for field 'id' of type Program at List(program, matches, observingModeGroup), found many
```

but it is unclear why.

modules/service/src/test/scala/lucuma/odb/graphql/query/observingModeGroup.scala
```diff
@@ -0,0 +1,357 @@
...
+                          "matches" : ${g.map { id => Json.obj("id" -> id.asJson) }.asJson }
+                        }
+                      }
+                    ]
+                  }
+                }
+              """.asRight
+            )
+
+  test("distinct observing modes edited to become same".ignore):
```
swalker2m, line 257: This test fails because there are multiple programs with observations using the same value for `c_mode_key`.  The `observations` that are included in the group span different programs despite the filter that limits them to a single program.

swalker2m:
I'll mark this as "ready" but recognizing that feedback on the binning calculation may necessitate redoing that chunk of it.

modules/service/src/main/resources/db/migration/V0927__source_profile_summary.sql
tpolecat: This is my fault … I was concerned that designing a table structure for source profiles would have driven me mad.

tpolecat (approved):
You have thought about all the issues that come to mind for me, and I'm also not really sure what the right approach is. But it seems to be correct and will probably perform fine once it's a triggered computation, so in the spirit of progress I think it's good to go. :+1:

PR #1471, opened 2024-11-11
Don't require same ITC target for acquisition and science

rpiaggio:
In an asterism, ITC may select different targets as the ones producing the shortest exposure time for acquisition and science.

This is actually OK, so I removed the assertion that the same target should be selected for both sequences.


cquiroz (approved):
LGTM

PR #1474, opened 2024-11-13
Use a different param for the amount of connections used by the calib…

cquiroz:
…ration service


swalker2m (approved):
Looks reasonable to me. :+1:

PR #1478, opened 2024-11-15
Adds an `ExecutionState` to `SequenceDigest`

swalker2m:
Renames `ObservationExecutionState` to `ExecutionState` and adds it to the `SequenceDigest`.  As part of the `SequenceDigest`, the result will be cached when possible.


tpolecat (approved):
Looks great, thanks! 👍🏻

PR #1482, opened 2024-11-21
batch queries for workspace computations

tpolecat:
This changes the `ObservationMapping` `effectField`s for `configurationRequests` and `workflow` to query returned observations at the same time rather than one at a time. It also combines some queries that were easily done together. It's kind of a big refactoring and it will be hard to review, sorry.

The goal here is to make the number of queries O(1) but there are still two places (`generatorParamsService.selectMany` and `itcService.selectAll`) where we query per-program and one (`generator.executionState`) where we query per-observation. So this is an improvement but there remains work to do.


modules/service/src/main/scala/lucuma/odb/graphql/mapping/ObservationMapping.scala
```diff
@@ -132,30 +134,61 @@ trait ObservationMapping[F[_]]
...
-  lazy val configurationRequestsQueryHandler: EffectHandler[F] = {
-    val readEnv: Env => Result[Unit] = _ => ().success
-
-    val calculate: (Program.Id, Observation.Id, Unit) => F[Result[List[ConfigurationRequest]]] =
-      (_, oid, _) =>
-        services.useTransactionally {
-          configurationService
-            .selectRequests(oid)
-        }
+  lazy val configurationRequestsQueryHandler: EffectHandler[F] = { pairs =>
```
toddburnside, line 137: I don't think I'd seen this method of creating an EffectHandler before.

modules/service/src/main/scala/lucuma/odb/service/ObservationService.scala
```diff
@@ -171,8 +171,8 @@ object ObservationService {
       .getOrElse(GenericConstraintViolationMessage(ex.message))
 
   // observation validation messages
-  val AsterismOutOfRangeMsg = "Asterism out of Call for Proposals limits."
-  val ExplicitBaseOutOfRangeMsg = "Explicit base out of Call for Proposals limits."
+  // val AsterismOutOfRangeMsg = "Asterism out of Call for Proposals limits."
```
toddburnside, line 174: Commented code

modules/service/src/main/scala/lucuma/odb/graphql/mapping/ObservationMapping.scala
```diff
@@ -132,30 +134,61 @@ trait ObservationMapping[F[_]]
...
+    val sequence: ResultT[F, List[((Program.Id, Observation.Id), Cursor, Context)]] =
+      ResultT.fromResult:
+        pairs.traverse: (query, cursor) =>
+          for {
+            p <- cursor.fieldAs[Program.Id]("programId")
+            o <- cursor.fieldAs[Observation.Id]("id")
+            c <- Query.childContext(cursor.context, query)
+          } yield ((p, o), cursor, c)
+
+    // Pass the pid+oid pars to configurationService.selectRequests to get the
```
toddburnside, line 150: Unimportant, but typo in `pars`

toddburnside (approved):
Wow, this was a lot of work. Thanks for taking it on! LGTM

modules/service/src/main/scala/lucuma/odb/graphql/mapping/ObservationMapping.scala
```diff
@@ -132,30 +134,61 @@ trait ObservationMapping[F[_]]
...
-        services.use { s =>
-          s.observationWorkflowService
-            .getWorkflow(oid, commitHash, itcClient, timeEstimateCalculator)
-        }
 
-    effectHandler(readEnv, calculate)
+  lazy val workflowQueryHandler: EffectHandler[F] = { pairs =>
+
+    // Here's the ordered sequence of stuff we need to deal with: 
+    // a pid+oid pair, the parent cursor, and the child context.
```
swalker2m, line 169: This one doesn't use a `pid`, it seems.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/ObservationMapping.scala
tpolecat: `EffectHandler` is a single abstract method (SAM) interface so you can define it using function syntax. Scala has supported this for a long time but it's not commonly used.

swalker2m: I only recently figured out that's also what's going on with `Matcher` implementations...

tpolecat:
(Added a test to make sure we actually get back different workflows for different observations.)

PR #1483, opened 2024-11-22
Migrate to new ITC API

rpiaggio:


cquiroz (approved):
LGTM

modules/service/src/test/scala/lucuma/odb/graphql/OdbSuite.scala
```diff
@@ -195,7 +195,12 @@ abstract class OdbSuite(debug: Boolean = false) extends CatsEffectSuite with Tes
         ).pure[IO]
 
       override def spectroscopy(input: SpectroscopyIntegrationTimeInput, useCache: Boolean): IO[IntegrationTimeResult] = {
-        IO.whenA(input.wavelength.some === lucuma.core.math.Wavelength.fromIntNanometers(666)) {
+        val signal = lucuma.core.math.Wavelength.fromIntNanometers(666).get
+        val wavelength = input.mode match
+          case lucuma.itc.client.InstrumentMode.GmosNorthSpectroscopy(w, _, _, _, _, _) => w
+          case lucuma.itc.client.InstrumentMode.GmosSouthSpectroscopy(w, _, _, _, _, _) => w
+          case _                                                                        => signal
+        IO.whenA(wavelength === signal) {
```
swalker2m, line 203: A test case is checking how we handle an exception when calling the ITC.  To trigger an exception artificially we set the central wavelength to a special value which the dummy ITC service finds and takes as a signal to raise an exception.  With the ITC input shakeup this got lost and was no longer triggering an exception.

PR #1485, opened 2024-11-25
Exclude deleted targets from v_generator_params

toddburnside:
Currently, deleted targets are still included when calculating times or sequences. https://app.shortcut.com/lucuma/story/3875/deleting-target-does-not-update-observation-duration-or-sequence
This excludes deleted targets.


swalker2m (approved):
Looks great to me. :+1:

PR #1497, opened 2024-12-02
Fix merging of instruments to build a CfpInfo

cquiroz:
odb is producing CfPs with a single instrument, we need to merrge all the rows (each with one instruments) grouped by cfp id

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/ObservationWorkflowService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/ObservationWorkflowService.scala
@@ -58,16 +58,16 @@ import Services.Syntax.*
 sealed trait ObservationWorkflowService[F[_]] {
 
   def getWorkflows(
-    oids: List[Observation.Id], 
-    commitHash: CommitHash, 
+    oids: List[Observation.Id],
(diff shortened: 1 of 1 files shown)


modules/service/src/main/scala/lucuma/odb/service/ObservationWorkflowService.scala
```diff
@@ -279,22 +281,23 @@ object ObservationWorkflowService {
...
+          session
             .stream(Statements.CfpInfos(enc))(nel, 1024)
+            // We get a row per instrument, so we need to group by CFP ID and fold the instruments
+            .fold(Map.empty[CallForProposals.Id, CfpInfo]) {
+              case (m, (cfp, oinst)) =>
+                m.updatedWith(cfp.cfpid) {
+                  case None    => cfp.addInstrument(oinst).some
+                  case Some(c) => c.addInstrument(oinst).some
+                }
+            }
```
swalker2m, line 293: The `.fold` could come on the `CompileOps` result of `compile` and eliminate the `.last.map(_.getOrElse(Map.empty))` bit at the end.

swalker2m (approved):
This seems right to me. :+1:

PR #1500, opened 2024-12-02
`addProgramUser` mutation

swalker2m:
This PR:

* tracks the addition of `UserProfile` and refactoring of `OrcidProfile` from core
* adds a `fallback` profile in `ProgramUser` which provides a location for name and email when not specified in the public ORCiD data
* separates the `ProgramUser` parts of the `ProgramService` into a new `ProgramUserService`
* adds a new `addProgramUser` mutation

The purpose of the mutation is to allow CoIs to be created and linked even before they have been extended / have accepted an invitation.  There is additional work to do here, I believe, to support the UI suggested in [ShortCut 3774](https://app.shortcut.com/lucuma/story/3774/allow-adding-investigators-directly).  In particular there is no clear link between invitations and program users. This will be addressed in a future PR but I think for now this will keep the invitation system working as is.


modules/service/src/main/scala/lucuma/odb/service/ProgramUserService.scala
```diff
@@ -0,0 +1,495 @@
...
+
+  def updateProgramUsers(
+    SET:   ProgramUserPropertiesInput,
+    which: AppliedFragment
+  )(using Transaction[F]): F[Result[List[(Program.Id, User.Id)]]]
+
+  def addProgramUser(
+    sso:   SsoGraphQlClient[F],
+    input: AddProgramUserInput
+  )(using NoTransaction[F]): F[Result[(Program.Id, User.Id)]]
```
swalker2m, line 77: Most of this was just moved from `ProgramService` in order to group everything `ProgramUser` related.  This mutation though, `addProgramUser`, is new.

modules/service/src/main/scala/lucuma/odb/service/ProgramUserService.scala
```diff
@@ -0,0 +1,495 @@
...
+        EXISTS (select c_duration from t_allocation where c_program_id = $program_id and c_ta_category=${lucuma.odb.util.Codecs.time_accounting_category} and c_duration > 'PT')
+      """.apply(programId, partner.timeAccountingCategory)
+
+    def existsUserAccess(
+      user:      User,
+      programId: Program.Id
+    ): Option[AppliedFragment] =
+      user.role match
+        case GuestRole                    => existsUserAsPi(programId, user.id).some
+        case StandardRole.Pi(_)           => (void"(" |+| existsUserAsPi(programId, user.id) |+| void" OR " |+| existsUserAsCoi(programId, user.id) |+| void")").some
```
swalker2m, line 397: I think perhaps we may want to say that a CoI only has access if they have accepted an invitation?  At any rate, at the moment they will have access immediately if added via `addProgramUser`.

modules/service/src/main/scala/lucuma/odb/service/UserInvitationService.scala
```diff
@@ -141,7 +141,7 @@ object UserInvitationService:
                   val xa = transaction
                   xa.savepoint.flatMap: sp =>
                     session
-                      .prepareR(ProgramService.Statements.LinkUser.command)
+                      .prepareR(ProgramUserService.Statements.LinkUser.command)
```
swalker2m, line 144: Here in the `UserInvitationService` when a user accepts an invitation, we attempt to "link" them to the program (i.e., add them as a `ProgramUser`).  If previously added explicitly via `addProgramUser` then the linking part will fail with a warning.  I think we'll want to handle this more smoothly.

tpolecat (approved):
This looks good. Can you add stories for the potential issues you identified in your comments? :+1:

swalker2m:
> Can you add stories for the potential issues you identified in your comments?

That's a good idea:

https://app.shortcut.com/lucuma/story/4141/handle-invitations-for-direct-program-users

https://app.shortcut.com/lucuma/story/4144/program-access-for-pre-acceptance-users

PR #1510, opened 2024-12-09
Unify the file attachments.

toddburnside:
Unifies the proposal and observation attachments into one table and one set of rest endpoints. This paves the way for adding custom SED attachments for targets.


modules/service/src/main/scala/lucuma/odb/graphql/AttachmentRoutes.scala
```diff
@@ -42,105 +43,112 @@ object ObsAttachmentRoutes {
...
 
-    given QueryParamDecoder[Tag] = QueryParamDecoder[String].map(s => Tag(s.toLowerCase))
+    extension[A](fe: F[Either[AttachmentException, A]])
+      def toResponse(fa: A => F[Response[F]]): F[Response[F]] =
+        fe.flatMap { _ match {
+          case Right(a) => fa(a)
+          case Left(e) => e.toErrorResponse
+        }
+        }
+
```
tpolecat, line 86: FYI you actually don't need the `match` here; you can use partial function syntax where a total function is expected; i.e., 

```scala
      def toResponse(fa: A => F[Response[F]]): F[Response[F]] =
        fe.flatMap {
          case Right(a) => fa(a)
          case Left(e) => e.toErrorResponse
        }
```

tpolecat (approved):
This was a big refactoring! Looks good. :+1:

modules/service/src/main/scala/lucuma/odb/graphql/AttachmentRoutes.scala
toddburnside: Thanks, I think my brain just got over full. 😆 I'll fix that and the formatting in another PR I'll have soon.

PR #1511, opened 2024-12-09
Use central wavelength for guide star stuff

toddburnside:
instead of the spectroscopy requirements wavelength.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/graphql/AttachmentRoutes.scala
+++ b/modules/service/src/main/scala/lucuma/odb/graphql/AttachmentRoutes.scala
@@ -78,10 +78,9 @@ object AttachmentRoutes {
 
     extension[A](fe: F[Either[AttachmentException, A]])
       def toResponse(fa: A => F[Response[F]]): F[Response[F]] =
-        fe.flatMap { _ match {
+        fe.flatMap {
           case Right(a) => fa(a)
           case Left(e) => e.toErrorResponse
-        }
         }
 
     def attachmentTypeFailure(s: String): ParseFailure =
--- a/modules/service/src/main/scala/lucuma/odb/service/GuideService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/GuideService.scala
@@ -185,16 +185,12 @@ object GuideService {
     programId:          Program.Id,
     constraints:        ConstraintSet,
     posAngleConstraint: PosAngleConstraint,
-    optWavelength:      Option[Wavelength],
     explicitBase:       Option[Coordinates],
     optObsTime:         Option[Timestamp],
     optObsDuration:     Option[TimeSpan],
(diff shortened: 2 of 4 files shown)


modules/service/src/main/scala/lucuma/odb/graphql/AttachmentRoutes.scala
```diff
@@ -78,11 +78,10 @@ object AttachmentRoutes {
 
     extension[A](fe: F[Either[AttachmentException, A]])
       def toResponse(fa: A => F[Response[F]]): F[Response[F]] =
-        fe.flatMap { _ match {
```
toddburnside, line 81: An unrelated minor improvement and formatting for a previous PR.

rpiaggio (approved):
LGTM

I wonder if by this point we are using the requirements wavelength for anything.

toddburnside:
> LGTM
> 
> I wonder if by this point we are using the requirements wavelength for anything.

Maybe only for filtering rows from the spectroscopy table?

modules/service/src/main/scala/lucuma/odb/service/GuideService.scala
```diff
@@ -962,7 +948,6 @@ object GuideService {
           obs.c_hour_angle_max,
           obs.c_pac_mode,
           obs.c_pac_angle,
-          obs.c_spec_wavelength,
```
cquiroz, line 965: Should we drop this column?

cquiroz (approved):
LGTM

modules/service/src/main/scala/lucuma/odb/service/GuideService.scala
toddburnside: We don't use it for anything, so it seems like we should. Is there a reason not to?

toddburnside: I have no idea why I originally added the `obs` table alias since there are no joins. I'll get rid of it.

cquiroz: Do we store there the requirement wavelength?

toddburnside: Yes

toddburnside: Oh, did you mean actually drop the column from the t_observation table? Not just the query. Yeah, that is the spectroscopy requirements wavelength value.

toddburnside: Raul did talk to Andy about dropping either the spec requirements `wavelength` or the `wavelengthAt` value, since they don't both seem necessary. Andy said "_we'll see_", or something to that effect.

PR #1513, opened 2024-12-12
workflow optimizations

tpolecat:
This adds a number of methods that bring the number of per-program/observation queries closer to a constant value, when doing a query for observations that includes the `ObservationWorkflow` struct. To wit:

- A new overload of `Generator#executionState` that allows passing of explicit `GeneratorParams` and `AsterismResults` if known, to avoid additional lookups.
- `GeneratorParamsService#selectMany` that does bulk selection over many programs.
- `ItcService#selectAll` that does bulk selection over many programs, similarly.

I'm assuming in general that access control has already been applied to any query that returns observations, so the effect computing the workflow doesn't need additional checks (in particular cross-program querying is simpler). Associated service methods demand `SuperUserAccess`.

Using these new service methods (`Generator#executionState` in particular) required a lot of refactoring in `ObservationWorkflowService`. It's a bit simpler now, primarily running in two phases with an effectful construction of `ObservationValidationInfo` followed by a mostly pure validation process thereof.


swalker2m (approved):
Looks great!  I feel guilty for not having done this before.

PR #1515, opened 2024-12-16
Switch adminer to adminerevo

hugo-vrijswijk:
Adminer is unmaintained and deprecated, but a fork is still active. See https://github.com/adminerevo/adminerevo


toddburnside (approved):
Seems like a good idea to me. 😄

PR #1517, opened 2024-12-16
batch execution cache check

tpolecat:
This is the last workflow optimization (for now). It adds `Generator#executionStates` (with an `s`) that performs a batch lookup for cached digests, which should speed things up in the typical case where the cache is up to date. It degrades to the sad linear "compute and cache every digest one by one" behavior in the worst case, but it does use pre-computed itc results and generator params, so this should be faster than before.


swalker2m (approved):
Looks great!  :+1:

PR #1524, opened 2024-12-19
CfP Creation Bug

swalker2m:
CfP semesters are capped at 1 year from the current year.  This is enforced by a Postgres `CHECK` on the semester domain. We were not checking for this though when creating a new CfP so using a semester far in advance would generate an Internal Server Error.


cquiroz (approved):
LGTM

PR #1528, opened 2024-12-20
Program user unique indices

swalker2m:
Per [ShortCut 3151](https://app.shortcut.com/lucuma/story/3151/allow-investigators-to-have-a-support-role), changes the uniqueness constraints on program user / user links such that:

* A user _cannot_ occupy the same role twice in the same program.
* A user _cannot_ be more than one of PI, CoI, CoI-RO in the same program.
* A user _can_ be one of Pi, CoI, CoI-RO and at the same time a support user.
* A user _can_ be both the primary and secondary support for a program.


swalker2m:
Note, this is an addendum to #1522.

swalker2m:
This will simply allow multiple `ProgramUser` entries to be associated with the same `User`.  Obviously that's annoying because all the attributes (names, email) are duplicated.  Another approach would allow one `ProgramUser` to have multiple roles.  That's more complex, but I could switch to that if necessary.

PR #1530, opened 2024-12-20
applicable observations

tpolecat:
This adds a top-level `configurationRequests` query, which is useful on its own I suppose, but mostly it's used by the new `applicableObservations` field on `ConfigurationRequest` (also added here).

@toddburnside does this do what you need?


toddburnside (approved):
Yes, I should be able to work with this. Thanks!
LGTM

PR #1533, opened 2024-12-23
specphot wavelength for s n measurement from average of science

cquiroz:
This is a bit of exotic requirement. For Spec Photometric calibrations we want to set the wavelength where to measure the signal to noise from the average of the science observations sharing the same configuration.

A tricky part is not only doing this during creation but also respond when the value of wavelength at changes across the science observations.


modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
```diff
@@ -148,39 +156,65 @@ object CalibrationsService extends CalibrationObservations {
...
-      private def toCalibConfig(all: List[(Observation.Id, ObservingMode)]): List[(Observation.Id, CalibrationConfigSubset)] =
+      private def toConfigForCalibration(all: List[(Observation.Id, Option[ItcInput], ObservingMode)]): List[(Observation.Id, Option[ItcInput], CalibrationConfigSubset)] =
         all.map(_.map(_.toConfigSubset))
 
-      private def uniqueConfiguration(all: List[(Observation.Id, ObservingMode)]): List[(Observation.Id, CalibrationConfigSubset)] =
-        toCalibConfig(all).distinctBy(_._2)
+      private def configAtWavelength(
+        calibConfigs: List[(Observation.Id, Option[ItcInput], CalibrationConfigSubset)]
+      ): Map[CalibrationConfigSubset, Option[Wavelength]] =
+        val atWavelengthPerConfig =
```
rpiaggio, line 186: Maybe it can help to have explicit types?

modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
```diff
@@ -253,26 +289,72 @@ object CalibrationsService extends CalibrationObservations {
...
+          }.void
+
       override def calibrationTargets(roles: List[CalibrationRole], referenceInstant: Instant)
         : F[List[(Target.Id, String, CalibrationRole, Coordinates)]] =
           session.execute(Statements.selectCalibrationTargets(roles))(roles)
             .map(targetCoordinates(referenceInstant))
 
       def recalculateCalibrations(pid: Program.Id, referenceInstant: Instant)(using Transaction[F]): F[(List[Observation.Id], List[Observation.Id])] =
         for
+          // Read calibration targets
```
rpiaggio, line 329: Thanks for the comments 🙏

rpiaggio (approved):
LGTM. I didn't follow the logic thoroughly but the new test passes.

PR #1539, opened 2025-01-02
subscription refactor

swalker2m:
I noticed a lot of duplication in the various `Topic` instances and since I need to add a new one, I thought I'd first see if I could extract all the common code.  Maybe there are less clumsy ways of doing it though?


cquiroz (approved):
LGTM

tpolecat (approved):
I couldn't figure out a great way to do this either, but this seems like an improvement. 👍🏻

PR #1540, opened 2025-01-03
Execution Event Events

swalker2m:
Adds a subscription for execution events, so that Explore can follow along with the execution.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/graphql/binding/ExecutionEventTypeBinding.scala
+++ b/modules/service/src/main/scala/lucuma/odb/graphql/binding/ExecutionEventTypeBinding.scala
@@ -1,0 +1,10 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.odb.graphql
+package binding
+
+import lucuma.odb.data.ExecutionEventType
+
+val ExecutionEventTypeBinding: Matcher[ExecutionEventType] =
+  enumeratedBinding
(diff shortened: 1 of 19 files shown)


modules/service/src/main/scala/lucuma/odb/graphql/binding/ExecutionEventTypeBinding.scala
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.odb.graphql
+package binding
+
+import lucuma.odb.data.ExecutionEventType
```
swalker2m, line 7: I think I should probably move this to `core`?

modules/service/src/main/scala/lucuma/odb/graphql/input/ExecutionEventAddedInput.scala
```diff
@@ -0,0 +1,32 @@
...
+import lucuma.core.model.Program
+import lucuma.core.model.Visit
+import lucuma.odb.data.ExecutionEventType
+import lucuma.odb.graphql.binding.*
+
+case class ExecutionEventAddedInput(
+  programId:     Option[Program.Id],
+  observationId: Option[Observation.Id],
+  visitId:       Option[Visit.Id],
+  eventType:     Option[WhereEq[ExecutionEventType]]
```
swalker2m, line 18: This is a bit more involved than the other subscription inputs, so that the subscription can filter on multiple event types.

modules/service/src/main/scala/lucuma/odb/graphql/binding/ExecutionEventTypeBinding.scala
tpolecat: Probably, maybe as a follow-on since it's such a pain to change core.

tpolecat (approved):
This looks good. :+1:

PR #1550, opened 2025-01-08
Add `centralWavelength` to GMOS steps

swalker2m:
Adds `centralWavelength` to JSON steps as requested in [ShortCut 4491](https://app.shortcut.com/lucuma/story/4491/compute-central-wavelength-for-each-step?vc_group_by=day&ct_workflow=all&cf_workflow=500000071).  

Central wavelength is taken from the grating config (if present) and otherwise from the filter (if present).  It will be `null` if neither are specified in the GMOS dynamic config.  There are two places where updates are required:

1. In the JSON codec for the GMOS dynamic config which is used when the sequence is generated
2. In the recorded steps which are mapped from database tables

To simplify the second item, I updated the GMOS dynamic config views to include a computed `c_central_wavelength` column.


cquiroz (approved):
LGTM

PR #1551, opened 2025-01-09
Drain body of attachment upload requests on validation errors.

toddburnside:
Unfortunately, whatever client uses in front of the ODB does not behave well if the body of the request has not been drained before the server returns a response. So, I make sure the body gets drained if there are validation errors.

This replaces #709. Instead of blindly always draining the body, this only drains it if validation failed and we haven't read the body yet. The `NetworkError` will still occur if the user tries to upload too large of a file, but Explore prevents that.

Of course, I can't be certain if this will solve the problem until it is actually deployed on Heroku... The problem does not manifest with a local instance of the ODB.


swalker2m (approved):
Sorry, for some reason I thought I'd already approved this.  Looks good. :+1:

PR #1552, opened 2025-01-09
observationsByWorkflowState

tpolecat:
Small PR that was hard to figure out.

This adds a top-level `observationsByWorkflowState` query that lets the user query observations using a normal `WHERE` plus an additional list of allowed workflow states. This couldn't be done by adding it to the existing `WhereObservation` struct because the state filtering needs to be done explicitly and can't appear in a nested and/or struct that includes SQL filters. This is potentially expensive and is limited to service users who we can trust (heh heh) to specify a filter for semester, instruments, etc., to limit the number of computed workflows.


swalker2m (approved):
Amazing. I think I would have given up. :+1:

PR #1559, opened 2025-01-15
Unify proposal title and program name

swalker2m:
There are two goals expressed in [ShortCut 2422](https://app.shortcut.com/lucuma/story/2422/unify-proposal-title-and-program-name):

* Remove the duplication between program `name` and proposal `title`
* Make it possible to add an abstract to all programs, even if they don't go through a TAC process

Accordingly I made these changes:

* Removed the proposal `title` in favor of the existing program `name`.  Whether a program has a proposal or not it can have a name.
* Removed the proposal `abstract` and added a program `description`.  This allows all programs to have a description, matches the `name` and `description` fields of `Group`, and removes the clashes-with-scala term `abstract`.

I expect the existing `Proposal` `title` and `abstract` to map to `Program` `name` and `description`.  Does this sound okay or do we want to keep `title` and `abstract` and just move them to `Program`.


toddburnside (approved):
Seems fine to me. It will, of course, break explore. I'll check with Andy to see how he wants the UI to work.

PR #1563, opened 2025-01-16
Fix bug on cache lookup

swalker2m:
This is a fix for a bug in which a time estimate query would:

* perform an execution digest cache lookup
* find a valid value
* ignore it and perform the calculation again
* proceed to update the cache
 
When the cache is updated, an observation edit event is now sent which in turn would prompt Explore to perform the query again ....

The problem boils down to significant whitespace and a missing `:` character.  I'm not sure why this compiled before.

We might want to add a comparison in the trigger function to ensure that the `OLD` and `NEW` records differ before sending an event as suggested.  If that had been done though, it would have masked this issue so I'm not sure.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/logic/TimeEstimateService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/logic/TimeEstimateService.scala
@@ -81,8 +81,8 @@ object TimeEstimateService:
     asterismResults: Option[AsterismResults],
     executionDigest: Option[ExecutionDigest]
   ):
-    def cachedFullTimeEstimate: Option[CategorizedTime] =
-      executionDigest.map(_.fullTimeEstimate)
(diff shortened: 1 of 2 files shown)


modules/service/src/main/scala/lucuma/odb/logic/TimeEstimateService.scala
```diff
@@ -116,7 +116,7 @@ object TimeEstimateService:
         OptionT.fromOption(m.get(oid)).flatMap: data =>
           OptionT
             .fromOption(data.cachedFullTimeEstimate)  // try the cache first
-            .orElse
+            .orElse:
```
swalker2m, line 119: This is the bug.  Adding the `:` makes it work as intended and requires fixing the `cachedFullTimeEstimate` value.

toddburnside (approved):
Good catch. Very hard to spot. LGTM

PR #1568, opened 2025-01-17
Bump versions/adapt code

rpiaggio:


swalker2m (approved):
Seems fine to me.  @tpolecat any reason to avoid `3.6.2`?

modules/sequence/src/main/scala/lucuma/odb/sequence/gmos/longslit/Science.scala
```diff
@@ -658,7 +658,7 @@ object Science:
           val blocksʹ     = blocks.updatedUnsafe(pos, block.add(n))
           ((blocksʹ, aix + 1, csʹ), atom)
         }
-        .cons1((blocks, 0, 0), atom0) // put the first atom back
+        .cons1((blocks, 0, TimeEstimateCalculator.Last.empty), atom0) // put the first atom back
```
toddburnside, line 661: Where did this change come from?

toddburnside (approved):
LGTM

modules/sequence/src/main/scala/lucuma/odb/sequence/gmos/longslit/Science.scala
rpiaggio: The type of that parameter so far is `TimeEstimateCalculator.Last` and Scala 3.6.2 (incorrectly) complains that the subsequent pattern match does not contemplate all cases.

The value of that parameter is ignored, I just set it to something that matched the type thus far to silence the warning.

PR #1573, opened 2025-01-21
SmartGCal support for Flamingos2

cquiroz:
This is for review I still need to do a core release before merging.
@swalker2m This is what I found was needed for SmartGCal but maybe I'm missing something else?


modules/smartgcal/src/main/scala/lucuma/odb/smartgcal/data/Flamingos2.scala
```diff
@@ -0,0 +1,124 @@
...
+import lucuma.core.util.TimeSpan
+import lucuma.odb.smartgcal.data.SmartGcalValue.LegacyInstrumentConfig
+import monocle.Focus
+import monocle.Lens
+import monocle.Optional
+import lucuma.core.model.sequence.f2.F2DynamicConfig
+
+object Flamingos2:
+
+  case class SearchKey(
```
cquiroz, line 24: This is taken from Gmos, I'm not sure if I need a separate `SearchKey` and `TableKey`

modules/service/src/main/resources/db/migration/V0943__smart_gcal_f2.sql
```diff
@@ -0,0 +1,34 @@
...
+  -- Primary key, which is also a foreign key into t_gcal for the matching GCal
+  -- configuration.  We don't really search this table by primary key but rather
+  -- by the instrument config (see Search Key below).  This key just links the
+  -- instrument row to the matching gcal row in t_gcal.
+  c_instrument       d_tag NOT NULL,
+  c_gcal_id          int4  NOT NULL,
+  PRIMARY KEY (c_instrument, c_gcal_id),
+
+  FOREIGN KEY (c_instrument, c_gcal_id) REFERENCES t_gcal(c_instrument, c_gcal_id),
+  CONSTRAINT check_is_gmos_south CHECK (c_instrument = 'Flamingos2'),
```
swalker2m, line 15: `check_is_f2`

modules/smartgcal/src/main/scala/lucuma/odb/smartgcal/data/Flamingos2.scala
swalker2m: I'm struggling to remember why I thought I needed distinct types for GMOS.  It may be that the search key once separated the wavelength.  I'm not sure.  If you don't see a need for the distinction by all means merge them.

swalker2m:
> maybe I'm missing something else?

We'll need to update the `lucuma.odb.logic.SmartGcalImplementation` class when it is time to do sequence generation, but this looks great so far.

toddburnside (approved):
Looks good, as far as I can tell.

PR #1574, opened 2025-01-21
ShortCut 4015: change program user role

swalker2m:
Updates the API to add a new mutation which [allows a program user role to be updated](https://app.shortcut.com/lucuma/story/4015/ability-to-update-collaborator-s-access-privileges?vc_group_by=day&ct_workflow=all&cf_workflow=500000071).  This could probably be folded into to `updateProgramUsers` but since the role is such an integral part of authorization this seemed much easier.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/ProgramUserService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/ProgramUserService.scala
@@ -162,7 +162,7 @@ object ProgramUserService:
 
         (for
           o  <- ResultT(insert)
-          id <- o.fold(ResultT.fromResult(OdbError.NotAuthorized(user.id, s"User ${user.id} is not authorized to perform this operation.".some).asFailure))(ResultT.pure)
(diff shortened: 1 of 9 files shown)


modules/service/src/main/scala/lucuma/odb/service/ProgramUserService.scala
```diff
@@ -161,6 +167,25 @@ object ProgramUserService:
...
+      override def changeProgramUserRole(
+        input: ChangeProgramUserRoleInput
+      )(using Transaction[F]): F[Result[ProgramUser.Id]] =
+        session
+          .prepare(Statements.SelectLinkData)
+          .flatMap(_.option(input.programUserId))
+          .flatMap:
+            case None                     =>
+              OdbError.InvalidProgramUser(input.programUserId, s"ProgramUser ${input.programUserId} was not found.".some).asFailureF
+            case Some((pid, prevRole, _)) =>
```
swalker2m, line 179: Would it be worth the trouble to compare the previous and new roles and return a no-op warning if they are the same?

tpolecat (approved):
Looks good :+1:

PR #1581, opened 2025-01-22
ShortCut 3347: Validate partners on submission

swalker2m:
Checks that there is an investigator associated with each partner from which time is requested.


swalker2m:
Since this is a new requirement on submission, a lot of test cases needed to be updated to add matching investigators.

cquiroz (approved):
LGTM

PR #1583, opened 2025-01-23
test for sc-3485

tpolecat:
This updates the `TestSsoClient` used by `OdbSuite` to actually generate and exchange JWTs so we can test that this mechanism is working as expected ... and it is, which is annoying because I expected it to fail somehow. The added test demonstrates that the email address is correctly pulled from the JWT and saved in the database, so the problem is elsewhere. In any case this improves the test setup so we might as well keep it.


swalker2m (approved):
:+1:

PR #1587, opened 2025-01-24
sc-4577 eng workflow

tpolecat:
This changes the workflow for engineering programs such that they never generate validation errors, and thus always start in out as `Defined`. This may or may not be entirely correct. 

The logic here is starting to seem OT-like in its weirdness so we may want to refactor once the dust settles. I'm thinking maybe we also need to change the logic for other types of programs.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/ObservationWorkflowService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/ObservationWorkflowService.scala
@@ -14,6 +14,7 @@ import lucuma.core.enums.ExecutionState
 import lucuma.core.enums.Instrument
 import lucuma.core.enums.ObservationValidationCode
 import lucuma.core.enums.ObservationWorkflowState
+import lucuma.core.enums.ProgramType
 import lucuma.core.enums.ScienceBand
 import lucuma.core.enums.Site
 import lucuma.core.math.Coordinates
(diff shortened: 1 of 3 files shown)


modules/service/src/main/scala/lucuma/odb/service/ObservationWorkflowService.scala
```diff
@@ -352,7 +354,7 @@ object ObservationWorkflowService {
               case Inactive   => List(executionState.getOrElse(validationStatus))
               case Undefined  => List(Inactive)
               case Unapproved => List(Inactive)
-              case Defined    => List(Inactive) ++ Option.when(isAccepted)(Ready)
+              case Defined    => List(Inactive) ++ Option.when(isAccepted || info.tpe === ProgramType.Engineering)(Ready)
```
swalker2m, line 357: What's true for `Engineering` is probably true for `Calibration` and all the others?  `info.tpe =!= ProgramType.Science`

swalker2m (approved):
Looks good.  :+1:

modules/service/src/main/scala/lucuma/odb/service/ObservationWorkflowService.scala
tpolecat: That is sensible, I'll make that change.

PR #1588, opened 2025-01-24
Generalize populate-db-from-dev to work for staging and production

toddburnside:
The script for populating the local DB from odb dev worked. However, the one for staging did not work, and the one for production did not exist. This generalizes a single script to work for all 3.


cquiroz (approved):
LGTM

PR #1589, opened 2025-01-24
ShortCut 4602: Fix import dependency issue

swalker2m:
Adds a FK reference from `t_program(c_program_type)` to `t_program_type(c_type)`.  In addition to being the right thing to do (I think), my theory is that this will cause the postgres dependency computation to order `t_program_type` imports before `t_program` imports and thereby fix the computation of program references when restoring from a backup.


cquiroz (approved):
LGTM

toddburnside (approved):
Sounds like a good theory. LGTM.

PR #1591, opened 2025-01-28
Use graphql-inspector to detect breaking schema changes

cquiroz:
I'm using this tool to fail PRs that produce breaking schema changes. You can still override the behavior, adding a label `expected-breaking-change`

If it is too obtrusive we could make the action not fail breaking PRs.

You can see how it looks on ITC:
https://github.com/gemini-hlsw/lucuma-itc/pull/883/files#diff-bfc5dcd05e7282fc07897a390a0829a5c14bcb76898cfc1ddaf231545bc9fea4L10246

More info at
https://the-guild.dev/graphql/inspector/docs/products/action

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,6 +91,7 @@ jobs:
         with:
           schema: 'main:modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql'
           approve-label: expected-breaking-change
+          fail-on-breaking: false
 
       - name: Aggregate coverage reports
         run: sbt -v -J-Xmx6g '++ ${{ matrix.scala }}' coverageReport coverageAggregate
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,7 +91,6 @@ jobs:
         with:
           schema: 'main:modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql'
           approve-label: expected-breaking-change
-          fail-on-breaking: false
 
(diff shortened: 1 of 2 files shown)


swalker2m:
This looks interesting and useful, particularly when we're further along.  At this point, almost every PR has a breaking change though. I vote for `fail-on-breaking: false` for now, but adding a label every time wouldn't be so bad I suppose.

cquiroz:
Sure, lets set the flag, it is still useful for example for explore, to have a clear list of break ing changes

tpolecat:
I always lean toward knowing when things break so I vote for requiring the label. Fine either way though.

toddburnside:
I think we're at the point, or getting close to it, where we should try to deprecated fields before removing them. We now seem to have some "production" users, and a breaking change will mean explore will break for them until they update it even if we promote the odb and explore in synch. Requiring the label would at least make us acknowledge when we need to make breaking changes.

swalker2m:
> I think we're at the point, or getting close to it, where we should try to deprecated fields before removing them

Dragging around the deprecated implementation is not going to exactly speed us up so I'm not inclined to do it until it is really necessary.  I don't mind labeling every PR as a breaking change if that's what people prefer though.

toddburnside:
Looking back at the commit history, it seems most of our recent commits to the schema are non-breaking additions. However, we do have some breaking changes. As a part time API developer, I'm fine with deferring to the full time API developers on this. We can wait until we actually start having problems and then worry about breaking changes.

swalker2m (approved):
Looks great, thanks for taking the time to add this. :+1:

PR #1592, opened 2025-01-29
Readonly COIs cannot access accepted program

toddburnside:
The PR makes the tests created in #1575 fail. I think the program needs to have a defined observation that will create a ConfigurationRequest upon proposal acceptance.


swalker2m:
Thanks for pinpointing the problem @toddburnside.  WIth that I found that the issue was in a query to `AsterismService.getAsterism` requiring write access to the program:

* `ConfigurationMapping` -> `referenceCoordinates`
* `GuideService.getObjectTracking` -> `getAsterism`
* `AsterismService.getAsterism`

We haven't been distinguishing read/write access but it seems like it is time that we start.  This may be a ham-fisted attempt to do so, I'm not sure.  My concern is that we could be doing some operation that intuitively by itself only requires read access but perhaps were using it to filter results for a later unchecked update.

Also, I haven't tried to update any logic in `AttachmentFileService`, but it might be that some of the operations suffice with read access.

tpolecat (approved):
Yeah this has grown organically and there's no central place to look if you want to know what the rules are, so I think we need to pull this logic into someplace central eventually. But this looks good for now. 👍🏻

toddburnside (commented):
I can't approve because I opened the PR, but it LGTM. Thanks for taking care of this!

PR #1593, opened 2025-01-29
Update core to get the new F2 FPU tags

cquiroz:
The schema was already correct but now it matches the tags on the enumeration.


tpolecat (approved):

swalker2m:
What happens with the F2 smart gcal table, since it has a FK reference to `t_f2_fpu`?

cquiroz:
I missed changing the spectroscopy config options table, this was added on the last commit

cquiroz:
@swalker2m I just realized we didn't import anything for F2 smart gcal, I think it maybe better to add that migration in another PR

PR #1597, opened 2025-01-31
Shortcut 1957: ExposureTimeMode

swalker2m:
Updates to allow setting `ExposureTimeMode` as either signal-to-noise or else time-and-count.  Depends on [matching ITC updates](https://github.com/gemini-hlsw/lucuma-itc/pull/885).


tpolecat (approved):
Wow, this would have taken me months. It touches a lot of stuff! Thanks for all the tests. 👍🏻

modules/service/src/main/scala/lucuma/odb/graphql/input/SpectroscopyScienceRequirementsInput.scala
```diff
@@ -9,34 +9,31 @@ import eu.timepit.refined.types.numeric.PosInt
...
   focalPlaneAngle:    Nullable[Angle],
   capability:         Nullable[SpectroscopyCapabilities]
 )
 
-object SpectroscopyScienceRequirementsInput {
+object SpectroscopyScienceRequirementsInput:
 
   val Default: SpectroscopyScienceRequirementsInput =
     SpectroscopyScienceRequirementsInput(
       wavelength         = Nullable.Null,
```
cquiroz, line 34: Did we discuss that perhaps here the defaults should be `Absent`?

swalker2m: Yeah, we might want to circle back and change that.

PR #1602, opened 2025-02-03
ShortCut 4602: Drop `t_program_type`

swalker2m:
Well, it *seemed* like a good idea at the time ...


tpolecat (approved):
Yeesh. Thanks for figuring this out.

PR #1606, opened 2025-02-06
attempt to fix backups

tpolecat:
Output from `pg_dump` can't be restored due to unqualified references in some of our stored procedures. This updates a few such cases and adds a script that [destructively] tests that the local database can be recreated, backed up, and restored. The end goal here is to allow us to use `heroku pg:restore`.

This shouldn't have any visible effect on the ODB but it will allow me to continue testing.


swalker2m (approved):
You're doing the lord's work.  :+1:

PR #1607, opened 2025-02-06
Allow staff users to call `observationsByWorkflowState`

tpolecat:
This needs to be called from the Navigate and Observe *clients* so I can't require service access here. Should be fine.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/graphql/mapping/QueryMapping.scala
+++ b/modules/service/src/main/scala/lucuma/odb/graphql/mapping/QueryMapping.scala
@@ -174,7 +174,7 @@ trait QueryMapping[F[_]] extends Predicates[F] {
 
         // I can't believe this works.
         services.useNonTransactionally:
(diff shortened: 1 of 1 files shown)


modules/service/src/main/scala/lucuma/odb/graphql/mapping/QueryMapping.scala
```diff
@@ -174,7 +174,7 @@ trait QueryMapping[F[_]] extends Predicates[F] {
 
         // I can't believe this works.
         services.useNonTransactionally:
-          requireServiceAccess: // N.B. this is only for services, at least for now
+          requireStaffAccess: // N.B. this is only for services, at least for now
```
jluhrs, line 177: Update the comment, maybe?

tpolecat: yep

jluhrs (approved):
LGTM

swalker2m (approved):
:+1:

PR #1609, opened 2025-02-11
ShortCut 3897: data sharing

swalker2m:
Adds the "data sharing" idea from [ShortCut 3897](https://app.shortcut.com/lucuma/story/3897/allow-pi-to-elect-who-can-download-proprietary-data), including:

  * the `External` `ProgramUserRole` concept
  * the `hasDataAccess` property of `ProgramUser`
  * a GOA user configuration for establishing the GOA user
  * a `goaDataDownloadAccess` query for the GOA user

The query is a bit odd for a couple of reasons

1. It is limited to one particular user, the GOA user, which must be configured in the application properties.  I was uncomfortable with granting `service` user access to code outside of our codebase and yet the query cannot be open to just anybody.
2. It could almost be implemented by a simple query to `programUsers` with an appropriate `WHERE` clause.  It needs to run as a super user though to have access to all the programs and program users, even external ones.

See the notes in the [corresponding update to core](https://github.com/gemini-hlsw/lucuma-core/pull/1027) for more details.


modules/service/src/main/scala/lucuma/odb/service/UserInvitationService.scala
```diff
@@ -55,14 +55,23 @@ object UserInvitationService:
...
+          case _                        => ("collaborate", "on")
+
         val subject: NonEmptyString = NonEmptyString.unsafeFrom(
-          s"Invitation to collaborate from ${user.displayName}"
+          s"Invitation to $action from ${user.displayName}"
         )
 
         val textMessage: NonEmptyString = NonEmptyString.unsafeFrom(
-          s"""${user.displayName} has invited you to collaborate on a Gemini proposal. To accept this invitation go to ${emailConfig.exploreUrl} to log in, then from the upper right menu select "Redeem Invitations" and enter the following key:
+          s"""${user.displayName} has invited you to $action $preposition a Gemini proposal. To accept this invitation go to ${emailConfig.exploreUrl} to log in, then from the upper right menu select "Redeem Invitations" and enter the following key:
```
swalker2m, line 74: What happens in Explore after a user redeems an invitation?  In the case of an external user, they won't be able to see the program so they cannot be redirected to it.

toddburnside: > What happens in Explore after a user redeems an invitation? In the case of an external user, they won't be able to see the program so they cannot be redirected to it.

I'd have to check and make sure, but I'm pretty sure that they get directed to that program. If they don't actually have access to that program the program they would get the "Manage Programs" dialog instead.

tpolecat (approved):
Very nice, this looks great :+1:

toddburnside (approved):
LGTM. I'll check with Andy and see where he wants this table, etc.

PR #1610, opened 2025-02-12
ShortCut 3897: allow multiple GOA users and staff

swalker2m:
A small update to #1609, opening up the `goaDataDownloadAccess` query to multiple GOA users (according to configuration) and to users with `Staff` access or better.  The intent is to simplify maintenance and testing.


tpolecat (approved):
:+1:

PR #1612, opened 2025-02-12
Test Case for Subscription Event Bug

swalker2m:
There's a report that a simple update to S/N requirements generates up to 13 observation edit events.  This PR will attempt to find the issue.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/graphql/mapping/SubscriptionMapping.scala
+++ b/modules/service/src/main/scala/lucuma/odb/graphql/mapping/SubscriptionMapping.scala
@@ -5,8 +5,11 @@ package lucuma.odb.graphql
 
 package mapping
 
+import cats.Eq
 import cats.data.Nested
+import cats.effect.kernel.Temporal
 import cats.syntax.all.*
+import fs2.concurrent.Topic
 import fs2.Stream
 import grackle.Env
 import grackle.Query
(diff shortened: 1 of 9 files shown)


tpolecat (approved):

tpolecat (approved):

modules/service/src/main/scala/lucuma/odb/graphql/mapping/SubscriptionMapping.scala
```diff
@@ -269,3 +276,18 @@ trait SubscriptionMapping[F[_]] extends Predicates[F] {
...
+  extension [F[_]: Temporal, E: Eq](topic: Topic[F, E])
+    def groupingSubscribe(
+      maxQueued: Int          = MaxQueued,
+      chunkSize: Int          = GroupingChunkSize,
+      timeout: FiniteDuration = GroupingTimeout
+    ): Stream[F, E] =
+      topic
+        .subscribe(maxQueued)
+        .groupWithin(chunkSize, timeout)
+        .flatMap(Stream.chunk(_).changes)
```
swalker2m, line 293: This may not be a great idea after all. By the time the event is fired, the expected subscription query can produce different results than it would have if it had been generated immediately.  For example, when a new group is created without specifying a parent index, it defaults to index 0.  It may be subsequently moved.  Then the group creation event fires after the `GroupingTimeout` expires.  At that point, the index is no longer 0. So you can see a `CREATED` event with parent index 1 followed by an `UPDATED` event also at 1 when you expected `CREATED` at 0 followed up `UPDATED` showing 1.
```
{
  "groupEdit" : {
    "editType" : "CREATED",
    "value" : {
      "id" : "g-10c",
      "parentId" : "g-10b",
      "parentIndex" : 1
    }
  }
},
{
  "groupEdit" : {
    "editType" : "UPDATED",
    "value" : {
      "id" : "g-10c",
      "parentId" : "g-10b",
      "parentIndex" : 1
    }
  }
}
```

I don't know whether this is a problem for production, but it definitely makes tests unpredictable.

PR #1615, opened 2025-02-13
SC-4596 Part 1

tpolecat:
https://app.shortcut.com/lucuma/story/4596/api-should-prevent-editing-of-observations-for-which-execution-has-started

This has evolved into a bit of a refactor of access control code, which is in yet another `Mapping` mixin because it needs access to path-dependent types from Grackle :-\

The main pattern is

```scala
ResultT(selectForUpdate(updateWidgetsInput)).flatMap: checkedUpdate =>
  ResultT(widgetService.upadateWidgets(checkedUpdate))
```

where `checkedUpdate` is an `AccessControl.Checked` that contains `SET: WidgetPropertiesInput.Edit` and either an `AppliedFragment`, a `Widget.Id`, a `List[Widget.Id]`, or `Empty`, representing the requested update and ids for which it's approved. 

`AccessControl.Checked` can only be constructed by access control code, or by calling `AccessControl.unchecked` as the superuser (this allows robot code like the cal service to call service methods). The `Result` returned by `selectForUpdate` contains warnings for any ids that were discarded due to business logic like illegal observation workflow state.


modules/service/src/main/scala/lucuma/odb/graphql/mapping/MutationMapping.scala
```diff
@@ -750,24 +774,47 @@ trait MutationMapping[F[_]] extends Predicates[F] {
...
+          SET.scienceBand.isDefined         ||
+          SET.posAngleConstraint.isDefined  ||
+          SET.targetEnvironment.isDefined   ||
+          SET.constraintSet.isDefined       ||
+          SET.timingWindows.isDefined       ||
+          SET.attachments.isDefined         ||
+          SET.scienceRequirements.isDefined ||
+          SET.observingMode.isDefined       ||
+          SET.existence.isDefined           ||
+          SET.observerNotes.isDefined
```
swalker2m, line 792: This idea seems good to me.  I wonder if we should move this into the `ObservationPropertiesInput.Edit` itself to help keep up with it as stuff gets added.

swalker2m (dismissed):
This problem seems hard and I'm pretty sure I would have made a huge mess of it, but this looks great!  :+1:

modules/service/src/main/scala/lucuma/odb/graphql/mapping/MutationMapping.scala
tpolecat: I had it there at first actually, but I wasn't sure it belonged there so I moved it. I may move it back.

modules/service/src/test/scala/lucuma/odb/graphql/issue/shortcut/4596.scala
```diff
@@ -0,0 +1,512 @@
...
+            {
+              "updateAsterisms": {
+                  "observations": []
+              }
+            }
+          """
+        )
+      )
+ 
+  test(s"Ongoing observations should not allow their asterism's targets to be edited".ignore):
```
tpolecat, line 367: I will do this one in a follow PR.

tpolecat:
I regret to inform you that this PR is ready for review.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/AccessControl.scala
```diff
@@ -0,0 +1,281 @@
...
+      Some(Predicates.observation.id.in(oids)),
+      includeCalibrations,
+      allowedStates
+    )
+
+  /**
+   * Given an operation that defines a set of observations and a proposed edit, select and filter this
+   * set based on access control policies and return a checked edit that is valid for execution.
+   */
+  def selectForUpdate(
```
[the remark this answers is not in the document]
tpolecat: Yes, I have implemented this for targets in a followup but it's not quite ready yet. It creates a map from targets to related observations, computes the workflows, and filters the target ids.

PR #1616, opened 2025-02-14
ShortCut 3202: specphot science band

swalker2m:
This task arose from [a comment in ShortCut 3202](https://app.shortcut.com/lucuma/story/3202/support-planned-time-estimates-by-band#activity-4670), though not directly related to time estimates per se. 

> Would it be possible for the spectrophotometric standards to inherit the band of their associated science observations? If a specphot standard is associated with multiple science observations and those science observations have different bands then inherit the highest band (under the assumption that we are more likely to execute the higher band observations).

I updated the calibrations service to do this, in theory.  It adds the science band to the calibration observation properties to set (along with the average wavelength of the science observations).

I think I'll leave it here until Carlos is back and can take a look.  I didn't find any relevant tests for the `CalibrationService` and I'm not sure what we do for testing.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
@@ -100,6 +100,11 @@ object CalibrationsService extends CalibrationObservations {
     def map[B](f: A => B): ObsExtract[B] =
       copy(data = f(data))
 
+  case class CalObsProps(
+    wavelengthAt: Option[Wavelength],
+    band:         Option[ScienceBand]
+  )
+
   private def targetCoordinates(when: Instant)(
     rows: List[(Target.Id, String, CalibrationRole, RightAscension, Declination, Epoch, Option[Long], Option[Long], Option[RadialVelocity], Option[Parallax])]
   ): List[(Target.Id, String, CalibrationRole, Coordinates)] =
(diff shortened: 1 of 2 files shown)


modules/service/src/main/scala/lucuma/odb/service/CalibrationsService.scala
```diff
@@ -318,11 +339,16 @@ object CalibrationsService extends CalibrationObservations {
...
               // Important: Only update the obs that need it or it will produce a cascade of infinite updates
               // TODO: This could be slightly optimized by grouping obs per configuration and updating in batches
-              sql"select $observation_id where c_calibration_role = $calibration_role and c_spec_signal_to_noise_at <> $wavelength_pm".apply(oid, role, w)
+              sql"""
+                SELECT $observation_id
+                  FROM t_observation
+                 WHERE c_observation_id = $observation_id
+                   AND c_calibration_role = $calibration_role
+                   AND ("""(oid, oid, role) |+| needsUpdate |+| void")"
+//              sql"select $observation_id where c_calibration_role = $calibration_role and c_spec_signal_to_noise_at <> $wavelength_pm".apply(oid, role, w)
```
swalker2m, line 351: I reworked this query, but I may be misunderstanding the original intent so I'm particularly unsure about my update here.

cquiroz (approved):
LGTM

PR #1618, opened 2025-02-17
Only notify if something changes

swalker2m:
Updates (most of) our functions that execute `pg_notify` to elide the notification if nothing has actually changed.  I'm not convinced this will make much of a difference, but it probably also doesn't hurt?

### DIFF
--- a/modules/service/src/main/resources/db/migration/V0952__filter_edit_events.sql
+++ b/modules/service/src/main/resources/db/migration/V0952__filter_edit_events.sql
@@ -1,0 +1,149 @@
+-- Update event notify to first compare values
+
+CREATE OR REPLACE FUNCTION ch_observation_edit_associated_table_update()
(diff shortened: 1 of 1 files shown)


tpolecat (approved):
This looks good, we should have been doing this all along. 👍🏻

rpiaggio:
Does this mean that the same subscription but with slightly different queries can produce a different number of events?

swalker2m:
> Does this mean that the same subscription but with slightly different queries can produce a different number of events?

I don't anticipate that would happen in general, since a query doesn't modify anything.  When we do a mutation that invalidates caches, the subscription query can cause the cached value to be recomputed which will generate another event. We could put those in a different channel but we decided it would be easier for Explore not to have to listen to a different thing.

This PR just avoids sending an event if the "updated" values in a row are no different than the original values.

PR #1619, opened 2025-02-18
Don't advance ids for opposite sequence type

swalker2m:
Makes an adjustment to sequence generation to avoid changing the generated ids of the opposite sequence type while executing.  For example, now while doing acquisition the future science step IDs shouldn't change and vice versa.


rpiaggio (approved):
LGTM

swalker2m:
Added a bug fix for a problem in which the first step id in an atom would be repeated.

PR #1620, opened 2025-02-21
fix decoding error

tpolecat:
This fixes an error with `observationsByWorkflowState` that I couldn't easily reproduce locally but it happens with a copy of the dev database. In the interest of unblocking @rpiaggio I'd like to merge this and will follow up with a test that runs locally.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/ConfigurationService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/ConfigurationService.scala
@@ -249,18 +249,18 @@ object ConfigurationService {
           crs <- hc.downFields("configurationRequests", "matches").as[List[ConfigurationRequest]]
         yield (id, crs)
 
-      private given Decoder[((Program.Id, Observation.Id), Configuration)] = hc =>
+      private given Decoder[((Program.Id, Observation.Id), Option[Configuration])] = hc =>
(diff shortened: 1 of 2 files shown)


modules/service/src/main/scala/lucuma/odb/service/ConfigurationService.scala
```diff
@@ -249,18 +249,18 @@ object ConfigurationService {
           crs <- hc.downFields("configurationRequests", "matches").as[List[ConfigurationRequest]]
         yield (id, crs)
 
-      private given Decoder[((Program.Id, Observation.Id), Configuration)] = hc =>
+      private given Decoder[((Program.Id, Observation.Id), Option[Configuration])] = hc =>
         for
           pid <- hc.downFields("program", "id").as[Program.Id]
           oid <- hc.downField("id").as[Observation.Id]
-          cfg <- hc.downField("configuration").as[Configuration]
+          cfg  = hc.downField("configuration").as[Configuration].toOption
```
tpolecat, line 256: The issue here is that some observations don't have a valid configuration and I need to filter these out.

swalker2m (approved):
Makes sense.  :+1:

rpiaggio:
Thank you!

PR #1621, opened 2025-02-21
ShortCut-4745: active period

swalker2m:
Adds a program-specific active period, initialized from the CfP (if applicable).  When the CfP active period changes, programs that use the CfP and which had a matching active period are also updated to match the CfP's new active period.

Adds filtering on the active period to `WhereProgram`.


swalker2m:
I started to add a simple `isActive` filter to `WhereProgram` but stopped when I realized that it needs to be site specific. The way I wanted to do this was to make a program view which compared the database current timestamp to the active period start/stop range but this needs to be done at 2 PM local time for each site.  I could add two columns to the view, one for each site and then add an `isActiveAt: Site` flag if it seems like it would be worth it?  I think this is mostly meant for Observe though?

rpiaggio:
Right, this is for observe and navigate to load the list of observations that can be ... well, observed.

modules/service/src/main/resources/db/migration/V0953__program_active_period.sql
```diff
@@ -0,0 +1,71 @@
+-- Add active start / end.  The defaults seem fairly random but correspond to
+-- the limiting dates for our LST calculation.
+
+ALTER TABLE t_program
+  ADD COLUMN c_active_start date NOT NULL DEFAULT '1901-01-01'::date,
+  ADD COLUMN c_active_end   date NOT NULL DEFAULT '2099-12-31'::date,
```
rpiaggio, line 6: What happens if the ODB makes it to the next century?

rpiaggio (approved):
LGTM

modules/service/src/main/resources/db/migration/V0953__program_active_period.sql
swalker2m: There's definitely a Y2100 problem here.  But it won't be our problem.

PR #1623, opened 2025-02-26
ShortCut 3202: Select the highest priority (lowest ordinal) band.

swalker2m:
I made a mistake in #1616 in which I selected the `max` `ScienceBand` which turns out to be the lowest priority band since the `Order` is done starting with `Band 1` and continuing through `Band 4`.  We might want to reverse the `Order`.


swalker2m:
@cquiroz I'm not sure how we test the calibrations stuff

toddburnside (approved):
LGTM

tpolecat (approved):
> We might want to reverse the Order.

I can't decide if reversing the order would be less confusing or more confusing. In any case this is fine :+1:

PR #1624, opened 2025-02-27
SC-4596 part two

tpolecat:
This adds an `AccessControl` implementation for target updates.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/TargetService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/TargetService.scala
@@ -35,6 +35,7 @@ import lucuma.odb.graphql.input.CloneTargetInput
 import lucuma.odb.graphql.input.CreateTargetInput
 import lucuma.odb.graphql.input.SiderealInput
 import lucuma.odb.graphql.input.TargetPropertiesInput
+import lucuma.odb.graphql.mapping.AccessControl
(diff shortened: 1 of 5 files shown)


modules/service/src/main/scala/lucuma/odb/service/TargetService.scala
```diff
@@ -172,7 +171,7 @@ object TargetService {
         val stmt = Statements.cloneTarget(pid, targetId, user)
         session.prepareR(stmt.fragment.query(target_id)).use(_.option(stmt.argument))
 
-      def cloneTargetImpl(input: CloneTargetInput)(using Transaction[F]): F[CloneTargetResponse] =
+      private def cloneTargetImpl(input: CloneTargetInput)(using Transaction[F]): F[CloneTargetResponse] =
```
tpolecat, line 174: Scala 2 knows that non-interface methods on anonymous classes are implicitly private and warns if they're unused. Scala 3 isn't that smart and you can end up with unreachable code unless you mark everything private.

tpolecat: I talked to Som Snytt about this and he's looking into it.

tpolecat:
This is ready for review now. We regret the inconvenience.

swalker2m (approved):
Looks excellent.  :+1:

PR #1625, opened 2025-02-28
ShortCut 4839: dataset events

swalker2m:
(Also ShortCut 3346.). Adds dataset edit events.


modules/service/src/main/scala/lucuma/odb/graphql/topic/DatasetTopic.scala
```diff
@@ -0,0 +1,53 @@
...
+      id"ch_dataset_edit",
+      _._3,
+      (update, users) => Element(update._1, update._2, update._3, update._4, update._5, users)
+    ) {
+      case Array(_did, _oid, _pid, _isWritten, _tg_op) =>
+        (
+          Gid[Dataset.Id].fromString.getOption(_did),
+          Gid[Observation.Id].fromString.getOption(_oid),
+          Gid[Program.Id].fromString.getOption(_pid),
+          "true".equalsIgnoreCase(_isWritten).some,
```
cquiroz, line 43: Why is this hardcoded?

swalker2m:
@tpolecat One stated use case for this is for `gacq`, the Gemini Acquisition script that science uses.  This is the second case we have (the first being the Gemini Observatory Archive) where we need a read-only user with some limited access to all the programs.  We could provide a `gacq` service user for this, but again, that seems like too much power for code outside of our realm.  I don't think we need to act on this immediately, but I suppose we could do the same thing that we did for GOA: configure the database with a special user account and hack it to allow that user access to particular items.  This may be a more general need though.

cquiroz (approved):
LGTM

modules/service/src/main/scala/lucuma/odb/graphql/topic/DatasetTopic.scala
swalker2m: Here is the corresponding `pg_notify` that produces the data

```plpgsql
  PERFORM pg_notify(
    'ch_dataset_edit',
    array_to_string(
      ARRAY[
        NEW.c_dataset_id,
        NEW.c_observation_id,
        pid,
        -- isWritten
        CASE WHEN NEW.c_end_time IS NOT NULL THEN 'true' ELSE 'false' END,
        TG_OP
      ],
      ','
    )
  );
```

We need to parse the `String` array elements into the types we expect.  It's either `true` or `false` according to the notify but I just treat everything other than `"true"` as `false`.  There's `java.lang.Boolean.parseBoolean` that I could probably use instead.  Or probably there's a better option?

cquiroz: ah right, seems fine

PR #1630, opened 2025-02-28
Remove program id from attachment endpoints

toddburnside:
This is support of Custom SEDs. They are attachments and the ITC needs to call the endpoints. This removes the need for it to know the program id. Only the GET endpoints needed changed for this, but all were changed for the sake of consistency.


rpiaggio (approved):
LGTM

PR #1638, opened 2025-03-03
Updated list of SEDs

cquiroz:


swalker2m (commented):
We store these as part of a target profile JSON blob, I believe.  I think this was asked and answered but just in case, will any existing profiles no longer be decodable as a result of this change?

cquiroz:
All existing tags remain, if you encoded only the tag it should work just fine

swalker2m (approved):

rpiaggio:
I pushed a commit to implement `flushDensitiesAttachment` encoding/decoding. I'm not sure if we need to re-insert the targets from https://github.com/gemini-hlsw/lucuma-odb/blob/main/modules/service/src/main/resources/db/migration/V0894__spectrophotometric_targets_sed.sql#L9

swalker2m:
> I pushed a commit to implement `flushDensitiesAttachment` encoding/decoding. I'm not sure if we need to re-insert the targets from https://github.com/gemini-hlsw/lucuma-odb/blob/main/modules/service/src/main/resources/db/migration/V0894__spectrophotometric_targets_sed.sql#L9

At least one of these targets was deleted by our user (because it was too faint) and I updated the code to ignore deleted targets.  It seems we'd need a migration of these targets rather than reinserting them?

rpiaggio:
This PR turned into 2 PRs into one, but they are both necessary simultaneously since they relate to the latest core bump.

Someone may want to review my 2 commits.

modules/service/src/main/resources/db/migration/V0955__add_fluxDensitiesAttachment.sql
```diff
@@ -0,0 +1,3 @@
+-- Add fluxDensitiesAttachment to SEDs
+
+UPDATE t_target SET c_source_profile = REPLACE(c_source_profile::text, '"fluxDensities": null', '"fluxDensities": null, "fluxDensitiesAttachment": null')::jsonb;
```
swalker2m, line 3: Hmm, if the decoder handles a missing `fluxDensitiesAttachment` maybe we don't need a migration after all?  If we do, would this fail for targets with a non-null `fluxDensities`?

rpiaggio: I don't think we have any targets with non-null fluxDensities, there's no way to do that from the UI. Someone would have had to call the API with that.

rpiaggio: I thought I could use `REGEXP_REPLACE` but it's not that simple since `fluxDensities` can contain an object. I suggest we leave it as it is, or remove the migration.

swalker2m: Assuming it works to do so, and it seems like it should, I vote to remove the migration altogether.

PR #1644, opened 2025-03-03
access control for `createObservation`

tpolecat:
This adds "new" style access control for `createObservation`.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/graphql/mapping/AccessControl.scala
+++ b/modules/service/src/main/scala/lucuma/odb/graphql/mapping/AccessControl.scala
@@ -12,12 +12,15 @@ import grackle.Predicate.*
 import grackle.Query
 import grackle.Query.*
 import grackle.Result
+import grackle.ResultT
 import lucuma.core.enums.ObservationWorkflowState
 import lucuma.core.model.Access
 import lucuma.core.model.Observation
+import lucuma.core.model.Program
 import lucuma.core.model.Target
 import lucuma.core.model.User
 import lucuma.itc.client.ItcClient
+import lucuma.odb.graphql.input.CreateObservationInput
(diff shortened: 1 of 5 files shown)


modules/service/src/main/scala/lucuma/odb/service/ObservationService.scala
```diff
@@ -89,13 +87,9 @@ sealed trait ObservationService[F[_]] {
...
 
   def createObservation(
-    input: CreateObservationInput
+    input: AccessControl.CheckedWithId[ObservationPropertiesInput.Create, Program.Id]
   )(using Transaction[F]): F[Result[Observation.Id]]
 
-  def selectObservations(
-    which: AppliedFragment
-  )(using Transaction[F]): F[List[Observation.Id]]
-
```
tpolecat, line 98: This service method was only used once and is inlined in `ObservationWorkflowService` below.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/AccessControl.scala
```diff
@@ -330,5 +375,50 @@ trait AccessControl[F[_]] extends Predicates[F] {
...
+    def props(pid: Program.Id): F[Result[ObservationPropertiesInput.Create]] =
+      val props = input.SET.getOrElse(ObservationPropertiesInput.Create.Default)
+      props.scienceBand match
+        case None =>
+          allocationService
+            .selectScienceBands(pid)
+            .map(_.toList)
+            .map:
+              case List(b) => Result(props.copy(scienceBand = Some(b)))
+              case _       => Result(props)
```
swalker2m, line 407: I wonder if we should pick the lowest band.  I guess we weren't doing that in the old code though.

swalker2m (approved):
This looks great.  I think the access control stuff has improved the code appreciably.

PR #1649, opened 2025-03-06
clone observation access control

tpolecat:
WIP - tests all pass, which means the tests aren't good enough. need to add a few more


swalker2m (approved):
This looks great, sorry about the last minute conflict... :+1:

PR #1650, opened 2025-03-06
Shortcut 4671: program notes (part 1)

swalker2m:
This is work in progress on adding program notes.  Included:

* `ProgramNote` mappings
* Private notes
* `createProgramNote`

Next up:

* Ordering program notes?  Depends on a pending conversation about scope.
* Otherwise editing program notes


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -3596,6 +3615,38 @@ input ProgramPropertiesInput {
...
+  isPrivate: Boolean
+
+  """
+  To place a program note at a particular index in the sequence of notes,
+  specify the note id before which this note should be added or moved.  If
+  not specified on creation, the note is added at the end.  If not specified
+  on edit, the note is not moved.  If set to `null` the note is moved or added
+  at the end.
+  """
+  before: ProgramNoteId
```
swalker2m, line 3642: This is not yet implemented and may be removed.  My idea was to just renumber all notes every time the order changes.  This, I think, would be the easiest thing that could possibly work and probably be sufficient given that we don't expect more than a handful of notes per program.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/AccessControl.scala
```diff
@@ -421,4 +428,19 @@ trait AccessControl[F[_]] extends Predicates[F] {
 
   } 
 
+  def selectForUpdate(
```
swalker2m, line 431: Does this seem correct?

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
tpolecat: I wonder how hard it would be to make notes GroupElements, so they use the same mechanics as groups and observations. Maybe we don't want this much flexibility?

tpolecat (approved):
This looks good. Thanks for doing the `selectForUpdate`

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
swalker2m: Thanks, that's a very interesting idea that hadn't occurred to me! We may not need notes within groups, but there would be no requirement that Explore make use of that feature.  If it would allow me to piggy back on the group element ordering and have it work the same way in the API then it seems like it could be a win.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/AccessControl.scala
```diff
@@ -376,24 +380,27 @@ trait AccessControl[F[_]] extends Predicates[F] {
...
+    programService
+      .resolvePid(pid, prop, prog)
+      .flatMap: r =>
+        r.flatTraverse: pid =>
+          selectForProgramUpdateImpl(None, List(pid))
+            .map: r =>
+              r.flatMap:
+                case List(pid) => Result(Some(pid))
+                case Nil       => Result(None)
+                case _         => Result.internalError("Unpossible: resovePidWritable returned multiple ids")
```
toddburnside, line 398: "Unpossible"! I like it.

toddburnside (approved):
LGTM

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
toddburnside: Explore may not have to allow the create of notes within groups, but if the API allowed it then explore would have to support displaying it. We might have to make a new tab for notes in that case so we could display the tree to the left side like we do for groups/observations. And then we might as well support nesting. We already have an example of that, so it can be done.

swalker2m: Per discussion in the team spirit meeting I'll just remove the option of controlling note order and always return them sorted by id.

PR #1655, opened 2025-03-07
Migration to add fluxDensitiesAttachment

rpiaggio:
Turns out we did need to add the new fluxDensitiesAttachment to existing targets.

This PR does so in a safe way by manipulating at the JSON level.


toddburnside (approved):
LGTM

PR #1658, opened 2025-03-10
Basic support for the new ITC supporting time and count exposure mode

cquiroz:
This PR adopts the ODB to use the new ITC client that supports time and count mode.
I ran some of the tests locally, but I need to wait for the ITC library to run them properly on CI.

But I'd appreciate a review before doing that as it is a breaking change.

Note that we are not yet using all the features of the new itc model, in particular we are using always `total` sn while we should store both `total` and `single`


swalker2m (approved):
Looks good. :+1: when the library is updated.

PR #1660, opened 2025-03-10
Shortcut 4836: Acquisition breakpoint update

swalker2m:
Though correct to pause before each image through the slit, Observe is already stopping the sequence and asking if another step should be taken after the initial image.  Having a breakpoint programmed into the sequence in this case causes an awkward interaction in which the user responds "Yes, take another image." only to be immediately stopped by a breakpoint.

I think the long-term solution might be to just let Observe handle acquisitions and not try to generate the acquisition sequence.  This PR should improve the situation in the meantime.


tpolecat (approved):
okie dokie

PR #1665, opened 2025-03-11
Fix duration used for parallactic angle calculation

toddburnside:
When calculating the parallactic angle, an explicitly set observation duration was being ignored and the full remaining time was always being used.


tpolecat (approved):
Makes sense.

PR #1666, opened 2025-03-11
Validate migrations and schema before running the build and tests

toddburnside:
This change seems to work and we can fail for these validations without waiting for all of the tests to pass...


cquiroz (approved):
Excellent, thanks

PR #1668, opened 2025-03-12
Update for AGS guide star selection changes

toddburnside:
The tests changed because the new algorithm in AGS is selecting a different angle from among equally good angles.


cquiroz (approved):
LGTM

PR #1669, opened 2025-03-12
Shortcut 4428: restore original GMOS Smart Gcal definition files

swalker2m:
To get #1648 merged as quickly as possible, I ended up editing the GMOS smart gcal files to remove obsolete entries that would otherwise stop the parser.  In this PR I've restored the original Smart Gcal files that science provided and updated the parser to skip obsolete items in the lines that it reads.


swalker2m:
Testing locally, a reload of GMOS North smart gcal data takes 14 seconds locally on an empty database and 17 seconds locally if it needs to clean existing smart gcal data first.

swalker2m:
This gives an [overview of what happens on an update](https://github.com/gemini-hlsw/lucuma-odb/blob/main/modules/service/src/main/scala/db/migration/SmartGcalLoader.scala).  Does anything jump out as particularly inefficient?

Locally, all four GMOS `.csv` files are read in a test case two times each and processed for testing in under a total of 3 seconds.  Doesn't seem like the file reading itself should be a big issue. It would have to be database IO.  It uses a `copy from stdin` to do it, which seems like it would be relatively fast.

cquiroz (approved):
LGTM

PR #1671, opened 2025-03-13
Shortcut 5017: workflow state / remaining time disagreement

swalker2m:
There is a [reported issue](https://app.shortcut.com/lucuma/story/5017/disagreement-between-sequence-workflow-state-and-pending-time) that workflow state and remaining time do not always agree.  This PR adds a test case which completely executes an observation and then samples the workflow state and remaining time.  It sometimes gives the expected results (Completed, 0 ms remaining) but often does not (Defined, 0 ms remaining).

There seems to be a timing issue somewhere.


modules/service/src/test/scala/lucuma/odb/graphql/issue/shortcut/5017.scala
```diff
@@ -0,0 +1,138 @@
...
+        val c = json.hcursor.downField("observation")
+        (for
+          s <- c.downFields("workflow", "state").as[ObservationWorkflowState]
+          t <- c.downFields("execution", "digest", "science", "timeEstimate", "total", "microseconds").as[Long]
+        yield (s, t)).leftMap(f => new RuntimeException(f.message)).liftTo[IO]
+
+    val obtained =
+      for
+        init     <- setup
+        executed <- currentState(init._1)
```
swalker2m, line 134: Delaying for a second before sampling the state above seems to make it work reliably.

tpolecat: urgh

swalker2m: When it fails, it is because the ITC results are not present.

swalker2m: And furthermore, it isn't that the ITC results have been deleted or that the hash no longer matches or something but rather they just haven't yet been written at all and only appear later after the workflow state has already been requested.  Or so it seems.

toddburnside: Is there any kind of subscription event sent when ITC results are updated? I thought there was, but I could be thinking of something else.

swalker2m: Yes, when the digest or itc cache result is updated there is an observation edit event.  I don't think that's related?

toddburnside: I would think explore should re-request the workflow when it gets the observation edit event for the itc cache update and the workflow would then match.

swalker2m: Hmm, yes, perhaps the nondeterministic part of this comes from the fact that I request the digest and the workflow state in the same query.  The workflow is calculated with an effect handler.  If I explicitly request the digest first and then ask for the workflow it seems to work.

cquiroz (approved):
LGTM

PR #1676, opened 2025-03-14
Shortcut 4973: Explicitly mark observations as complete

swalker2m:
This PR implements [Shortcut 4973](https://app.shortcut.com/lucuma/story/4973/allow-the-workflow-transition-ongoing-observed?vc_group_by=week&ct_workflow=all&cf_workflow=500000071).  It adds a `declaredComplete` flag to observations.  When set, the observation will be considered complete even if steps remain.  If unset, the natural completion state calculation based upon the sequence and datasets is used.

Looking at the task, it requests to enable the workflow transition `Ongoing` -> `Complete`, which this implementation isn't quite doing.  `Completed` isn't included in the `UserState` subtype and it doesn't seem like it should be?  I could add a `DeclaredComplete` workflow state and add _that_ to `UserState`.  @tpolecat what do you think?


swalker2m:
Now that I think about it, I added it to the API as an explicit thing you can set, but maybe instead we want to just set the flag when the transition from `Ongoing` to `Complete` is made?

rpiaggio:
I'm guessing we should disable running the sequence in observe if it's complete.

Should we show the remaining steps? Same question for explore

swalker2m:
> I'm guessing we should disable running the sequence in observe if it's complete.

Let's ask Andy but I bet they'd want to undo the flag and enable running the remainder.

> Should we show the remaining steps? Same question for explore

I think yes.  I think the remaining time estimate will be zero though and the workflow state will be `Complete`.

tpolecat (approved):
This looks good.

We need to add the allowed transition from Ongoing to Complete, and add a handler that sets your special bit when this transition is requested. I think this bit should probably be distinct from UserState. We also need to allow changing Complete back to Ongoing if the bit is set.

There is a weird case that's possible in the database at least, where an observation that's not even ready for execution could be marked as complete, and when it becomes Ready it will automatically transition to Complete. I think we can probably ignore this case.

In any case if you want to merge this I can handle the changes to WorkflowService.

👍🏻

PR #1690, opened 2025-03-24
WhereExecutionEvent Update

swalker2m:
A minor requested change to `WhereExecutionEvent`.  Instead of filtering on observation id, it instead allows filtering on any `WhereObservation` parameter.  This should permit the requested feature of getting all events for a particular program via:

```
query {
  events(
    WHERE: {
      observation: { program: { reference: { label: { EQ: " ... " } } } }
    }
  ) {
    matches { id }
  }
}
```


toddburnside (approved):
LGTM

tpolecat (approved):

PR #1692, opened 2025-03-24
Fix for error calculating itc invalid tagged type

cquiroz:
The error comes from ITC because the SED on the odb can be set to null.


rpiaggio (approved):
LGTM

swalker2m (approved):
:+1:

PR #1701, opened 2025-03-27
SC-4973 part 2

tpolecat:
This removes `declaredComplete` from the schema and folds it into workflow state. By setting an `Ongoing` observation to `Completed` the bit is set. A completed observation can also be set back to `Ongoing` if it was set to `Completed` in this way (but not if it's "naturally" complete in terms of its science requirements).

This took my small brain a long time to do because the execution digest cache wasn't being invalidated … not sure if I did this right. @swalker2m see the note below.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/logic/Generator.scala
+++ b/modules/service/src/main/scala/lucuma/odb/logic/Generator.scala
@@ -259,6 +259,9 @@ object Generator {
           // Commit Hash
           md5.update(commitHash.hashBytes)
 
+          // Completion Declaration
+          md5.update(params.declaredComplete.hashBytes)
+
           Md5Hash.unsafeFromByteArray(md5.digest())
         }
 
--- a/modules/service/src/main/scala/lucuma/odb/logic/Generator.scala
+++ b/modules/service/src/main/scala/lucuma/odb/logic/Generator.scala
@@ -247,8 +247,8 @@ object Generator {
         val hash: Md5Hash = {
           val md5 = MessageDigest.getInstance("MD5")
 
-          // Observing Mode
-          md5.update(params.observingMode.hashBytes)
+          // Generator Params
+          md5.update(params.hashBytes)
 
           // Integration Time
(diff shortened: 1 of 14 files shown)


modules/service/src/main/scala/lucuma/odb/logic/Generator.scala
```diff
@@ -259,6 +259,9 @@ object Generator {
           // Commit Hash
           md5.update(commitHash.hashBytes)
 
+          // Completion Declaration
+          md5.update(params.declaredComplete.hashBytes)
+
```
tpolecat, line 264: I don't know if this is the right approach. The issue is that the execution state query wasn't respecting changes to `c_declared_complete` so I was still getting `Ongoing` back even if the bit was set.

tpolecat:
What do I do when "Something is wrong with your schema"?

modules/service/src/main/scala/lucuma/odb/logic/Generator.scala
swalker2m: D'oh I added the declared complete flag to the `GeneratorParams` hash, but here we're picking individual items out of the params so what you've done is correct.  We may want to switch to `md5.update(params.hashBytes)` and remove the lines that pick individual items out of it.  It would then include the science band which doesn't really impact sequence generation yet, but that also doesn't change often.  It would also include the calibration role, which does matter and probably should have been added here as well.

swalker2m (approved):
Pending a bit of test case fixing, this looks great.  :+1:

modules/service/src/main/scala/lucuma/odb/logic/Generator.scala
tpolecat: I guess calibration role never changes either, so maybe it's not a big deal. But yeah just to simplify (since caching is hard) I will change it as you suggest.

tpolecat:
I accidentally only ran the mutation tests locally, sorry. The others revealed a few other cases I needed to consider.

PR #1702, opened 2025-03-27
Switch to Http4s Ember server

hugo-vrijswijk:
See also https://github.com/gemini-hlsw/navigate-server/pull/868


swalker2m:
At one point long ago, we tried the Ember server and it wasn't working for some reason.  Maybe it didn't do web sockets or something.  @tpolecat do you recall? I wish I could remember the details, but if it is working now :+1:

swalker2m:
Found this in chats, though it's so old now as to be probably irrelevant and I never did get to the bottom of it anyway.  I don't know which "tests" I was referring to because our tests take like 45 minutes.

> the ember version works and the tests take 32 seconds
> the blaze version works and the tests take 2 seconds
> i must be doing something wrong
> sounds like a 30 timeout of some sort is happening

tpolecat:
Yes, there were a lot of problems but I don't recall what they were. We can try this but need to be ready to roll it back.

toddburnside:
> Found this in chats, though it's so old now as to be probably irrelevant and I never did get to the bottom of it anyway. I don't know which "tests" I was referring to because our tests take like 45 minutes.
> 
> > the ember version works and the tests take 32 seconds
> > the blaze version works and the tests take 2 seconds
> > i must be doing something wrong
> > sounds like a 30 timeout of some sort is happening

@armanbilge was recommending switching to Ember. IIRC, Ember was supposed to have improved and is now faster than blaze? At one point I started to change it but was running into some issues. Unfortunately, I don't remember what those issues are.

toddburnside:
> Yes, there were a lot of problems but I don't recall what they were. We can try this but need to be ready to roll it back.

I think we should try it. Rolling back isn't hard and if we break dev for awhile. 🤷‍♂️

toddburnside:
> > Cancelled after 60m
> 
> I think there was an ember shutdown timeout of 30 seconds. I switched it to 1 and sped things up in an aborted attempt to switch before.

Argh. Now that you  mention it, I seem to recall something about those timeouts. We're also getting `java.net.BindException: Address already in use` in the tests, I think in response to the attachment uploading.

hugo-vrijswijk:
The tests should be green now. The `bindAny` is replaced with `.withPort(port"0")`, which does the same thing. But I forgot to add it yesterday

hugo-vrijswijk:
There's still a 60 minute timeout, unfortunately. Everything is running fine locally. I'm not sure why Ember seems slower

modules/service/src/main/scala/lucuma/odb/Main.scala
```diff
@@ -178,15 +179,15 @@ object FMain extends MainParams {
...
+  def serverResource[F[_]: Async: Network](
     port: Port,
     app:  WebSocketBuilder2[F] => HttpApp[F]
   ): Resource[F, Server] =
-    BlazeServerBuilder
-      .apply[F]
-      .bindHttp(port.value, "0.0.0.0")
+    EmberServerBuilder.default[F]
+      .withPort(port)
+      .withHost(host"0.0.0.0")
```
swalker2m, line 188: I think it might be the shutdown timeout that adds 30 seconds to every test case. I believe I "solved" this by decreasing the default.  I don't know what the ramifications of that would be though.  Maybe test `.withShutdownTimeout(Duration.Zero)`?

hugo-vrijswijk: Yeah I think this was the fix for the tests. Locally tests take 5 seconds longer without this line with Ember (but not with Blaze). And with this Ember and blaze takes the same amount of time

swalker2m (approved):
Looks great. :+1:

PR #1703, opened 2025-03-28
Store the full SignalToNoiseAt of itc results

cquiroz:
This turned out simpler than what I expected as we store the result as a blob


swalker2m (approved):
Looks great. :+1:

PR #1706, opened 2025-03-31
Changes to Average Parallactic Angle calculation and obs duration

toddburnside:
For sc-5061 ([app.shortcut.com/lucuma/story/5061/average-parallactic-angle-calculation-should-only-include-science-portion](https://app.shortcut.com/lucuma/story/5061/average-parallactic-angle-calculation-should-only-include-science-portion)) this changes the average parallactic angle calculation to ignore the setup time and calculate with science time (obs time + setup time) and science duration (obs duration - setup time).

For [sc-5193] ([app.shortcut.com/lucuma/story/5193/do-not-automatically-recalculate-average-parallactic-angles](https://app.shortcut.com/lucuma/story/5193/do-not-automatically-recalculate-average-parallactic-angles), obs duration is now required and must be greater than the setup time and no greater than the estimated remaining time.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/GuideService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/GuideService.scala
@@ -196,9 +196,12 @@ object GuideService {
     def obsTime: Result[Timestamp] =
       optObsTime.toResult(generalError(s"Observation time not set for observation $id.").asProblem)
 
-    def validGuideStarName(generatorHash:Md5Hash, remainingTime: TimeSpan): Option[GuideStarName] =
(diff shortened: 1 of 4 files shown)


swalker2m (approved):
This looks good.  :+1:  I wonder if we could or should, at some point, separate the duration and start time from the observation itself.

toddburnside:
> This looks good. 👍 I wonder if we could or should, at some point, separate the duration and start time from the observation itself.

That's a interesting idea. It doesn't follow the readonly rules that the rest of the observation does, and it isn't really part of the "observation" except that it affects parallactic angle/guide star selection. Maybe all of that stuff should be separate?

PR #1707, opened 2025-04-02
Enable Slew Events

swalker2m:
Slew events will be sent by Navigate, while all other execution events (sequence, step, dataset) are sent by Observe.  All events, regardless of source, are associated with a "visit".  The visit is a time accounting concept and a hopefully intuitive notion of a session spent observing some part of an observation.

The issue tackled by this PR is to allow Navigate to send slew events without knowing with which visit they will be associated. Accordingly the API is adjusted such that the `AddSlewEventInput` now takes an `ObservationId` instead of a `VisitId`.

In the implementation, when a slew event arrives we find the current "chargeable" visit.  Only one "chargeable" visit is active at a time per site because we cannot double charge time.  A visit is chargeable if time spent doing it will be charged against the program.  If there is already a chargeable visit for the same observation, we use that visit.  If not we create a new "preliminary" visit and it becomes the chargeable visit.  It is preliminary because it will be missing its static instrument configuration.

Observe continues recording visits explicitly.  If the observation is chargeable, we'll first try to find an existing chargeable visit.  If there is one, and it is for this observation, and it is preliminary, we just add the static configuration.  Otherwise we make a new visit as usual.

Notes
* Please know that I'm sorry about all of this and will gladly accept alternative suggestions.
* The most import updates are in `VisitService`.  Some chaff was generated as I meandered through various options
* A followup PR will address adding an "overlap" discount similar to what we have in `ocs` such that when a chargeable visit begins time spent on any already ongoing chargeable visit is free.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/TimeService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/TimeService.scala
@@ -1,0 +1,48 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.odb.service
+
+import cats.effect.Concurrent
+import cats.syntax.functor.*
+import lucuma.core.enums.Site
+import lucuma.core.model.ObservingNight
+import lucuma.core.util.Timestamp
+import skunk.*
+import skunk.codec.temporal.timestamptz
+import skunk.implicits.*
+
+import Services.Syntax.*
+
+/**
+ * A service for database-time related queries.
+ */
+trait TimeService[F[_]]:
+
+  /**
+   * The database time associated with the current transaction.
+   */
+  def transactionTime(using Transaction[F]): F[Timestamp]
+
+  /**
+   * The observing night associated with the site and transaction time.
+   */
+  def currentObservingNight(site: Site)(using Transaction[F]): F[ObservingNight]
+
+object TimeService:
+
+  def instantiate[F[_]: Concurrent](using Services[F]): TimeService[F] =
+    new TimeService[F]:
+
+      override def transactionTime(using Transaction[F]): F[Timestamp] =
+        session.unique(Statements.TransactionTime)
+
+      override def currentObservingNight(site: Site)(using Transaction[F]): F[ObservingNight] =
+        transactionTime.map(t => ObservingNight.fromSiteAndInstant(site, t.toInstant))
+
+  object Statements:
+
+    val TransactionTime: Query[Void, Timestamp] =
+      sql"""SELECT CURRENT_TIMESTAMP(6)"""
+        .query(timestamptz(6))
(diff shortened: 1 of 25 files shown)


modules/service/src/main/scala/lucuma/odb/service/VisitService.scala
```diff
@@ -36,82 +47,179 @@ trait VisitService[F[_]] {
...
+        def obsNight(site: Site): ResultT[F, ObservingNight] =
+          ResultT.liftF(timeService.currentObservingNight(site))
+
+        // Application transaction advisory lock on visit creation. The idea is
+        // to prevent two callers (Observe and Navigate) from doing a lookup
+        // simultaneously and coming to the conclusion to each create a new
+        // visit.
+        val lockVisitCreation: ResultT[F, Unit] =
+          ResultT.liftF:
+            session.unique(Statements.LockCreation)(Statements.VisitCreationLockId)
```
swalker2m, line 120: I'll just call attention to arguably the sketchiest part of the PR above.

cquiroz:
Is a visit associated with an observation? Can we have more than one visit running concurrently? for example one instrument doing science and another with calibrations.

cquiroz:
Observe keeps track of the visit IDs internally. It records a visit and keeps a reference to the visit ID.

If the visit was started by navigate you will get the ID from the visit, correct?.

and otherwise if we start the visit in observe, navigate will not care, right?

tpolecat:
Thank you for the detailed description. I will look at the code when I have stopped crying and possibly had some scotch.

modules/service/src/main/scala/lucuma/odb/service/TimeService.scala
```diff
@@ -0,0 +1,48 @@
...
+import skunk.*
+import skunk.codec.temporal.timestamptz
+import skunk.implicits.*
+
+import Services.Syntax.*
+
+/**
+ * A service for database-time related queries.
+ */
+trait TimeService[F[_]]:
```
tpolecat, line 20: :+1: This is a good idea.

modules/service/src/main/scala/lucuma/odb/service/VisitService.scala
tpolecat: I have never used an advisory lock but this seems reasonable to me, since there is no way to detect this situation via a key collision.

tpolecat (approved):
I'm very sorry you had to figure this out but the solution is good. :+1:

swalker2m:
> Is a visit associated with an observation?

Yes.

> Can we have more than one visit running concurrently? for example one instrument doing science and another with calibrations.

Yes.  Only one visit at a time is being charged.  Multiple calibrations may be running though.  You can even be writing out datasets while slewing to another observation.  When the slew starts, any remaining time in the science observation it overlaps will be discounted.

> If the visit was started by navigate you will get the ID from the visit, correct?.

Yes.

> and otherwise if we start the visit in observe, navigate will not care, right?

Correct, the slew will just be associated with the visit opened by Observe.

PR #1672, opened 2025-03-13
ProgramService access control

tpolecat:
This adds the new access control setup to `ProgramService`

PR #1708, opened 2025-04-02
WIP - targetservice access control

tpolecat:
Ignore, this needs to be rebased atop #1672


modules/service/src/main/scala/lucuma/odb/graphql/mapping/AccessControl.scala
```diff
@@ -622,5 +640,53 @@ trait AccessControl[F[_]] extends Predicates[F] {
...
+            Group(List(
+              Select("id", None),
+              Select("program", Select("id", None))
+            ))
+          ),
+          Context(QueryType, List("targets"), List("targets"), List(TargetType))
+        ).flatMap(_.fragment)
+          .flatTraverse: af =>
+            // N.B. the ordering of return columns here is not obvious
+            session.prepareR(af.fragment.query(program_id *: target_id)).use: pq =>
```
swalker2m, line 664: Hmm, yes it seems like it would be reversed.

swalker2m (approved):
Looks great. :+1:

PR #1709, opened 2025-04-03
Migrate core: ImageQuality and CloudExtinction Presets

rpiaggio:


swalker2m (approved):
The (automatic?) reformatting of the graphql schema makes it a bit difficult to spot the changes but this seems reasonable.  :+1:

PR #1713, opened 2025-04-03
Create base image for db tests

hugo-vrijswijk:
Adds a base Docker image with a pre-initialized database. With this each testcontainer does not have to re-run the init scripts for every suite, speeding up the total test run. The image will be built once and then re-used for every suite thereafter

### DIFF
--- a/modules/service/src/Dockerfile
+++ b/modules/service/src/Dockerfile
@@ -16,6 +16,9 @@ ENV PGDATA=/var/lib/postgresql/odb-data
 # Copy migration scripts into the image
 COPY ./main/resources/db/migration/ /docker-entrypoint-initdb.d/
 
+# Remove the Postgres daemon from the entrypoint script to only run the init scripts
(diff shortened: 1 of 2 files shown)


toddburnside (approved):
It looks good to me. But, it didn't make as big of a difference in test time as I expected.

toddburnside:
Actually, I don't understand how this creates a single docker image used by all of the test suites. Maybe you could put some comments in explaining that?

hugo-vrijswijk:
> Actually, I don't understand how this creates a single docker image used by all of the test suites. Maybe you could put some comments in explaining that?

I've added a comment now. Essentially the Docker image will still be "rebuilt" every suite, but it will be in the Docker cache. So all the steps are skipped

hugo-vrijswijk:
> It looks good to me. But, it didn't make as big of a difference in test time as I expected.

Yeah it's not as drastic as going from 45 to 15 😅. But it seems to save around 3 minutes per shard

PR #1716, opened 2025-04-04
Filter zero exposure time

swalker2m:
This PR just rejects `TimeAndCount` exposure time mode when the exposure time is zero.  It is correct that biases have a zero exposure time, but they will not be specified via a `TimeAndCount` exposure time mode science requirement.

Down inside the GMOS generators, another check is made in order to fail more gracefully than "divide by zero".


tpolecat (approved):
Excellent 👍🏻

PR #1717, opened 2025-04-04
Use timestamp of a custom SED as part of the itc validation hash

toddburnside:
For custom seds, the attachment id is stored in the json blob of the source profile. The source profile is part of the validation hash for the itc calls, so if the attachment id changes, the itc is requeried. However, the user can also upload a new file for that attachment and the id does not change. In this case we still need to requery the itc. The best way I could figure to get around this was to include the timestamp for the attachment in the hash, because that will change if a new file is uploaded.

If someone has a better idea, let me know. :)


cquiroz (approved):
LGTM

PR #1719, opened 2025-04-07
binding module

tpolecat:
This adds a JVM-only published module `lucuma-odb-binding` that contains every class from `graphql.binding` and referenced classes from `data`, most of which will probably end up in core at some point. I was thinking it might make sense to leave some of this stuff in `service` but I had a hard time thinking of a good reason. Anyway, comments pls. @swalker2m @cquiroz


cquiroz (approved):
Thanks a lot, LGTM

swalker2m (approved):
Seems fine to me. :+1:

PR #1720, opened 2025-04-07
Overlap Discount (Prevent Double-Charge)

swalker2m:
Adds a new time accounting discount, the `Overlap` discount.  This is time spent in a visit that is overlapped by another chargeable visit.  The quintessential example of this is time spent writing out the last datasets of an observation while slewing to a new observation.  In this case, the observation we are slewing to is charged immediately and the one writing out datasets is no longer charged.

I've taken the simplifying assumption that once you've been overlapped by another observation you will no longer be charged, even if the original observation outlasts the overlapping observation.

The time accounting "invoice" entry for this would look like:

```json
                {
                  "interval" : {
                    "start" : "2023-12-20 08:36:12.241526",
                    "end" : "2023-12-20 08:36:22.241526"
                  },
                  "amount" : {
                    "seconds" : 10.000000
                  },
                  "comment" : "Time spent observing after an overlapping chargeable observation started.",
                  "observation" : {
                    "id" : "o-11c"
                  }
                }
```

where `o-11c` is the observation that overlapped this one.


modules/service/src/main/resources/db/migration/V0978__overlap_discount.sql
```diff
@@ -0,0 +1,75 @@
...
+    FROM
+      t_execution_event
+    GROUP BY
+      c_visit_id
+  ) sub
+ WHERE v.c_visit_id = sub.c_visit_id;
+
+-- Add an index to make the overlap query faster.
+CREATE INDEX idx_t_visit_site_chargeable_start
+  ON t_visit (c_site, c_chargeable, c_start);
```
swalker2m, line 45: I think we need an index on the start time at least, but we also require visits to be for the same site and to be chargeable so those are in the overlap query.  Tbh, I'm not sure whether the index should include these as well but it seemed like it?

modules/service/src/main/scala/lucuma/odb/service/TimeAccountingService.scala
```diff
@@ -375,13 +385,47 @@ object TimeAccountingService {
...
+          v2.c_site = v.c_site          AND
+          v2.c_chargeable = true        AND
+          v2.c_start < v.c_end          AND
+          (
+            -- The usual case if v2 overlaps
+            v2.c_start > v.c_start                                    OR
+
+            -- The edge case where the start times are the same on two visits.
+            -- Here we just pick the one with the newer id to be charged.
+            (v2.c_start = v.c_start AND v2.c_visit_id > v.c_visit_id)
```
swalker2m, line 422: This will never happen, of course, but if it did we'd double charge these observations without this bit.  It just picks one at random (the one whose visit was last created) to charge and the other one is free.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -12126,18 +12132,46 @@ type TimeChargeNoDataDiscount implements TimeChargeDiscount {
   interval: TimestampInterval!
 
   "Time amount to discount from the program."
-  program: TimeSpan!
+  program: TimeSpan! @deprecated(reason: "renamed to 'amount'")
```
cquiroz, line 12135: Do we want to deprecate fields? I thought at this stage we'd just delete them

swalker2m: In this case it was trivial and doesn't leave (much) cruft so I just deprecated.  Might be that nobody uses it anyway.

cquiroz (approved):
LGTM

modules/service/src/main/resources/db/migration/V0978__overlap_discount.sql
tpolecat: That seems correct if you're joining on all of those columns.

tpolecat (approved):
This looks good. Continued sorrow about the complexity of time accounting.

PR #1723, opened 2025-04-08
Sample test to reproduce shortcut 5331

cquiroz:


tpolecat:
Added a fix for the variables case. The GraphQL literal case will never work because it's invalid syntactically.

PR #1734, opened 2025-04-14
Add flamingos2longslit observingmode

cquiroz:
This is a partial implementation at supporting F2 long slit, it is very much based on gmos long slit
It includes: 
* DB model for f2
* schema updates to create observations with f2 long slit
* Implementation of `F2LongSlitService` that for now it only supports creating items
* Tests

Next (as in another PR)
* Support select, update, delete clone operations
* Make the read mode depend on the exposure time


modules/service/src/main/resources/db/migration/V0980__flamingos2_longslit.sql
```diff
@@ -0,0 +1,164 @@
...
+  CASE WHEN o.c_spec_wavelength_coverage IS NOT NULL THEN o.c_observation_id END AS c_spec_wavelength_coverage_id,
+  CASE WHEN o.c_spec_focal_plane_angle   IS NOT NULL THEN o.c_observation_id END AS c_spec_focal_plane_angle_id,
+  CASE WHEN o.c_observation_duration     IS NOT NULL THEN o.c_observation_id END AS c_observation_duration_id,
+  c.c_active_start::timestamp + (c.c_active_end::timestamp - c.c_active_start::timestamp) * 0.5 AS c_reference_time
+  FROM t_observation o
+  LEFT JOIN t_proposal p on p.c_program_id = o.c_program_id
+  LEFT JOIN t_cfp c on p.c_cfp_id = c.c_cfp_id
+  LEFT JOIN t_gmos_north_long_slit  mode_gnls ON o.c_observation_id = mode_gnls.c_observation_id
+  LEFT JOIN t_gmos_south_long_slit  mode_gsls ON o.c_observation_id = mode_gsls.c_observation_id
+  LEFT JOIN t_flamingos_2_long_slit mode_f2ls ON o.c_observation_id = mode_f2ls.c_observation_id;
```
swalker2m, line 163: I can't remember why we join on the modes there.  We don't seem to use it for anything?  I may have used them at one point, removed it, and forgot to clean up.  Perhaps we could remove these last three joins?  Am I missing something?

cquiroz: Not sure, I took it from the existing table and assumed we needed it for f2 too
It could be problematic as we add more modes.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/package.scala
```diff
@@ -0,0 +1,27 @@
...
+
+package lucuma.odb.graphql
+package mapping
+
+import cats.syntax.all.*
+import grackle.skunk.SkunkMapping
+
+import scala.reflect.ClassTag
+
+trait OptionalFieldMapping[F[_]] { this: SkunkMapping[F] =>
```
swalker2m, line 12: :+1:

swalker2m (approved):
Seems great so far. :+1:

modules/service/src/main/resources/db/migration/V0980__flamingos2_longslit.sql
swalker2m: It doesn't seem to contribute anything to the view so let's delete them?

cquiroz: ok I'll update the table

PR #1738, opened 2025-04-16
Support cloning f2 observations

cquiroz:


modules/service/src/main/scala/lucuma/odb/service/ObservingModeServices.scala
```diff
@@ -47,7 +47,7 @@ sealed trait ObservingModeServices[F[_]] {
 
 object ObservingModeServices {
 
-  def instantiate[F[_]: MonadCancelThrow](using Services[F]): ObservingModeServices[F] =
+  def instantiate[F[_]: {MonadCancelThrow, Services}]: ObservingModeServices[F] =
```
swalker2m, line 50: That syntax is new to me.  Is it the same as `F[_]: MonadCancelThrow: Services` ?

swalker2m (approved):
:+1:

modules/service/src/main/scala/lucuma/odb/service/ObservingModeServices.scala
cquiroz: Yes, it came on 3.6 with SIP-64
https://docs.scala-lang.org/sips/sips/typeclasses-syntax.html#2-new-syntax-for-aggregate-context-bounds

PR #1739, opened 2025-04-17
more access control

tpolecat:
This adds a few `Checked` implementations for service methods, but also marks a lot of service methods as requiring `SuperUserAccess`. The idea is:

- Service methods all need an access control indication in their types (either a `Checked` parameter or an `XxxAccess` token).
- Service methods that are invoked directly by GraphQL mutations generally need a `Checked` parameter.
- Service methods that are invoked only by other services can be marked as requiring `SuperUserAccess`, indicating to the caller that no further access control checks are going to happen.
- Service methods that are invoked in read mappings (i.e., their results are serialized back as JSON) are also marked as requiring `SuperUserAccess`; access control is provided here via predicates added during query elaboration.

Sorry for the length, it's mostly mechanical.


swalker2m (approved):
This all makes sense and the code looks great. :+1:

PR #1748, opened 2025-04-24
Shortcut 4434: Obscalc Service

swalker2m:
This is the first part of an update that proposes a change to the way we work with relatively slow calculations like the execution digest. In particular the idea is  an "Obscalc" service that is always running and computing results for updated observations. That way they are more likely to be ready when needed.

First note that at this point:

* the changes are mostly additive and do not modify the existing API or how it works and
* the new Obscalc Service is written but wouldn't be turned on yet.

Though not part of this PR, the API would be changed to accommodate the uncertainty around computed results.  For example currently we have

```graphql
type Execution {
  digest: ExecutionDigest
...
}
```

which could be wrapped in a `Calculated` type to mark it as potentially stale:

```graphql
enum CalculationState {
  PENDING
  RETRY
  CALCULATING
  READY
}

interface Calculated {
  state: CalculationState!
  ... other metadata? ...
}

type CalculatedExecutionDigest implements Calculated {
  state: CalculationState!
  digest: ExecutionDigest
}

type Execution {
  digest: CalculatedExecutionDigest!
  ...
}
```

Today queries for the `ExecutionDigest` block when not available.  This can involve waiting for the ITC service to be called and iterating through the entire sequence.  If these changes are adopted, queries for the `ExecutionDigest` would no longer block. Instead the last known value would be returned along with the "state" of the calculation.  Included in this PR is a new event topic, `ch_obscalc_update` that (when added to the API) will make it possible to be notified when calculations move between `pending`, `ready`, etc.

The PR is large but a good chunk of it is changing sequence generation to work directly with `OdbError` instead of its own error types that are ultimately converted into `OdbError`.  This was necessary in order to store errors when the calculation cannot be completed.

The places to focus the review would be:

* the `runObscalcDaemon` in the `obscalc-service` module `Main`
* the migration `V0984__obscalc.sql`
* the `ObscalcService` itself


rpiaggio:
Looks really good. I have two questions:

- What is `RETRY`?
- Could we keep a blocking endpoint too that only returns `READY`? I'm worried about the impact in clients.

toddburnside:
We touched on this a bit in the meeting today, but what about `workflow`? Would it still block? In explore we currently query execution times separately to avoid the potential slow response, but I'm not exactly sure if that is very useful at the moment since we include the workflow in the main query. 

With this PR, we could potentially move the execution times into the main query with retries when necessary. But, I'm not sure what to do about workflow.

Disclaimer: I have not reviewed this PR in depth yet.

tpolecat:
@toddburnside Workflow state is fast to compute when everything is cached up, which should normally be the case once this service is running.

toddburnside:
> @toddburnside Workflow state is fast to compute when everything is cached up, which should normally be the case once this service is running.

I think the `CalculatedState` in the description is supposed to be useful for when the observation is edited? For example, if Explore edits the observation it can invalidate the cache so that itc needs to be requeried, etc. The new service won't really help in that instance since explore would request the new times immediately. It also requests the workflow immediately. 

Maybe we could have a `CalculatedWorkflow` which uses the last known values for it's calculation? Just throwing some thoughts against the wall.

swalker2m:
> We touched on this a bit in the meeting today, but what about workflow? Would it still block? 

I assume that we wouldn't want to compute a workflow state based on a possibly stale execution state?  Let me think about it a bit.  In the worst case, using this service will mean that usually most execution states will be ready since we don't have to wait for someone to ask in order to update the value.  If the workflow service reads an execution state tied to a non-`ready` calculation it could wait for the result as it does today.

> Maybe we could have a `CalculatedWorkflow` which uses the last known values for it's calculation?

I could definitely provide a workflow state value which never blocks but is sometimes a wee bit stale.  Would that be useful?

I think never waiting on the planned time but using a possibly stale value until the latest computation is done is useful?

toddburnside:
> I could definitely provide a workflow state value which never blocks but is sometimes a wee bit stale. Would that be useful?

I think it would be useful. I think not all workflow states need to get the execution state? So this would only affect some of the observations. If we knew that the workflow state was stale, we could prevent the user changing workflow states (and maybe make the observation readonly?) until we get a "real" value. I need to think about this part a bit more.

> I think never waiting on the planned time but using a possibly stale value until the latest computation is done is useful?

Yes, it's definitely useful. It would eliminate the need for querying the times separately from the main observation query, which does cause a few edge cases in the UI. We recently implemented a way to keep the old time values in the UI while we were waiting to update them. This would simplify that.

swalker2m:
I went ahead and added API for the subscription that would be offered:

```graphql
type Subscription {
  ...
  """
  Subscribes to an event that is generated whenever an observation calculation
  is updated.  If a program id is provided then the event must correspond to an
  observation referenced by that program.  If observation id is provided then
  the event must correspond to that particular observation. The old and new
  calculation states can also be matched to catch transitions, for example to
  and from 'Ready'.
  """
  obscalcUpdate(input: ObscalcUpdateInput): ObscalcUpdate!
  ...
}

"""
Input to the obscalcUpdate subscription.  Specify programId and/or observation
id to filter events to that program and/or observation.  Specify the old and
new states to limit events to only those transitions.
"""
input ObscalcUpdateInput {
  programId:     ProgramId
  observationId: ObservationId
  oldState:      WhereOptionEqObscalcState
  newState:      WhereOptionEqObscalcState
}

"""
Observation background-update calculation states.
"""
enum ObscalcState {

  """
  PENDING signifies that the observation will be updated by a background worker
  in the future.  Calculated items like the execution digest may be updated at
  that point.
  """
  PENDING

  """
  RETRY signifies that the observation update calculation failed and will be
  automatically attempted again even if its properties are not modified.
  """
  RETRY

  """
  CALCULATING signifies that an observation background calculation is currently
  ongoing.
  """
  CALCULATING

  """
  READY signifies that there are no pending calculation updates at the moment
  and calculated values like the execution digest will not change.
  """
  READY
}
```

swalker2m:
> How does this affect the program and group time estimates? Do they just use the stale values for obs times until something new is available? If the observation doesn't even have stale values, will it block until the observation has values?

Yeah, I was thinking you'd get both the time estimate and a combined calculation state.  So if anything in the program is `Pending` the overall calculation state for the program time estimate is also `Pending`.  If for example you have one observation that just became edited such that it is possible to estimate its time, a call to get the time estimate could unfortunately show 0 seconds (`Pending`).  That is, no blocking.  You'd get an event though if you're watching the `obscalcUpdate` events and know when it is updated.

toddburnside:
> Yeah, I was thinking you'd get both the time estimate and a combined calculation state. So if anything in the program is `Pending` the overall calculation state for the program time estimate is also `Pending`. If for example you have one observation that just became edited such that it is possible to estimate its time, a call to get the time estimate could unfortunately show 0 seconds (`Pending`). That is, no blocking. You'd get an event though if you're watching the `obscalcUpdate` events and know when it is updated.

That would work fine. :)

PR #1760, opened 2025-05-05
Shortcut 4434: Obscalc (Version 2)

swalker2m:
(Reissues #1748 after moving some code to core and adding new GraphQL API fields for program and group time estimates using background values.)

This is the first part of an update that proposes a change to the way we work with relatively slow calculations like time estimates. In particular the idea is to add an "Obscalc" service that is always running and computing results for updated observations. That way they are more likely to be ready when needed.

At this point the changes are mostly additive and *do not remove the existing API or how it works*.  Some comments are marked with `OBSCALC TODO` for followup.

API for time estimates has been added (while temporarily leaving the existing API in place) to accommodate the uncertainty around computed results.  For example we currently have

```graphql
type Program {
...
  timeEstimateRange: CategorizedTimeRange
...
}
```

This PR would add:

```graphql
type Program {
...
  timeEstimateRange: CategorizedTimeRange

  # OBSCALC TODO: replace original version with this one
  timeEstimateRange2: CalculatedCategorizedTimeRange
...
}

type CalculatedCategorizedTimeRange {
  state: CalculationState!
  value: CategorizedTimeRange
}
```
which wraps the `CategorizedTimeRange` in a `CalculatedCategorizedTimeRange` type to mark it as volatile.

Today queries for the time range block when not available for one or more observations in the program.  This can involve waiting for the ITC service to be called and iterating through the entire sequence for each observation.  If these changes are adopted, queries for the time estimate range would no longer block. Instead the last known value would be returned along with the "state" of the calculation.  Included in this PR is a new event topic, `ch_obscalc_update`, that enables notifications when calculations move between `pending`, `ready`, etc.

The PR is large but a good chunk of it is changing sequence generation to work directly with `OdbError` instead of its own error types that are ultimately converted into `OdbError`.  This was necessary in order to store errors when the calculation cannot be completed.

The places to focus the review would be:

* the `runObscalcDaemon` in the `obscalc-service` module `Main`
* the migration `V0985__obscalc.sql`
* the `ObscalcService` itself


swalker2m:
## Addendum

* Running locally on my laptop, the 3,305 observations in the development ODB were processed in 32 minutes (or about 100 per minute).  This uses the default of `2 * #processors + 1` connections, or 17 in my case.  I will need to update postgres to increase the maximum number of connections (from 20 to 40) and assign some number to `Obscalc`, perhaps 8.
* Of the 3,305 observations, 1,610 are not fully defined or have some error which prevents generating the sequence.  These errors would now be stored in the `t_obscalc` table in JSON and a subsequent query would immediately return the same error.  Obviously, an update to the observation resets the calculation state to `pending` so that it will be repeated.
* Storing the errors in JSON was expedient because each has its own fields.  It does introduce a maintenance issue though in that we would need to ensure that they remain readable as the error Scala code is changed.  We have the same problem with `SourceProfile`.

modules/service/src/main/scala/lucuma/odb/logic/TimeEstimateService2.scala
```diff
@@ -0,0 +1,195 @@
...
+    calculator:   TimeEstimateCalculatorImplementation.ForInstrumentMode,
+  )(using Services[F]): TimeEstimateService2[F] =
+    lazy val obscalcService: ObscalcService[F] =
+      services.obscalcService(commitHash, itcClient, calculator)
+
+    new TimeEstimateService2[F]:
+
+      // OBSCALC TODO: In core, need to fiddle with `NotGiven` in `CalculatedValue`
+      // to avoid the need to do this.
+      given Semigroup[CalculatedValue[CategorizedTime]] =
```
tpolecat, line 87: You can fix this by slicing the givens into interfaces of increasing generality. This will provide the most specific instance you ask for.

```scala
case class Foo[A](a: A)

object Foo extends FooSlice1:
    given [A: CommutativeMonoid]: CommutativeMonoid[Foo[A]] = ???

trait FooSlice1 extends FooSlice2:
  given [A: Monoid]: Monoid[Foo[A]] = ???

trait FooSlice2:
  given [A: Semigroup]: Semigroup[Foo[A]] = ???
```

tpolecat (approved):
This looks great. A few minor comments. :+1: 👍🏻

swalker2m:
> Running locally on my laptop, the 3,305 observations in the development ODB were processed in 32 minutes (or about 100 per minute).

Rob pointed out what should have been obvious to me before: the long delay was mostly because I'm only processing `connectionLimit` entries per 10 second poll period on startup.  We really want to cycle through them all quickly.  If I grab say, all 3,305 entries, it goes through them in just a few minutes.  That's like 18/s overall or 9.5 _valid_ observations per second.

swalker2m:
> > Running locally on my laptop, the 3,305 observations in the development ODB were processed in 32 minutes (or about 100 per minute).
> 
> Rob pointed out what should have been obvious to me before: the long delay was mostly because I'm only processing `connectionLimit` entries per 10 second poll period on startup. We really want to cycle through them all quickly. If I grab say, all 3,305 entries, it goes through them in just a few minutes. That's like 18/s overall or 9.5 _valid_ observations per second.

I suppose though that some of that speedup could also be that the ITC results got cached in the remote ITC after the initial attempt.

modules/service/src/main/scala/lucuma/odb/service/ObscalcService.scala
```diff
@@ -0,0 +1,563 @@
...
+import skunk.*
+import skunk.circe.codec.json.*
+import skunk.codec.numeric._int8
+import skunk.codec.numeric.int4
+import skunk.data.Arr
+import skunk.implicits.*
+
+import scala.collection.immutable.SortedSet
+
+sealed trait ObscalcService[F[_]]:
```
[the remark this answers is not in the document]
swalker2m: That's a good idea.  The only issue is that I don't want the `Result` that the `requireServiceAccess` methods require.  I ended up wrapping the result in a `Success` only to `getOrElse` it and then blowup if not success:

```scala
def withTransactionalServiceAccess[A](fa: (ServiceAccess, Transaction[F], Services[F]) ?=> F[A]): F[A] =
  services.useTransactionally:
    requireServiceAccess(fa.map(grackle.Result.success))
      .map(_.getOrElse(sys.error("Obscalc must be run with service user.")))
```

which is maybe a bit awkward.  There's probably a better idea?  Or maybe we could add something like that to `Services`?  The only options available now return a `Result` and but I just want it to fail immediately.

modules/service/src/main/scala/lucuma/odb/graphql/topic/OdbTopic.scala
```diff
@@ -49,11 +49,11 @@ object OdbTopic:
   )(update: PartialFunction[Array[String], Option[U]]): OdbTopic[E] =
     new OdbTopic[E]:
 
-      def updates[F[_]: Logger](
+      def updates[F[_]: Concurrent: Logger](
         s:         Session[F],
         maxQueued: Int
       ): Stream[F, U] =
-        s.channel(channel).listen(maxQueued).flatMap: n =>
+        s.channel(channel).listen(maxQueued).prefetchN[F](1024).flatMap: n =>
```
swalker2m, line 56: I have an issue here.  When the world is updated and the calculator has to run through all the observations there will be 1000s of events.  When the number of events surpasses `maxQueued` then the channel seems to hang and I can't shut the service down cleanly.  I added `prefetch` here and use `concurrently` instead of `merge` to combine the streams, but these only mitigate the problem.  For now I've just increased the `maxQueued` value to be greater than the number of observations we have.

tpolecat (approved):
Still looks good :+1:

PR #1761, opened 2025-05-06
Remove readMode and reads from the configuration

cquiroz:
Both values will be generated during sequence generation. Overrides are supported, though.


swalker2m (approved):
Looks great. :+1:

PR #1763, opened 2025-05-08
F2 Planned Time Estimates

swalker2m:
Adds definitions for F2 time estimate calculations.  I'll need to implement the sequence for testing.


modules/service/src/main/resources/db/migration/V0986__f2_planned_time.sql
```diff
@@ -0,0 +1,71 @@
+INSERT INTO t_time_estimate VALUES(
+  'f2_longslit_setup',
+  'Flamingos 2 Longslit Setup',
+  'Flamingos 2 longslit mode full setup cost',
+  'Flamingos2',
+  '20 minutes'
```
cquiroz, line 6: Where do these numbers come from?

cquiroz (approved):
LGTM

modules/service/src/main/resources/db/migration/V0986__f2_planned_time.sql
swalker2m: From the old `ocs` project.

swalker2m: Each instrument "data object" implements something called a `StepCalculator` where you can see the config change and the readout / write estimates.  The instruments define a setup time and reacq time as well, often depending on the configuration.

PR #1764, opened 2025-05-09
Update F2 spectroscopy modes

cquiroz:
I got a ping that the modes list had a bunch of new F2 modes
However, there are also changes to some existing modes
Would this break existing observations?


swalker2m (approved):
These are just observing mode instances so -- collections of values for the existing mode class so I think it's fine.

PR #1769, opened 2025-05-12
Add gmos imaging modes to the odb config

cquiroz:
First part. it adds the tables to hold the configurations and a parser for the tsv file provided by science

The next PR will update the schema to return the options


modules/service/src/main/resources/db/migration/V0989__imaging_matrix_gmos.sql
```diff
@@ -0,0 +1,47 @@
+
+-- Missing GmosS filters
+insert into t_gmos_south_filter values ('OVI', 'OVI', 'OVI_G0347', 684000, '[681600, 696500]', 'NarrowBand');
+insert into t_gmos_south_filter values ('OVIC', 'OVIC', 'OVI_G0349', 679000, '[679000, 680500]', 'NarrowBand');
```
swalker2m, line 4: In `ocs` it's `OVIC_G0348`.  Is this correct?

modules/phase0/src/main/resources/phase0/Phase0_Instrument_Matrix - Imaging.tsv
```diff
@@ -0,0 +1,16 @@
+instrument	FoV	filters	AO	capabilities	site
+GMOS-S	330	u,g,r,i,CaT,z,Z,Y,HeII,HeIIC,OIII,OIIIC,Ha,HaC,SII,OVIC,OVI,g+GG455,g+OG515,r+RG610,i+CaT,z+Cat	no		s
```
swalker2m, line 2: Didn't you update `z+Cat`?

PR #1774, opened 2025-05-14
Some changes to when groupEdit events are sent

toddburnside:
This PR makes 2 changes to when `groupEdit` events are sent:

1. If a group is moved out of another group, an event is sent for the old parent group. This allows Explore to know that the time estimates for the old group need to be updated.
2. Events with a `value` of null are not sent. These referred to the "root" group, but are not useful for Explore. When any group or observation is updated, we know that the times for the program (root group) need to be updated.


swalker2m (approved):
The code looks great.  It seems fine to me.  The `Program` is conceptually a group but it isn't a `Group` in the API sense.  I guess an alternative might be to add a filter in the `GroupEditInput` for program-level group events?  At any rate, :+1:

swalker2m:
Sorry, I grabbed migration number 0992 so it'll have to be renumbered.

PR #1775, opened 2025-05-15
Shortcut 4365: F2 placeholder sequence

swalker2m:
This PR is a snapshot of progress on developing F2 placeholder sequences.  It adds something basic for acquisition and science, but both need testing.  (`Acquisition` is closer to a complete version than `Science`.) I'd like to stop and offer this now though because:

* I need to refactor `ExecutionTestSupport` to split F2 and GMOS and that will generate a lot of noise
* I'd like to make some adjustments to `Flamingos2DynamicConfig` before we go much further

I think those should be done in separate PRs before completing sequence testing.

### DIFF
--- a/modules/sequence/src/main/scala/lucuma/odb/sequence/flamingos2/longslit/Acquisition.scala
+++ b/modules/sequence/src/main/scala/lucuma/odb/sequence/flamingos2/longslit/Acquisition.scala
@@ -42,7 +42,9 @@ import lucuma.odb.sequence.util.IndexTracker
 
 import java.util.UUID
 
-
+/**
+ * Flamingos 2 long slit acquisition.
+ */
 object Acquisition:
 
   case class Steps(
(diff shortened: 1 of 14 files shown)


modules/sequence/src/main/scala/lucuma/odb/sequence/flamingos2/longslit/Acquisition.scala
```diff
@@ -0,0 +1,195 @@
...
+      NonEmptyList.of(slit)
+
+  private object StepComputer extends SequenceState[F2] with Flamingos2InitialDynamicConfig:
+    def compute(
+      builtin:      Flamingos2Fpu,
+      exposureTime: TimeSpan
+    ): Steps =
+      eval:
+        for
+          _  <- F2.exposure    := exposureTime
```
cquiroz, line 67: Same acq as for longslit?

PR #1777, opened 2025-05-15
Shortcut 4365: track changes to core

swalker2m:
Updates to core v0.129.0 making matching changes to the API.


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -7043,19 +7043,24 @@ type Flamingos2Dynamic {
   lyotWheel: Flamingos2LyotWheel!
 
   """
-  Flamingos 2 mask, if any.
+  Flamingos 2 FPU, if any.
   """
-  mask: Flamingos2Mask
+  fpu: Flamingos2FpuMask
```
swalker2m, line 7048: I switched the field name here to match GMOS.  GMOS has the same builtin vs. custom mask so it seemed to me that they should have the same name.  If `mask` is a better name, we could switch them both.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -7043,19 +7043,24 @@ type Flamingos2Dynamic {
...
+  Flamingos 2 readout mode.
+  """
+  readoutMode: Flamingos2ReadoutMode!
 
   """
-  Reads, if specified.
+  Flamingos 2 reads.
   """
-  reads: Flamingos2Reads
+  reads: Flamingos2Reads!
```
swalker2m, line 7063: `readoutMode` and `reads` are now required.  They will be generated from the long slit config and sent back to the ODB from Observe as for everything else.

cquiroz (approved):
LGTM

PR #1779, opened 2025-05-15
Shortcut 4365: Use the proper F2 acquisition filter

swalker2m:
The acquisition filter to use is supposed to be the one (among the valid options `J`, `H` and `KShort`) which is closest to the science wavelength.


cquiroz (approved):
LGTM

PR #1780, opened 2025-05-16
Shortcut 4365: F2 acquisition sequence tests

swalker2m:
Adds F2 acquisition sequence tests.

This is basically a copy of the GMOS acquisition sequence tests.  I played with factoring out the commonality and configuring GMOS and F2 specific versions but:

* it's a lot of configuration
* they seem to work the same, down to the offset in `p` in the second step, but there may be more differences that come to light with testing
* other observing modes probably won't have the same acquisition

So, it might be best left for when we see another acquisition sequence.

### DIFF
--- a/modules/schema/src/main/scala/lucuma/odb/json/flamingos2.scala
+++ b/modules/schema/src/main/scala/lucuma/odb/json/flamingos2.scala
@@ -43,7 +43,9 @@ trait Flamingos2Codec:
         "useElectronicOffsetting" -> a.useElectronicOffseting.asJson
       )
 
-  given Decoder[Flamingos2FpuMask.Custom] =
+  // This needs an explicit name because the default, given_Decoder_Custom
+  // would clash with the default for the GMOS custom mask.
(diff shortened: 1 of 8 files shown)


modules/schema/src/main/scala/lucuma/odb/json/flamingos2.scala
```diff
@@ -43,7 +43,9 @@ trait Flamingos2Codec:
         "useElectronicOffsetting" -> a.useElectronicOffseting.asJson
       )
 
-  given Decoder[Flamingos2FpuMask.Custom] =
+  // This needs an explicit name because the default, given_Decoder_Custom
+  // would clash with the default for the GMOS custom mask.
+  given given_Decoder_Flamingos2FpuMask_Custom: Decoder[Flamingos2FpuMask.Custom] =
```
swalker2m, line 48: This is unfortunate, but came to light when I added F2 to the `all` syntax.

cquiroz (approved):
LGTM

PR #1782, opened 2025-05-16
Shortcut 4365: return the proper ITC imaging mode for F2 spectroscopy acquisition

swalker2m:


modules/service/src/main/scala/lucuma/odb/service/GeneratorParamsService.scala
```diff
@@ -118,13 +121,19 @@ object GeneratorParamsService {
...
     def asImaging(λ: Wavelength): InstrumentMode =
-      mode match {
+      mode match
+        case Flamingos2Spectroscopy(_, f, _)         =>
+          InstrumentMode.Flamingos2Imaging(lucuma.odb.sequence.flamingos2.longslit.Acquisition.toAcquisitionFilter(f))
+        case GmosNorthImaging(_, _)                  =>
+          mode
         case GmosNorthSpectroscopy(_, _, _, _, _, _) =>
           InstrumentMode.GmosNorthImaging(Acquisition.filter(GmosNorthFilter.acquisition, λ, _.wavelength), none)
+        case GmosSouthImaging(_, _)                  =>
```
cquiroz, line 131: Do we need to include F2 imaging?

swalker2m: Yes, we should. We'll need it when we actually do F2 imaging.

cquiroz (approved):
LGTM

PR #1785, opened 2025-05-19
Add migration for F2 smartgcal

cquiroz:
Somehow I completly missed this

### DIFF
--- a/modules/smartgcal/src/main/scala/lucuma/odb/smartgcal/parsers/gcal.scala
+++ b/modules/smartgcal/src/main/scala/lucuma/odb/smartgcal/parsers/gcal.scala
@@ -55,6 +55,7 @@ trait GcalParsers:
       "ND2.0"        -> GcalFilter.Nd20,
       "ND3.0"        -> GcalFilter.Nd30,
       "ND4.0"        -> GcalFilter.Nd40,
+      "ND4.5"        -> GcalFilter.Nd45,
       "ND4-5"        -> GcalFilter.Nd45,
       "ND5.0"        -> GcalFilter.Nd45, // 5.0 was removed
       "GMOS balance" -> GcalFilter.Gmos,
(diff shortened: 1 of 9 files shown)


modules/smartgcal/src/main/scala/lucuma/odb/smartgcal/parsers/gcal.scala
```diff
@@ -55,6 +55,7 @@ trait GcalParsers:
       "ND2.0"        -> GcalFilter.Nd20,
       "ND3.0"        -> GcalFilter.Nd30,
       "ND4.0"        -> GcalFilter.Nd40,
+      "ND4.5"        -> GcalFilter.Nd45,
```
cquiroz, line 58: F2 tables use `NDA4.5` but there is a test flagging this case as a failure
Should I disable the test?

swalker2m: I'm not sure.  Which test is failing?

swalker2m:
This looks great.  There's a `FileReaderSuite` that just makes sure that all the file entries parse.  It only does GMOS now but I think it'd probably good to update that with F2 if possible.

modules/smartgcal/src/main/scala/lucuma/odb/smartgcal/parsers/gcal.scala
cquiroz: https://github.com/gemini-hlsw/lucuma-odb/blob/main/modules/smartgcal/src/test/scala/lucuma/odb/smartgcal/parsers/GcalParsersSuite.scala#L70

swalker2m: The GMOS files use `ND4-5`.  I think we maybe just need to add a line for `ND4.5` after this one: https://github.com/gemini-hlsw/lucuma-odb/blob/999c13dd078855b8bd41c199d7cdd4ce582d04e7/modules/smartgcal/src/main/scala/lucuma/odb/smartgcal/parsers/gcal.scala#L58

swalker2m (approved):
Looks great. :+1:

PR #1786, opened 2025-05-19
Shortcut 4365: Hook up F2 AGS

swalker2m:
We were missing the F2 AGS params in the `GuideService`.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/GuideService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/GuideService.scala
@@ -22,6 +22,7 @@ import io.circe.syntax.*
 import lucuma.ags
 import lucuma.ags.*
 import lucuma.catalog.votable.*
+import lucuma.core.enums.Flamingos2LyotWheel
 import lucuma.core.enums.GuideProbe
 import lucuma.core.enums.GuideSpeed
 import lucuma.core.enums.PortDisposition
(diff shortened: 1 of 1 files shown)


modules/service/src/main/scala/lucuma/odb/service/GuideService.scala
```diff
@@ -276,8 +278,8 @@ object GuideService {
         (Site.GN, AgsParams.GmosAgsParams(mode.fpu.asLeft.some, PortDisposition.Side), mode.centralWavelength)
       case mode: gmos.longslit.Config.GmosSouth =>
         (Site.GS, AgsParams.GmosAgsParams(mode.fpu.asRight.some, PortDisposition.Side), mode.centralWavelength)
-      case mode: flamingos2.longslit.Config =>
-        ???
+      case mode: flamingos2.longslit.Config     =>
+        (Site.GS, AgsParams.Flamingos2AgsParams(Flamingos2LyotWheel.F16, Flamingos2FpuMask.Builtin(mode.fpu), PortDisposition.Side), mode.filter.wavelength)
```
rpiaggio, line 282: Are these values a temprary measure?

cquiroz: we always use F16 lyot, the other options are for Gmes + F2 which we never use

cquiroz: Eventually we neeed to parametrize the port

cquiroz (approved):
LGTM

PR #1788, opened 2025-05-23
Support imaging requirements in the odb

cquiroz:
This pull request includes the changes required to define science requirements for imaging of observations. This should generally apply to all imaging modes but is focused for now on GMOS.


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -4301,14 +4301,20 @@ Edit science requirements
 """
 input ScienceRequirementsInput {
   """
-  The mode field must be either specified or skipped altogether.  It cannot be unset with a null value.
+  ExposureTimeMode, which may be unset by assigning a null value, or ignored by
+  skipping it altogether.
   """
-  mode: ScienceMode
```
cquiroz, line 4306: `mode` is removed, this matches other inputs. We don't set `mode` instead it is inferred depending on whether `spectroscopy` or `imaging` is set

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -9173,6 +9173,102 @@ type SpectroscopyConfigOptionFlamingos2 {
...
+  combinedFilters, which may be unset by assigning a null value, or ignored by
+  skipping it altogether.
+  """
+  combinedFilters: Boolean
+
+  """
+  gmosNorth, which may be unset by assigning a null value, or ignored by
+  skipping it altogether.
+  """
+  gmosNorth: ImagingScienceRequirementsGmosNorthInput
```
cquiroz, line 9209: This is debatable, whether we should include mode specific properties in the requirements.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -10993,9 +11089,19 @@ type ScienceRequirements {
   mode: ScienceMode!
 
   """
-  Spectroscopy requirements
+  Requested exposure time mode.
   """
-  spectroscopy: SpectroscopyScienceRequirements!
+  exposureTimeMode: ExposureTimeMode
```
cquiroz, line 11094: `exposureTimeMode` was in `spectroscopy` but it is now on `ScienceRequirements`. This matches the table structure, where the mode is shared between the science modes

modules/service/src/main/resources/db/migration/V0995__imaging_science_requirements.sql
```diff
@@ -0,0 +1,142 @@
...
+);
+COMMENT ON TABLE t_imaging_requirements_gmos_north IS 'GMOS North imaging requirements';
+
+-- A view to repackage the filters as an array value so Grackle can digest
+-- it. Seems overkill but well.
+CREATE VIEW v_imaging_requirements_gmos_north AS
+  SELECT
+    c_observation_id,
+    -- keep insertion order
+    array_remove(array_agg(c_filter), NULL) AS c_filters
```
cquiroz, line 30: Trick to make grackle support scalar arrays

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
toddburnside: Having this here to allow the UI to retain row selections between visits/page refreshes will also mean that 2 people looking at the same observation will be changing the selections for each other. I'm not sure if that is good or bad. It seems a little weird, but then I guess editing the requirements changes the list of available observations and could have the side effect of changing the selections.

We will have to maintain this for each new imaging mode, though.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
toddburnside: We'll have to change our decoders some, but we can work around this.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
cquiroz: I agree it is debatable. Well, the main use is to keep the selection between reloads, but maybe that's not that important?
We could remove it

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
cquiroz: Yes I'll make the adjustments in explore

toddburnside: I can do it. I'm working in explore, anyway.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
cquiroz: After some discussions we decided to remove those fields

modules/service/src/main/scala/lucuma/odb/util/GmosCodecs.scala
```diff
@@ -61,6 +63,9 @@ trait GmosCodecs {
   val gmos_north_detector: Codec[GmosNorthDetector] =
     enumerated(Type.varchar)
 
+  val _gmos_north_filter: Codec[Arr[GmosNorthFilter]] =
```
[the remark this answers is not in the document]
cquiroz: not now but the config will use them

modules/service/src/main/scala/lucuma/odb/graphql/BaseMapping.scala
```diff
@@ -201,6 +201,9 @@ trait BaseMapping[F[_]]
   lazy val ImagingConfigOptionGmosNorthType        = schema.ref("ImagingConfigOptionGmosNorth")
   lazy val ImagingConfigOptionGmosSouthType        = schema.ref("ImagingConfigOptionGmosSouth")
   lazy val ImagingConfigOptionType                 = schema.ref("ImagingConfigOption")
+  lazy val ImagingGmosNorthScienceRequirementsType = schema.ref("ImagingScienceRequirementsGmosNorth")
```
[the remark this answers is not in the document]
cquiroz: I'll remove them

modules/service/src/main/scala/lucuma/odb/graphql/input/ImagingScienceRequirementsInput.scala
```diff
@@ -0,0 +1,47 @@
...
+import lucuma.odb.data.Nullable
+import lucuma.odb.graphql.binding.BooleanBinding
+import lucuma.odb.graphql.binding.Matcher
+import lucuma.odb.graphql.binding.ObjectFieldsBinding
+
+case class ImagingScienceRequirementsInput(
+  minimumFov:      Nullable[Angle],
+  narrowFilters:   Nullable[Boolean],
+  broadFilters:    Nullable[Boolean],
+  combinedFilters: Nullable[Boolean]
```
swalker2m, line 19: If the mode is imaging, can these be `null`?

swalker2m (approved):
What a lot of work!  It looks great to me. :+1:

modules/service/src/main/resources/db/migration/V0995__imaging_science_requirements.sql
```diff
@@ -0,0 +1,77 @@
+-- Add imaging science requirements tables
+ALTER TABLE t_observation
+  ADD COLUMN c_img_minimum_fov      d_angle_µas null default null,
+  ADD COLUMN c_img_narrow_filters   boolean     null default null,
+  ADD COLUMN c_img_broad_filters    boolean     null default null,
+  ADD COLUMN c_img_combined_filters boolean     null default null;
```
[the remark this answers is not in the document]
cquiroz: Right, my thinking was that if spec these are null. We could do some checks on a trigger

modules/service/src/main/scala/lucuma/odb/graphql/input/ImagingScienceRequirementsInput.scala
swalker2m: If you cannot have `null` values when imaging, it seems like this should be `Option` rather than `Nullable`.  Same over in `SpectroscopyScienceRequirementsInput`. Or is the idea to assign it to `null` in order to go back to whatever the default was?

modules/service/src/main/scala/lucuma/odb/graphql/mapping/ImagingScienceRequirementsMapping.scala
```diff
@@ -0,0 +1,21 @@
...
+import grackle.skunk.SkunkMapping
+import lucuma.odb.graphql.table.ObservationView
+
+trait ImagingScienceRequirementsMapping[F[_]] extends ObservationView[F]:
+
+  import ObservationView.ScienceRequirements.Imaging
+
+  lazy val ImagingScienceRequirementsMapping: ObjectMapping =
+    ObjectMapping(ImagingScienceRequirementsType)(
+      SqlField("id", ObservationView.Id, key = true, hidden = true),
```
swalker2m: How did you resolve the `"imaging": null` issue?  Is it still giving us:

```
  "imaging": {
     "minimumFov": null,
     ...
  }
```

modules/service/src/main/resources/db/migration/V0995__imaging_science_requirements.sql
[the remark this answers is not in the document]
swalker2m: I see the `science_mode_consistency` check below so I'll resolve this.

swalker2m (approved):
I had a minor question about `Nullable` vs `Option` but I don't think it matters.  This still looks great to me. :+1:

modules/service/src/main/resources/db/migration/V0995__imaging_science_requirements.sql
```diff
@@ -0,0 +1,150 @@
...
+    -- If imaging fields exist, set mode to imaging
+    NEW.c_science_mode := 'imaging';
+  ELSE
+    -- If no fields are set, set mode to null
+    NEW.c_science_mode := NULL;
+  END IF;
+
+  RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
```
swalker2m, line 79: We might consider making `c_science_mode` a `GENERATED` column if it is just a function of the other columns like this?

modules/service/src/main/resources/db/migration/V0995__imaging_science_requirements.sql
```diff
@@ -1,93 +1,102 @@
...
+  spec_wavelength          d_wavelength_pm,
+  spec_resolution          integer,
+  spec_wavelength_coverage d_wavelength_pm,
+  spec_focal_plane         d_tag,
+  spec_focal_plane_angle   d_angle_µas,
+  spec_capability          d_tag
+) RETURNS boolean AS $$
+BEGIN
+  RETURN num_nulls(spec_wavelength, spec_resolution, spec_wavelength_coverage,
+                   spec_focal_plane, spec_focal_plane_angle, spec_capability) < 6;
```
swalker2m, line 37: I expected it to return `num_nulls(....) > 0` means spectroscopy mode?  Am I reading it wrong?

cquiroz: you're right and it is cleaner just `>0`

PR #1790, opened 2025-06-03
Shortcut 4434: obscalc5

swalker2m:
This (draft) update to the background `Obscalc` service adds `ObservationWorkflow` to the data computed and stored by the service.  The idea is to make workflow accessible via ordinary database mapping for the UI / API instead of requiring a potentially expensive effect computation on demand.  This would be in addition to the normal on-demand workflow calculation used internally to determine access control.

At this point, there is a single new `calculatedWorkflow` field on `Observation`.  Given fast access to the workflow state, it can be added to observation filters as well.  Also, we may not need the existing `workflow` `Observation` field and the `observationsByWorkflow` query in the API?

Additional comments in-line.


modules/obscalc/src/main/scala/lucuma/odb/obscalc/Main.scala
```diff
@@ -227,12 +233,14 @@ object CalcMain extends MainParams:
...
       enums <- Resource.eval(pool.use(Enums.load))
 
+      http  <- c.httpClientResource
       itc   <- c.itcClient
       ptc   <- Resource.eval(pool.use(TimeEstimateCalculatorImplementation.fromSession(_, enums)))
 
       t     <- topic(pool)
       user  <- Resource.eval(serviceUser[F](c))
-      o     <- runObscalcDaemon(c.database.maxObscalcConnections, c.commitHash, c.obscalcPoll, itc, ptc, t, pool.evalMap(services(user, enums)))
+      mapping = OdbMapping.forObscalc(pool, SkunkMonitor.noopMonitor[F], user, c.commitHash, enums, http)
```
swalker2m, line 242: We need a GraphQL mapping since `Obscalc` now computes the observation workflow (which does a nested GraphQL call).  For that reason, the service user code is shuffled around a bit to make a user available here to pass to the mapping.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -9678,6 +9678,8 @@ type Observation {
   configurationRequests: [ConfigurationRequest!]!
 
   workflow: ObservationWorkflow!
+
+  calculatedWorkflow: CalculatedObservationWorkflow
```
swalker2m, line 9682: I don't know whether we need the original workflow computation in the API?  I wonder if we can get away with just calling it `workflow` and returning a `CalculatedObservationWorkflow`.

modules/service/src/main/scala/lucuma/odb/graphql/OdbMapping.scala
```diff
@@ -641,4 +643,36 @@ object OdbMapping {
...
+      monitor,
+      user,
+      null,        // Topics[F]
+      null,        // ItcClient[F]
+      commitHash,
+      Set.empty,   // GOA users
+      enums,
+      null,        // TimeEstimateCalculatorImplementation.ForInstrumentMode
+      httpClient,
+      null,        // Config.Email
```
swalker2m, line 673: I just pass `null` for things that we don't need for `Obscalc`.

modules/service/src/main/scala/lucuma/odb/service/ObscalcService.scala
```diff
@@ -196,28 +206,43 @@ object ObscalcService:
...
+
+        def workflow(
+          itc: Option[ItcService.AsterismResults],
+          dig: Option[ExecutionDigest]
+        ): F[ObservationWorkflow] =
+          services
+            .transactionally:
+              observationWorkflowService.getCalculatedWorkflow(pending.observationId, itc, dig.map(_.science.executionState))
+            .flatMap: r =>
+              r.toOption.fold(Logger[F].warn(s"Failure calculating workflow: $r").as(UndefinedWorkflow))(_.pure[F])
```
swalker2m, line 227: I'm unsure about this.  The workflow service returns an `ObservationWorkflow` in a `Result` wrapper.  I'm not sure when the `Result` might be anything other than success (particularly since no ITC call is being made in the `getCalculatedWorkflow` version).

modules/service/src/main/scala/lucuma/odb/service/ObservationWorkflowService.scala
```diff
@@ -494,37 +517,48 @@ object ObservationWorkflowService {
...
+        execs: Map[Observation.Id, ExecutionState]
+      ): Result[Map[Observation.Id, ObservationWorkflow]] =
+        infos
+          .toList
+          .traverse: (oid, info) =>
+            val obsErrors = errs.get(oid).toList.flatMap(_.toList)
+            workflowStateAndTransitions(info, execs.get(oid), obsErrors.map(_.code))
+              .map: (s, ss) =>
+                 oid -> ObservationWorkflow(s, ss, obsErrors)
+          .map(_.toMap)
```
swalker2m, line 532: This bit was separated from `getWorkflows` so that it could be shared by the `Obscalc` version.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
tpolecat: That seems fine to me. Same with removing the top-level query.

modules/service/src/main/scala/lucuma/odb/service/ConfigurationService.scala
```diff
@@ -162,13 +162,16 @@ object ConfigurationService {
...
     ): ResultT[F, (Map[Program.Id, List[ConfigurationRequest]], Map[(Program.Id, Observation.Id), Configuration])] =
-      ResultT:
-        services.runGraphQLQuery(Queries.RequestsAndConfigurations(pids, oids)).map: r =>
-          r.flatMap: json =>            
-            import Queries.RequestsAndConfigurations.given
-            json.hcursor.as[Queries.RequestsAndConfigurations.Response] match
-              case Right(r)    => Result(r)
-              case Left(other) => Result.internalError(other.getMessage)
+      if pids.isEmpty && oids.isEmpty then
+        ResultT.pure((Map.empty[Program.Id, List[ConfigurationRequest]], Map.empty[(Program.Id, Observation.Id), Configuration]))
```
swalker2m, line 166: When testing with the full `dev` database contents, the obscalc workers would always hang here when `pids` and `oids` were empty.  I'm not sure why that is, but it seems like it is valid to skip the call when we know the result will be empty anyway?

PR #1794, opened 2025-06-05
Set idleConnectionTime for http client

rpiaggio:
Otherwise we get `[WARN ] org.http4s.ember.client.EmberClientBuilderCompanionPlatform - timeout (90 seconds) is >= idleConnectionTime (60 seconds). It is recommended to configure timeout < idleConnectionTime, or disable one of them explicitly by setting it to Duration.Inf.`

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/Config.scala
+++ b/modules/service/src/main/scala/lucuma/odb/Config.scala
@@ -62,7 +62,10 @@ case class Config(
   // People also send us their API keys. We need to be able to exchange them for [longer-lived] JWTs
   // via an API call to SSO, so we need an HTTP client for that.
   def httpClientResource[F[_]: Async: Network]: Resource[F, Client[F]] =
-    EmberClientBuilder.default[F].withTimeout(httpClient.timeout).build
+    EmberClientBuilder.default[F]
+      .withTimeout(httpClient.timeout)
+      .withIdleConnectionTime(httpClient.effectiveIdleConnectionTime)
+      .build
 
   // ITC client resource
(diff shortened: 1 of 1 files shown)


cmanto (approved):
LGTM

modules/service/src/main/scala/lucuma/odb/Config.scala
```diff
@@ -242,13 +245,19 @@ object Config:
...
+      if idleConnectionTime > timeout
+        then idleConnectionTime
+        else timeout.plus(1.second)
 
   object HttpClient:
     lazy val fromCiris: ConfigValue[Effect, HttpClient] = (
-      envOrProp("HTTP_CLIENT_TIMEOUT").as[Duration].default(45.seconds) // 45s is Ember's default
-    ).map(HttpClient.apply)
+      envOrProp("HTTP_CLIENT_TIMEOUT").as[Duration].default(45.seconds), // 45s is Ember's default
+      envOrProp("HTTP_CLIENT_IDLE_CONNECTION_TIME").as[Duration].default(60.seconds) // 60s is Ember's default
```
toddburnside, line 259: I wonder if, instead of having both be config parameters, we could just set the idle connection time to be 15 second more than the timeout. That way, whoever sets the config value wouldn't need to know that the idle connection time should be greater than the timeout.

rpiaggio: That's the behavior (but wit + 1s) via `effectiveIdleConnectionTime` if you don't explicitly set it.

PR #1795, opened 2025-06-05
Support ft roles

cquiroz:


modules/service/src/main/resources/db/migration/V0997__ft_mentor_reviewer.sql
```diff
@@ -0,0 +1,68 @@
+-- Add reviewer and mentor references to proposal table
+-- Each FT proposal can have designated reviewer and mentor
+ALTER TABLE t_proposal
+  ADD COLUMN c_reviewer_id d_program_user_id NULL DEFAULT NULL REFERENCES t_program_user(c_program_user_id) ON DELETE SET NULL,
+  ADD COLUMN c_mentor_id d_program_user_id NULL DEFAULT NULL REFERENCES t_program_user(c_program_user_id) ON DELETE SET NULL;
+
+-- Add constraint to prevent same user being both reviewer and mentor
+ALTER TABLE t_proposal
+ADD CONSTRAINT chk_reviewer_mentor_different
+CHECK (c_reviewer_id IS NULL OR c_mentor_id IS NULL OR c_reviewer_id != c_mentor_id);
```
toddburnside, line 10: Should there be a constraint ensuring that the reviewer and mentor are program users for the same program as the proposal?

cquiroz: This got me thinking for a while. I was thinking that program users have a program id, but you are right. As it is now, the reviewer ID can point to a program user of a different program. I will add a constraint.

PR #1796, opened 2025-06-06
Shortcut 4434: ExecutionDigest

swalker2m:
The next query to convert to the results of the background calculation is the `ExecutionDigest`.  In this case, the digest is already computed and stored in the `t_obscalc` table.  To add access to it via the GraphQL schema, I've added an effect handler that just reads the `ExecutionDigest` from the `t_obscalc` table along with any errors and returns it as Json, reporting warnings as before.

For now, I've added a `calculatedDigest` field to `Execution` as a sibling of the existing on-demand version, `digest`, from before.  When explore has been converted, I'll delete `digest` and rename `calculatedDigest` to `digest`.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/graphql/table/ObscalcTable.scala
+++ b/modules/service/src/main/scala/lucuma/odb/graphql/table/ObscalcTable.scala
@@ -6,7 +6,6 @@ package table
 
 import lucuma.odb.util.Codecs.*
 import skunk.circe.codec.all.*
-import skunk.codec.numeric._int8
 
 trait ObscalcTable[F[_]] extends BaseMapping[F]:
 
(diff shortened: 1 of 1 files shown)


modules/service/src/main/scala/lucuma/odb/graphql/mapping/ExecutionMapping.scala
```diff
@@ -58,7 +60,7 @@ trait ExecutionMapping[F[_]] extends ObservationEffectHandler[F]
     ObjectMapping(ExecutionType)(
       SqlField("id", ObservationView.Id, key = true, hidden = true),
       SqlField("programId", ObservationView.ProgramId, hidden = true),
-      EffectField("digest", digestHandler, List("id", "programId")),
+      SqlObject("digest", Join(ObservationView.Id, ObscalcView.Digest.Id)),
```
swalker2m, line 63: Here is where the old compute-on-demand effect handler is swapped out for a normal `SqlObject` mapping.

tpolecat:
I can't see a clean way to do this, so I guess my suggestion is **Alternative 3**, which will require collecting the ids from the parent cursors and doing a single subquery.

toddburnside (approved):
LGTM. The temporary calculatedDigest field is very helpful. 😄

PR #1797, opened 2025-06-06
Scala 3.7

toddburnside:
Scala 3.7 has a lot more `unused` warnings. The `Transaction` and access implicits can trigger these warnings. For cases where the implicit is specified in an interface, implementations of those interface methods do not trigger the warning (as of 3.7.1). However, other methods need to be marked with `@nowarn`.

This also found unused explicit method parameters in some places. Probably leftovers from refactoring?


cquiroz (approved):
LGTM, this looks like a lot of work

PR #1808, opened 2025-06-12
Add gmos imaging observing mode to the odb

cquiroz:
This pull request introduces support for GMOS North and South imaging modes in the GraphQL schema, database, and service layer. The changes include new types and inputs for imaging modes, updates to the database schema to accommodate imaging configurations, and modifications to service mappings for proper integration.

Includes:
* GraphQL Schema Updates:
* Database Schema Updates:
* Service Layer Updates:
* Multiple tests


modules/service/src/main/resources/db/migration/V0999__gmos_imaging_modes.sql
```diff
@@ -0,0 +1,163 @@
...
+COMMENT ON TABLE t_gmos_south_imaging IS 'GMOS South Imaging mode configuration';
+
+-- Normalized tables for filter arrays following the d_tag array pattern
+
+-- GMOS North Imaging Filters - normalized table for array support
+CREATE TABLE t_gmos_north_imaging_filter (
+  c_observation_id d_observation_id NOT NULL,
+  c_filter         d_tag            NOT NULL REFERENCES t_gmos_north_filter(c_tag),
+  PRIMARY KEY (c_observation_id, c_filter),
+  FOREIGN KEY (c_observation_id) REFERENCES t_observation(c_observation_id) ON DELETE CASCADE
```
swalker2m, line 50: I wonder if this should instead be a FK constraint to `t_gmos_north_imaging(c_observation_id)`?  We presumably don't want values hanging out in this filter table if the mode switches and there is no longer an entry for the observation in `t_gmos_north_imaging`.

cquiroz: Good point, I'll implement that

swalker2m (approved):
It looks great so far.  I was wondering why there were no updates to the `Generator`, to at least produce an error if an imaging sequence is requested.  Then I noticed that `ObservingMode` itself hasn't been updated with the new case so that must be for a future step?

PR #1809, opened 2025-06-12
Bump versions, switch Gaia adapter

rpiaggio:


toddburnside (approved):
Looks good to me. Thanks for making the gaia client.
Did you want to "mock" the gaiaClient rather than the httpClient in that test?

PR #1816, opened 2025-06-13
Shortcut 4365: denormalize step start/finish times

swalker2m:
Updates the step record to add the first/last event times.  This data is already in the execution event table, but is needed frequently for sequence generation so I decided it would be better to make accessing it fast and simple.

### DIFF
--- a/modules/service/src/main/resources/db/migration/V0998__step_record_update.sql
+++ b/modules/service/src/main/resources/db/migration/V0998__step_record_update.sql
@@ -1,11 +1,12 @@
--- Add an updated timestamp to the step record table.
+-- Add first and last event times to the step record table.  These come from
+-- t_execution_event but are needed repeatedly for sequence generation so we'll
+-- denormalize them here and keep it up to date with a trigger.
 ALTER TABLE t_step_record
   ADD COLUMN c_first_event_time timestamp NULL,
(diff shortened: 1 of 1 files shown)


modules/service/src/main/scala/lucuma/odb/graphql/mapping/StepRecordMapping.scala
```diff
@@ -38,24 +40,32 @@ trait StepRecordMapping[F[_]] extends StepRecordView[F]
...
-      SqlField("created",        StepRecordView.Created),
-      SqlField("executionState", StepRecordView.ExecutionState),
-      EffectField("interval",    intervalHandler, List("id")),
+      SqlField("id",              StepRecordView.Id, key = true),
+      SqlField("index",           StepRecordView.StepIndex),
+      SqlField("instrument",      StepRecordView.Instrument, discriminator = true),
+      SqlObject("atom",           Join(StepRecordView.AtomId, AtomRecordTable.Id)),
+      SqlField("created",         StepRecordView.Created),
+      SqlField("executionState",  StepRecordView.ExecutionState),
+      SqlField("_firstEventTime", StepRecordView.FirstEvent, hidden = true),
```
cquiroz, line 49: is it a convention for hidden fields to start with `_`?

cquiroz (approved):
LGTM

PR #1818, opened 2025-06-16
Remove piAffiliation from FastTurnaround proposals

cquiroz:


modules/service/src/main/scala/lucuma/odb/service/ProposalService.scala
```diff
@@ -207,8 +202,7 @@ object ProposalService {
             scienceSubtype.fold(().success) { s =>
               missingOrInvalidSplits(pid, s).asFailure.whenA(
                 splitsSum =!= 100 &&
-                ((s === ScienceSubtype.Classical)      ||
-                 (s === ScienceSubtype.FastTurnaround) ||
+                ((s === ScienceSubtype.Classical) ||
```
toddburnside, line 205: Obviously we no longer need to validate the partner splits for FT, but we do need to validate that the PI has a partner affiliation and that the affiliation matches on of the partners in the CfP.

Andy is creating a new story for API validations found here https://app.shortcut.com/lucuma/story/5269/explain-why-submit-proposal-button-is-inactive I don't see a new story yet, but maybe the FT validation could be added to that list?

cquiroz: I thought this was already validated. soat this moment the pi for FT could be on `HasNoPartner` type?

PR #1819, opened 2025-06-16
Trigger obsCalc on CfP and Configuration Request changes.

toddburnside:


modules/service/src/main/resources/db/migration/V1001__more_obscalc_triggers.sql
```diff
@@ -0,0 +1,98 @@
...
+  pid d_program_id
+)
+  LANGUAGE plpgsql AS $$
+DECLARE
+  obs_id d_observation_id;
+BEGIN
+  FOR obs_id IN
+    SELECT c_observation_id
+    FROM t_observation
+    WHERE c_program_id = pid
```
swalker2m, line 12: I suppose we could skip deleted observations.  If they are undeleted then it's an obs edit a calc is triggered for that.

swalker2m (approved):
Looks great. :+1:

PR #1822, opened 2025-06-18
Shortcut 5914: Add F2 centralWavelength

swalker2m:
Adds `centralWavelength` to the GraphQL API as required by Navigate.

### DIFF
--- a/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -7212,6 +7212,11 @@ type Flamingos2Dynamic {
   Flamingos 2 reads.
   """
   reads: Flamingos2Reads!
+
+  """
+  Central wavelength, which is taken from the filter wavelength.
+  """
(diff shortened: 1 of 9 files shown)


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -7212,6 +7212,11 @@ type Flamingos2Dynamic {
   Flamingos 2 reads.
   """
   reads: Flamingos2Reads!
+
+  """
+  Central wavelength, which is taken from the filter wavelength.
```
cquiroz, line 7217: is this for one step? l'm thinking of imaging where we have n filters

cquiroz (approved):
LGTM

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
swalker2m: Correct, it’s the instrument config for one particular step. If there are steps with different filters they would have corresponding different wavelengths.

PR #1830, opened 2025-06-23
Add support for f2 longslit offsets

cquiroz:


modules/service/src/main/scala/lucuma/odb/graphql/input/Flamingos2LongSlitInput.scala
```diff
@@ -66,15 +73,17 @@ object Flamingos2LongSlitInput {
           Flamingos2ReadModeBinding.Option("explicitReadMode", rReadMode),
           Flamingos2ReadsBinding.Option("explicitReads", rReads),
           Flamingos2DeckerBinding.Option("explicitDecker", rDecker),
-          Flamingos2ReadoutModeBinding.Option("explicitReadoutMode", rReadoutMode)
+          Flamingos2ReadoutModeBinding.Option("explicitReadoutMode", rReadoutMode),
+          OffsetInput.Binding.List.Option("explicitSpatialOffsets", rSpatialOffsets)
```
swalker2m, line 77: F2 long slit requires exactly four offset positions.  I think we should check and fail here if a list is provided that is anything other than four in size.  Can we do anything at the database level to guarantee that the data is correct as well?  One idea is to introduce an `Flamingos2LongSlitOffsets` type in the GraphQL schema with `a0`, `b0`, `b1`, and `a1` offset fields, and to store these explicitly in 8 columns `c_spatial_offset_a0_p`, `c_spatial_offset_a0_q` ... in the f2 long slit mode table but perhaps that is overkill.

swalker2m: Sequence generation is going to depend on finding four offset positions and will have to fail otherwise.

cquiroz: it can probably be done in the db in the check, make the regex accept either 0 or 4

swalker2m (approved):
Thanks for taking care of this!  Looks great. :+1:

PR #1831, opened 2025-06-23
remove inner `run`

tpolecat:
Grasping at straws here but maybe calling `run` inside `flatMap` breaks stack safety somehow.


swalker2m (approved):
Regardless, that looks better anyway. :+1:

PR #1839, opened 2025-06-24
Short Cut 5912: F2 acquisition exposure adjustments

swalker2m:
Generalizes a hack in the `ItcService` that adjusts acquisition exposure times when they are out of range or when the source is too bright.  Before this, it was using GMOS-specific values even for F2.


cquiroz (approved):
LGTM

PR #1843, opened 2025-06-25
SC-5657 TOO clone/update

tpolecat:
This completes the implementation of TOO targets, in the same sense that nonsidereal targets are complete (i.e., you can create, edit, and update them but they don't take part in workflow calculations).


swalker2m (approved):
Looks great.  :+1:

PR #1845, opened 2025-06-30
Include spatial offsets on F2 config

cquiroz:


modules/sequence/src/main/scala/lucuma/odb/sequence/flamingos2/longslit/Config.scala
```diff
@@ -33,7 +34,8 @@ case class Config private[longslit](
   defaultDecker: Flamingos2Decker,
   explicitDecker: Option[Flamingos2Decker],
   defaultReadoutMode: Flamingos2ReadoutMode,
-  explicitReadoutMode: Option[Flamingos2ReadoutMode]
+  explicitReadoutMode: Option[Flamingos2ReadoutMode],
+  explicitSpatialOffsets: Option[List[Offset]]
```
toddburnside, line 38: Could this be an `Option[NonEmptyList[Offset]]`? Or does the Option indicate something other than an empty list?

toddburnside (approved):
LGTM

modules/sequence/src/main/scala/lucuma/odb/sequence/flamingos2/longslit/Config.scala
swalker2m: We really want `None` or a `List` of exactly 4. At sequence generation time, I'll have to fail the config if it doesn't have 4 offset positions.

cquiroz: right at least IG2 and GMOS use that abba

PR #1846, opened 2025-06-30
Update to the new itc lib with multiple modes support

cquiroz:
In preparation to merge 
https://github.com/gemini-hlsw/lucuma-itc/pull/1020


modules/service/src/main/scala/lucuma/odb/service/ItcService.scala
```diff
@@ -343,24 +343,24 @@ object ItcService {
...
+                        ).asRight
+                      case Right(r) if r.times.focus.exposureTime > max =>
+                        r.copy(times = r.times.map(_.copy(exposureTime = max))).asRight
+                      case Right(r) if r.times.focus.exposureTime < min =>
+                        r.copy(times = r.times.map(_.copy(exposureTime = min))).asRight
+                      case other => other
+                .partitionErrors
+                .leftMap(convertErrors(targets))
+
+        targets.imaging.modes.head match
```
toddburnside, line 363: I can't follow this code in the diff. Why are we only using the `head` here? Is there anything that says all of the modes in the list have to be the same?

modules/service/src/main/scala/lucuma/odb/service/ItcService.scala
```diff
@@ -376,9 +376,12 @@ object ItcService {
       private def callRemoteItc(
         targets: ItcInput
       )(using NoTransaction[F]): F[Either[OdbError, AsterismResults]] =
+        // verify only one mode is passed
+        assert (targets.spectroscopy.modes.length == 1)
```
toddburnside, line 380: Why does this list need to be a length of 1? Is that enforced somewhere else?

toddburnside (approved):
LGTM with some questions.

modules/service/src/main/scala/lucuma/odb/service/ItcService.scala
cquiroz: this is the minimum to get it to compile in general, but going beyond science confirmed that we do acquisition with a single filter. I should add an assert as for spectroscopy

swalker2m: I'm confused, why is there more than one mode?  Imaging supports multiple filters but it is one mode right?

cquiroz: right, so it comes down to two aspects:
* old itc doesn't have the concept of multifilters, it will do n calculations with n results, we could model that as imaging modes with n filters or n imaging modes, internally it will have to do n calculations anyway
* from explore is more handy to pass multiple modes each with one filter, as we display a row per filter.

 I think it is useful in general to support requesting n modes but I could be convinced otherwise

cquiroz:
After some conversations, we will not be doing this change.

PR #1853, opened 2025-07-07
Shortcut 6054: handle `t_target` deletion better

swalker2m:
Fixes a mysterious issue in which `t_obscalc` rows would be deleted for calibration observations when the corresponding `t_target` row was deleted.  Targets are deleted by the calibration service, for example, when the observation time and/or duration changes significantly and causes a different target to be preferred.

`t_obscalc` references targets by id in its ITC results columns and had set a FK constraint with `ON DELETE CASCADE`.  In reality, the deletion of a `t_target` should not remove the `t_obscalc` row altogether but rather simply invalidate it so that it can be subsequently updated.


toddburnside (approved):
Great job finding this!

PR #1862, opened 2025-07-14
Allow not specifying filters when editing gmos imaging modes.

toddburnside:
When updating an imaging configuration, a non-empty of list was always required. But, it should be `ignorable`.

Someone who works with grackle Inputs more should probably verify I'm not doing anything wrong here...

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/graphql/input/GmosImagingInput.scala
+++ b/modules/service/src/main/scala/lucuma/odb/graphql/input/GmosImagingInput.scala
@@ -133,6 +133,14 @@ object GmosImagingInput:
   end Create
 
   object Edit:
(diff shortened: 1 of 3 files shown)


modules/service/src/main/scala/lucuma/odb/graphql/input/GmosImagingInput.scala
```diff
@@ -189,9 +205,10 @@ object GmosImagingInput:
       val Binding: Matcher[South] =
         SouthData.rmap:
           case (filters, exBin, exAmpReadMode, exAmpGain, exRoi, exSpatialOffsets) =>
-            Result(South(
-              filters.flatMap(NonEmptyList.fromList),
```
toddburnside, line 193: With the `toCreate` in ObservingModeInput removed, this was treating an empty list and a None the same. It now requires either None or a non empty list.

PR #1866, opened 2025-07-18
Add MultipleFiltersMode to the gmos imaging observing modes

toddburnside:


modules/service/src/main/scala/lucuma/odb/service/CalibrationConfigSubset.scala
```diff
@@ -123,6 +123,7 @@ object CalibrationConfigSubset:
         GmosImagingInput.Create.North(
           filters,
           GmosImagingInput.Create.Common(
+            none, // Do we need multipleFiltersMode here?
```
swalker2m, line 126: I see the similar comment for offsets.  I assume `none` means not explicit multiple filters mode which seems correct to me.

swalker2m (approved):
This looks great.  Are there instructions on how this impacts sequence generation?

modules/service/src/main/scala/lucuma/odb/service/CalibrationConfigSubset.scala
toddburnside: Right, this means no explicit multiple filters mode. It is probably correct, but Carlos says that science hasn't yet specified the behavior.

toddburnside:
> Are there instructions on how this impacts sequence generation?

I don't see anything specified in the epic anywhere. My understanding is the in `Grouped` mode, all the steps for the first filter will be together, followed by all for the second filter, etc. While in `Interleaved` mode, each step in the sequence will alternate between the filters. Of course, some filters will have more steps than others. 

Note that my understanding may not reflect reality...

PR #1869, opened 2025-07-22
Replace `digest` mapping with `calculatedDigest`

swalker2m:
Removes the unpredictable-delay version of `digest` in favor of the `calculatedDigest` version, which is then renamed to `digest`.  This will require some renaming changes in Explore I imagine.


toddburnside (approved):
Vaguely disappointing it doesn't remove 1000s of lines of code, but LGTM. 😆

toddburnside:
> This will require some renaming changes in Explore I imagine.

Yes, but it should be fairly minor. I'll do that this morning (my time) and let you know when I have something ready.

PR #1872, opened 2025-07-23
Move sequence generation to top-level `Query`

swalker2m:
Following up on an idea from @toddburnside, this PR moves the sequence generation from the `Execution` type to the top-level `Query`.  Because `Execution` is accessed from `Observation`, it was possible to write a query that would generate all steps of all atoms of all observations.  Moving sequence generation to `Query` and parameterizing it by observation id / reference, only one sequence at a time can be requested.


toddburnside (approved):
This seems good to me (although I can't contribute much about the AccessControl bit).

Since it only deprecates the other method of access through Observation, I don't see any reason you can't merge it at will. Hugo said he can work on the Observe side and I can fix up Explore and we'll tell you when we're done so you can remove it entirely.

PR #1874, opened 2025-07-23
Remove `observationsByWorkflowState`

swalker2m:
The subsequently deprecated `observationsByWorkflowState` query was added to address the service need to obtain all ready and ongoing observations.  Because the workflow state calculation can involve sequence generation, it can be slow and wasn't included in the normal `WhereObservation` filter.  Now there is a background "calculated" `workflow` field:

```
input WhereObservation {

  ...

  """
  Matches on the observation workflow state.
  """
  workflow: WhereCalculatedObservationWorkflow

  ...
}
```

It works with already-calculated workflow states, is appropriate for normal observation filtering and should be used instead.


hugo-vrijswijk:
Navigate still uses this query. I can make the change tomorrow to use the new where variables for observations. But it would be nice if this doesn't make it to odb-dev until then, so navigate-dev doesn't break.

PR #1875, opened 2025-07-24
Shortcut 5684: read binning from mode table

swalker2m:
Refactors the GMOS long slit mode selection to read the default binning from the `t_gmos_long_slit` table directly instead of recomputing it in Scala.  This not only simplifies the code considerably but also works around an apparent discrepancy between the Scala binning calculation and the database version.

Unfortunately I was stopped in the refactoring, for now, by the GMOS imaging mode.  We will need to compute and maintain the binning in the database the way it is done for long slit so that grouping by mode will work.  I can then take advantage of that to continue refactoring the mode selection.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/GmosLongSlitService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/GmosLongSlitService.scala
@@ -19,11 +19,9 @@ import lucuma.core.enums.GmosSouthGrating
 import lucuma.core.enums.GmosXBinning
 import lucuma.core.enums.GmosYBinning
 import lucuma.core.math.Wavelength
-import lucuma.core.model.ImageQuality
 import lucuma.core.model.Observation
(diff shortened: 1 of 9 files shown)


modules/service/src/main/scala/lucuma/odb/service/GmosLongSlitService.scala
```diff
@@ -38,11 +36,11 @@ trait GmosLongSlitService[F[_]] {
 
   def selectNorth(
     which: List[Observation.Id]
-  ): F[Map[Observation.Id, SourceProfile => GmosNorth]]
+  ): F[Map[Observation.Id, GmosNorth]]
```
swalker2m, line 39: The `SourceProfile` was part of the binning calculation.  Since we're reading it from the table directly now, it is no longer needed.

toddburnside (approved):
LGTM

PR #1885, opened 2025-07-28
Skip shortcut actions by default (for me)

swalker2m:
This is a copy of the [`lucuma-core` PR](https://github.com/gemini-hlsw/lucuma-core/pull/1148) that does the same thing.  Basically, turn off the automatic premature marking of Shortcut tasks as done (for me at least).  Credit goes to ChatGPT for writing this for me, but maybe there's a better way?

This should probably be the default and we _remove_ the label if we want it to mark the task done.  I'm content if it works that way for me at least though.


toddburnside (approved):
LGTM. I'll consider adding myself to the list. 😄

PR #1886, opened 2025-07-28
Shortcut 6298: CfP title override

swalker2m:
Adds a `title` property to the Call for Proposals creation and update properties. If set, the normal title computation is short-circuited to use the override.  The override is optional and nullable.

### DIFF
--- a/modules/service/src/main/resources/db/migration/V1016__cfp_title_override.sql
+++ b/modules/service/src/main/resources/db/migration/V1016__cfp_title_override.sql
@@ -1,6 +1,6 @@
 -- Add a nullable title override column.
 ALTER TABLE t_cfp
-  ADD COLUMN c_title_override text NULL DEFAULT NULL;
+  ADD COLUMN c_title_override text NULL DEFAULT NULL CHECK (c_title_override IS NULL OR length(c_title_override) > 0);
 
 -- Drop the v_cfp view which calls the title formatting function.
 DROP VIEW v_cfp;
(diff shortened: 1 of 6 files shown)


modules/service/src/main/resources/db/migration/V1016__cfp_title_override.sql
```diff
@@ -0,0 +1,87 @@
...
+) RETURNS text AS $$
+DECLARE
+  instrument_list text;
+  name text;
+BEGIN
+  SELECT c_name INTO name FROM t_cfp_type WHERE c_type = cfpType;
+  instrument_list := array_to_string(instruments, ', ');
+  IF (titleOverride IS NOT NULL) THEN
+    RETURN titleOverride;
+  ELSE
```
swalker2m, line 26: This function is the same as before, except for the addition of the `titleOverride` parameter and this `IF` that respects it.

modules/service/src/main/resources/db/migration/V1016__cfp_title_override.sql
```diff
@@ -0,0 +1,87 @@
...
+      LEFT JOIN
+        v_cfp_partner p ON c.c_cfp_id = p.c_cfp_id
+      GROUP BY
+        c.c_cfp_id
+    )
+    SELECT
+      c.*,
+      cte_instrument.c_instruments,
+      cte_partner.c_is_open,
+      format_cfp_title(c.c_type, c.c_semester, c.c_title_override, c.c_active_start, c.c_active_end, cte_instrument.c_instrument_names) AS c_title,
```
swalker2m, line 79: The view is the same as before, except for passing the title override into the format function.

PR #1888, opened 2025-07-30
Shortcut 4938: dataset chron update query

swalker2m:
Adds a query that can be used by the GOA to determine which datasets have been updated.  Or rather, it will be usable once I add the "WHERE" filtering on timestamp in the next installment.


toddburnside (approved):
LGTM

PR #1891, opened 2025-07-31
sc-5673 workflow updates

tpolecat:
In an attempt to only break things once I'm afraid this has turned into a bigger PR than I intended. In any case this updates the configuration and workflow logic to support:

- `Opportunity` targets
- GMOS imaging
- F2 longslit

It also generalizes a bunch of tests to operate on multiple target types and instrument modes. This caused a bit of carnage in the merge, sorry.

This contains breaking schema changes.

TODO:
- [x] rename migration files


modules/service/src/main/resources/db/migration/V9998__configuration_request_region.sql
```diff
@@ -0,0 +1,29 @@
...
+
+-- Replace view to pick up new columns.
+DROP VIEW v_configuration_request;
+CREATE VIEW v_configuration_request AS
+  SELECT 
+    *,
+    CASE WHEN c_reference_ra IS NOT NULL THEN c_configuration_request_id END AS c_reference_id,
+    CASE WHEN c_region_ra_arc_type IS NOT NULL THEN c_configuration_request_id END AS c_region_id,
+    CASE WHEN c_observing_mode_type = 'gmos_north_long_slit' THEN c_configuration_request_id END AS c_gmos_north_longslit_id,
+    CASE WHEN c_observing_mode_type = 'gmos_south_long_slit' THEN c_configuration_request_id END AS c_gmos_south_longslit_id,
```
swalker2m, line 26: We'll need synthetic ids for each mode, right?

modules/service/src/main/scala/lucuma/odb/service/GuideService.scala
```diff
@@ -598,15 +603,16 @@ object GuideService {
...
           positions     = getPositions(obsInfo.availabilityAngles, genInfo.offsets)
-          neededLists  <- ResultT.fromResult(
-                            neededPeriods.traverse(p =>
-                              buildAvailabilityList(p, obsInfo, genInfo, genInfo.centralWavelength, asterism, tracking, candidates, positions)
-                            )
-                          )
+          neededLists  <- ResultT.fromResult:
+                            neededPeriods.traverse: p =>
+                              tracking match
+                                case None => ???
```
swalker2m, line 614: Should this be a `Success` result as you did in `candidates` ?

swalker2m (approved):
Looks great to me.  :+1:

modules/service/src/main/resources/db/migration/V9998__configuration_request_region.sql
tpolecat: Correct.

modules/service/src/main/scala/lucuma/odb/service/GuideService.scala
tpolecat: Yes, good catch.

tpolecat:
Ok @todd it's finally up to date and building. Godspeed.

PR #1896, opened 2025-08-04
Shortcut 5099: execution event client id

swalker2m:
Adds an optional `clientId` field to event mutations.  The ID is prefixed with a `c-` and followed by an arbitrary UUID.  For example: `c-82a166b0-4852-46dd-aa79-942a3f377ead`.  If the ID is provided, it must be unique across all time.  If an existing event has already been added with the same ID, the new event is rejected with a `duplicated_event` error.


tpolecat (approved):
Looks perfect to me. 👍🏻

swalker2m:
> I think the idea is that observe uses this so it can safely retry on failure without duplicating events in the odb.

Right, it is an optional field so it shouldn't break existing code but its use would allow preventing duplicate events.

toddburnside:
> what if the same observe sends different events like start an stop an event?

It is not an id that identifies a client, such as observer, but a client provided id that identifies a particular event.

PR #1899, opened 2025-08-04
Validate CfP deadline when submitting a proposal.

toddburnside:
I need the deadline to include in the submission email. Since I needed to get, I figured I'd just add the submission validation as well. It turns out changing the validation for submission changes a lot of tests, too...

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/ProposalService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/ProposalService.scala
@@ -21,6 +21,7 @@ import lucuma.core.model.Program
 import lucuma.core.model.Semester
 import lucuma.core.model.User
 import lucuma.core.util.Enumerated
+import lucuma.core.util.Timestamp
 import lucuma.itc.client.ItcClient
 import lucuma.odb.data.*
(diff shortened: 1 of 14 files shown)


modules/service/src/main/scala/lucuma/odb/service/ProposalService.scala
```diff
@@ -567,15 +589,29 @@ object ProposalService {
...
-        LEFT JOIN t_cfp cfp
+        LEFT JOIN v_cfp cfp
           ON prop.c_cfp_id = cfp.c_cfp_id
+        LEFT JOIN t_program_user pi
+          ON prog.c_program_id = pi.c_program_id
+          AND pi.c_role = 'pi'
+        LEFT JOIN v_cfp_partner cfp_pi
+          ON cfp.c_cfp_id = cfp_pi.c_cfp_id
+          AND cfp_pi.c_partner = pi.c_partner
+        LEFT JOIN LATERAL
```
toddburnside, line 609: By using `LATERAL` to calculate the deadline, I am also able to use it above to check if it is past the deadline.

PR #1904, opened 2025-08-07
Rename offsets columns and remove support for explicit offsets

cquiroz:


modules/service/src/main/scala/lucuma/odb/graphql/mapping/GmosImagingMapping.scala
```diff
@@ -40,35 +40,25 @@ trait GmosImagingMapping[F[_]]
...
+      CursorFieldJson("offsets",
         cursor =>
           cursor
-            .field("spatialOffsetsString", None)
-            .flatMap(_.as[Option[String]].map(_.map(decodeSpatialOffsets)))
-            .map(_.getOrElse(defaultSpatialOffsetsJson)),
-        List("explicitSpatialOffsets", "defaultSpatialOffsets")
+            .field("offsetsString", None)
+            .flatMap(_.as[String].map(s => if (s.isEmpty) defaultOffsetsJson else decodeSpatialOffsets(s))),
+        List("offsetsString")
```
swalker2m, line 49: Should `offsetsString` also be common?

swalker2m (approved):
Seems good to me.  :+1:

modules/service/src/main/scala/lucuma/odb/graphql/mapping/GmosImagingMapping.scala
swalker2m: Hmm, I see it down in `GmosNorthImagingMapping`:

```
SqlField("offsetsString", GmosNorthImagingView.Offsets, hidden = true),
```

PR #1915, opened 2025-08-12
Use a noop for Honeycomb if there is no configuration

cquiroz:


swalker2m (approved):
Seems okay to me. :+1:

PR #1916, opened 2025-08-13
Shortcut 6465: Deleted target / generator params bug

swalker2m:
Andy reported a bug (in production 🎊)  wherein a sequence could not be produced for an observation because "observation is missing target".  An investigation revealed that the observation had three associated targets in `t_asterism_target` but that two of those targets were deleted.

I'd expected that the `v_generator_params` view would hide the deleted targets and everything would work.  I believe the issue though was the way that select from `t_observation` was joined with the targets:

```
LEFT JOIN t_asterism_target a
  ON o.c_observation_id = a.c_observation_id
LEFT JOIN t_target t
  ON  a.c_target_id  = t.c_target_id
  AND t.c_existence <> 'deleted'
```

That is, the LEFT JOIN with `t_asterism_target` was still producing a row per deleted target.  I switched the logic to do a sub-query that includes (I think) only non-deleted targets:

```
LEFT JOIN LATERAL (
  SELECT t.c_target_id,
         t.c_sid_rv,
         t.c_source_profile
    FROM t_asterism_target a
    INNER JOIN t_target t
      ON a.c_target_id = t.c_target_id
     AND t.c_existence = 'present'
   WHERE a.c_observation_id = o.c_observation_id
) t ON TRUE
```

so that a single observation row with `null` values for the target appears only when all targets are deleted or when there are no targets whatsoever.  It had to be `LATERAL` so that it could reference `o.c_observation_id`.

Someone who knows better than me might have an alternative suggestion?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -164,6 +164,7 @@ lazy val service = project
       "org.tpolecat"             %% "natchez-honeycomb"                  % natchezVersion,
       "org.tpolecat"             %% "natchez-http4s"                     % natchezHttp4sVersion,
       "org.tpolecat"             %% "natchez-log"                        % natchezVersion,
+      "org.tpolecat"             %% "natchez-noop"                       % natchezVersion,
       "org.tpolecat"             %% "skunk-core"                         % skunkVersion,
       "org.tpolecat"             %% "skunk-circe"                        % skunkVersion,
       "com.lihaoyi"              %% "pprint"                             % pprintVersion,
--- a/modules/calibrations/src/main/scala/lucuma/odb/calibrations/Main.scala
+++ b/modules/calibrations/src/main/scala/lucuma/odb/calibrations/Main.scala
@@ -26,9 +26,8 @@ import lucuma.odb.sequence.util.CommitHash
 import lucuma.odb.service.Services
 import lucuma.odb.service.Services.Syntax.*
 import lucuma.odb.service.UserService
-import natchez.EntryPoint
+import lucuma.odb.util.LucumaEntryPoint
 import natchez.Trace
-import natchez.honeycomb.Honeycomb
 import org.http4s.Credentials
 import org.http4s.client.Client
 import org.http4s.headers.Authorization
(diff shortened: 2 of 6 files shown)


modules/service/src/main/resources/db/migration/V1028__generator_params_view_update.sql
```diff
@@ -0,0 +1,46 @@
...
+LEFT JOIN LATERAL (
+  SELECT t.c_target_id,
+         t.c_sid_rv,
+         t.c_source_profile
+    FROM t_asterism_target a
+    INNER JOIN t_target t
+      ON a.c_target_id = t.c_target_id
+     AND t.c_existence = 'present'
+   WHERE a.c_observation_id = o.c_observation_id
+) t ON TRUE
```
swalker2m, line 43: Another alternative, maybe, would be a common table expression:

```
CREATE OR REPLACE VIEW v_generator_params AS
WITH cte_asterism_target AS (
  SELECT
    a.c_target_id,
    a.c_observation_id,
    t.c_sid_rv,
    t.c_source_profile
  FROM t_asterism_target a
  INNER JOIN t_target t ON a.c_target_id = t.c_target_id AND t.c_existence = 'present'
)
SELECT
  o.c_program_id,
  o.c_observation_id,
  o.c_calibration_role,
  o.c_image_quality,
  o.c_cloud_extinction,
  o.c_sky_background,
  o.c_water_vapor,
  o.c_air_mass_min,
  o.c_air_mass_max,
  o.c_hour_angle_min,
  o.c_hour_angle_max,
  o.c_exp_time_mode,
  o.c_etm_signal_to_noise,
  o.c_etm_signal_to_noise_at,
  o.c_etm_exp_time,
  o.c_etm_exp_count,
  o.c_observing_mode_type,
  o.c_science_band,
  o.c_declared_complete,
  o.c_acq_reset_time,
  t.c_target_id,
  t.c_sid_rv,
  t.c_source_profile
FROM
  t_observation o
LEFT JOIN cte_asterism_target t ON t.c_observation_id = o.c_observation_id
ORDER BY
  o.c_observation_id,
  t.c_target_id;
```

Would this be better here?  I think the CTE query is evaluated just once whereas the subquery would run for each row.

tpolecat: The query optimizer is pretty smart so there may not be much difference. You can do `EXPLAIN ANALYZE SELECT * FROM v_generator_params` and look at the query plan and see if there's any obvious difference. The plan depends on the table statistics so you would need to run it on a copy of the production db.

tpolecat (approved):
This fixes the bug so it seems good to me. 👍🏻

modules/service/src/main/resources/db/migration/V1028__generator_params_view_update.sql
swalker2m: Thanks.  I ran 20 iterations of both versions on the test database and they average out essentially the same.  If anything, the subquery `LATERAL` version was faster so I'll just go with that.

PR #1917, opened 2025-08-14
Catch execution digest foreign key violation

toddburnside:
This is caused by calibration observations being deleted before the execution digest is being written. At least that's the theory.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/ExecutionDigestService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/ExecutionDigestService.scala
@@ -7,6 +7,7 @@ import cats.data.NonEmptyList
 import cats.effect.Concurrent
 import cats.implicits.catsKernelOrderingForOrder
 import cats.syntax.applicative.*
+import cats.syntax.applicativeError.*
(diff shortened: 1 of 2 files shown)


modules/service/src/main/scala/lucuma/odb/service/ExecutionDigestService.scala
```diff
@@ -142,7 +144,11 @@ object ExecutionDigestService {
           digest.science.offsets.toList,
           digest.science.atomCount,
           digest.science.executionState
-        ).void
+        )
+        .void
+        .recoverWith:
+          case SqlState.ForeignKeyViolation(ex) =>
+            Logger[F].info(ex)("Failed to insert or update execution digest. Probably due to a deleted calibration observation.")
```
swalker2m, line 151: Perhaps we should add the `oid` in the message?

swalker2m (commented):
Thanks for handling this. :+1:  I'm guessing we'll hit this again in obscalc, but given the FK constraint and the `ON DELETE CASCADE` I don't think it will leave cruft in the table.  Clearly I haven't been thinking about observations getting hard-deleted.

PR #1922, opened 2025-08-21
Shortcut 6532: filter programs on cfp

swalker2m:
Two unrelated items:

* Adds a `call` (for "call for proposals") field to `WhereProposal`
* Cleans up a bunch of left over warnings from the recent Scala update.


tpolecat (approved):
👍🏻

PR #1924, opened 2025-08-21
Shortcut 6531: Handle `null` instrument

swalker2m:
An exception in the production logs revealed that we do not try to prevent recording a visit for an observation that is not associated with any instrument (i.e., has no observing mode).  In order to collect data, an observing mode is necessary.  This PR will filter out observations that have no associated instrument.


toddburnside (approved):
LGTM

PR #1943, opened 2025-09-01
Shortcut 6451: New GMOS Long Slit Sequence

swalker2m:
Currently the GMOS long slit sequence creates atoms which apply to a single wavelength dither + spatial offset.  Each atom consists of an arc, a flat and then _n_ science steps using that configuration.  This task updates the logic to combine multiple spatial offsets for each spectral dither into a single atom.  For example, if previously we had

```
atom dither offset
---- ------ ------
   0   0 nm     0"
   0   0 nm     0"
-
   1   5 nm    15"
   1   5 nm    15"
-
   2   0 nm   -15"
   2   0 nm   -15"
-
   3   5 nm     0"
   3   5 nm     0"
-
   4   0 nm    15"
   4   0 nm    15"
-
   5   5 nm   -15"
   5   5 nm   -15"
```

we would instead have

```
atom dither offset
---- ------ ------
   0   0 nm     0"
   0   0 nm    15"
   0   0 nm   -15"
-
   1   5 nm     0"
   1   5 nm    15"
   1   5 nm   -15"
-
   2   0 nm     0"
   2   0 nm    15"
   2   0 nm   -15"
-
   3   5 nm     0"
   3   5 nm    15"
   3   5 nm   -15"
```

This is (greatly) complicated by the need to recover from failed datasets, long delays after calibrations etc.  Unfortunately the logic is incomprehensible, but not much more so than before.


swalker2m:
> should the estimates of current programs be updated?

Yeah, I think the estimate should correspond to the current sequence.  In any case (famous last words), the time estimate shouldn't drastically change.

PR #1954, opened 2025-09-03
Shortcut 6261: GMOS imaging binning

swalker2m:
Updates the GMOS binning computation in the database to work for imaging mode as well as long slit.  This will enable adding GMOS imaging to the configuration grouping as was done in #1470 for GMOS long slit.

It also simplifies observing mode lookup because there's no longer a need to pass around `SourceProfile => ObservingMode` functions.  This was done previously because default GMOS spatial binning requires information about the `SourceProfile` (in order to determine the object size).  Instead, the default binning is calculated and stored directly in the database.

### DIFF
--- a/modules/service/src/test/scala/lucuma/odb/graphql/mutation/updateObservations.scala
+++ b/modules/service/src/test/scala/lucuma/odb/graphql/mutation/updateObservations.scala
@@ -10,14 +10,17 @@ import eu.timepit.refined.types.numeric.NonNegShort
 import io.circe.Json
 import io.circe.literal.*
 import io.circe.syntax.*
+import lucuma.core.enums.GmosBinning
 import lucuma.core.enums.ObservingModeType
 import lucuma.core.enums.ScienceBand
 import lucuma.core.enums.TimeAccountingCategory
 import lucuma.core.model.Group
+import lucuma.core.model.ImageQuality
 import lucuma.core.model.Observation
 import lucuma.core.model.Program
 import lucuma.core.model.Target
 import lucuma.core.model.User
+import lucuma.core.syntax.string.*
 import lucuma.core.syntax.timespan.*
 import lucuma.odb.graphql.input.AllocationInput
 import lucuma.odb.service.ObservationService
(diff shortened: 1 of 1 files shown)


modules/sequence/src/main/scala/lucuma/odb/sequence/flamingos2/longslit/Config.scala
```diff
@@ -98,18 +97,4 @@ object Config:
       DefaultFlamingos2ReadoutMode,
       explicitReadoutMode,
       explicitSpatialOffsets
-    )
-
-  def reconcile(a: Config, modes: List[ObservingMode]): Option[Config] =
```
swalker2m, line 103: Before, it was possible to compute distinct default binning values across targets in an asterism.  The binning calculation handles this now by selecting the smallest binning across all targets if there are multiple options.

swalker2m:
> I'm a bit lost, so we are still calculating binning from exposure time but instead of doing it online we store it on the db?

The GMOS default spatial binning calculation doesn't need the exposure time but rather the object size, which is determined by the source profile and the image quality.  The computation was already in the database but was only being used for GMOS long slit.

swalker2m:
> Can we remove passing SourceProfile => ObservingMode all over then

Yes definitely. I think I got them all and I don't think it would compile if any were left.  If I'm missing one though let me know.

modules/service/src/main/scala/lucuma/odb/service/ObservingModeServices.scala
```diff
@@ -22,7 +21,7 @@ sealed trait ObservingModeServices[F[_]] {
 
   def selectObservingMode(
     which: List[(Observation.Id, ObservingModeType)]
-  )(using Transaction[F], SuperUserAccess): F[Map[Observation.Id, SourceProfile => ObservingMode]]
+  )(using Transaction[F], SuperUserAccess): F[Map[Observation.Id, ObservingMode]]
```
swalker2m, line 24: No more `SourceProfile => ObservingMode`.  It's just an `ObservingMode` now.  This makes the `reconcile` concept moot as well so I could also remove that.

modules/service/src/main/resources/db/migration/V1032__gmos_binning_update.sql
```diff
@@ -0,0 +1,218 @@
...
+          iq, pixel_scale_µas, src_profile, sampling := 2.5
+        );
+
+      ELSE
+        RETURN;
+    END CASE;
+
+    -- Track minimum xbin and ybin values
+    min_xbin := LEAST(COALESCE(min_xbin, current_xbin), current_xbin);
+    min_ybin := LEAST(COALESCE(min_ybin, current_ybin), current_ybin);
```
swalker2m, line 148: Here is where we just take the minimum binning if the object sizes of the asterism result in distinct binning values.

PR #1969, opened 2025-09-09
Add target_disposition column

cquiroz:
As discussed, we'll add this column to the targets table to support blind offsets
The column is read by the mapping, but not exposed in the GraphQL schema.


modules/service/src/main/resources/db/migration/V1034__target_table_disposition.sql
```diff
@@ -0,0 +1,34 @@
...
+    else 'science'::e_target_disposition
+  end;
+
+alter table t_target
+alter column c_target_disposition set not null;
+
+-- Set default value
+alter table t_target
+alter column c_target_disposition set default 'science'::e_target_disposition;
+
```
swalker2m, line 17: How about adding a constraint that states something like this:

```sql
  (c_target_disposition = 'calibration' AND c_calibration_role IS NOT NULL)
  OR
  (c_target_disposition <> 'calibration' AND c_calibration_role IS NULL)
```

PR #1982, opened 2025-09-14
Shortcut 6589: cornucopia of steps issue

swalker2m:
Addresses the "cornucopia" issue (see shortcut description and diagnosis) by essentially deprecating the abandoned atom concept and making atom events optional.  Atoms are `NOT_STARTED`, `ONGOING` or `COMPLETED` but never `ABANDONED` (regardless of how many steps might be left on the table).  If we couple this with _simply no longer sending atom events_ it should solve the cornucopia issue.  The question for, I think, @rpiaggio is whether this is acceptable.

For now, the `ABANDONED` atom execution state will remain in the API but would be removed in a subsequent PR.  Atoms are only marked complete when an event associated with a _different_ atom arrives or a terminal sequence event for the observation arrives.  If neither of these occur, the atom remains `ONGOING`.  This has no impact on time accounting or sequence generation but might be an issue for the sequence displays in Explore and Observe?

### DIFF
--- a/modules/sequence/src/main/scala/lucuma/odb/data/StepExecutionState.scala
+++ b/modules/sequence/src/main/scala/lucuma/odb/data/StepExecutionState.scala
@@ -36,4 +36,7 @@ enum StepExecutionState(val tag: String) derives Enumerated:
       case Aborted | Stopped | Abandoned => completedFailure
       case Completed                     => completedSuccess
 
+  def isTerminal: Boolean =
+    fold(false, true, true)
+
 end StepExecutionState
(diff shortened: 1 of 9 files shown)


rpiaggio:
This is completely acceptable. We never check for `AtomExecutionState.Abandoned` in explore or observe. In fact, the only place we check for any `AtomExecutionState` at all is in explore and we only look whether last atom of the last visit is `Ongoing`, in which case we hide it so that we don't show it twice (it's already shown as `nextAtom`). This will continue behaving correctly.

We *DO* care about `Abandoned` steps, but if I understand correctly, that behavior is unchanged.

PR #1985, opened 2025-09-15
Shortcut 5099: make adding events idempotent

swalker2m:
At a suggestion from @rpiaggio, this PR updates the `add*Event` methods to simply return (instead of failing with a duplicate client id error) when provided a client id that matches an already existing event.


rpiaggio (approved):
LGTM

PR #1986, opened 2025-09-16
Shortcut 5099: Remove duplicate client id error

swalker2m:
As of #1985, we detect duplicate client ids when adding events but simply return the existing event instead of generating an error.  This PR removes the unused error.


rpiaggio (approved):
LGTM

PR #1988, opened 2025-09-16
Add blind offset field on the observation db

cquiroz:
This PR support for blind offsets, observations get a boolean flag whether they a blind offsets and then a blind offset target is added to `TargetEnvironment`. 

I added a new `BlindOffsetsService` to do the db operations and tests

### DIFF
--- a/modules/sequence/src/test/scala/lucuma/odb/sequence/data/ItcInputSuite.scala
+++ b/modules/sequence/src/test/scala/lucuma/odb/sequence/data/ItcInputSuite.scala
@@ -33,8 +33,8 @@ class ItcInputSuite extends FunSuite:
     )
 
   test("acquisition sequence uses blind offset target when available"):
-    val regularTarget = (Target.Id.fromLong(1L).get, arbitrary[TargetInput].sample.get, Option.empty[Timestamp])
-    val blindOffsetTarget = (Target.Id.fromLong(2L).get, arbitrary[TargetInput].sample.get, Option.empty[Timestamp])
+    val regularTarget = (Target.Id.fromLong(1L).get, arbitrary[TargetInput].sample.get, None)
+    val blindOffsetTarget = (Target.Id.fromLong(2L).get, arbitrary[TargetInput].sample.get, None)
 
     val itcInput = createItcInput(
       NonEmptyList.one(regularTarget),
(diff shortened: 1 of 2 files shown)


modules/service/src/main/resources/db/migration/V1036__blind_offset_asterism.sql
```diff
@@ -0,0 +1,162 @@
...
+  END IF;
+  RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+-- Create trigger to enforce the constraint
+CREATE TRIGGER trg_blind_offset_unique
+  BEFORE INSERT OR UPDATE ON t_asterism_target
+  FOR EACH ROW
+  EXECUTE FUNCTION check_blind_offset_unique();
```
swalker2m, line 32: Maybe this could be a `CONSTRAINT TRIGGER` that is `DEFERRABLE`.  Seems like at some point we could need to make a number of changes in a transaction that will be eventually but not immediately correct.

modules/service/src/main/resources/db/migration/V1036__blind_offset_asterism.sql
```diff
@@ -0,0 +1,162 @@
...
+LEFT JOIN LATERAL (
+  SELECT t.c_target_id,
+         t.c_sid_rv,
+         t.c_source_profile
+    FROM t_asterism_target a
+    INNER JOIN t_target t
+      ON a.c_target_id = t.c_target_id
+     AND t.c_existence = 'present'
+   WHERE a.c_observation_id = o.c_observation_id
+     -- Don't include blind offset for sequence generation
```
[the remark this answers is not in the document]
swalker2m: Good question, this may be tricky.  It seems like we need the blind offset target for the acquisition exposure time but need to exclude it for the science exposure time.  Maybe we need to include the target disposition in the generator parameters.

modules/service/src/main/resources/db/migration/V1036__blind_offset_asterism.sql
swalker2m, line 32: Maybe this could be a `CONSTRAINT TRIGGER` that is `DEFERRABLE`. Seems like at some point we could need to make a number of changes in a transaction that will be eventually but not immediately correct.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/TargetEnvironmentMapping.scala
```diff
@@ -116,6 +133,11 @@ trait TargetEnvironmentMapping[F[_]: Temporal]
       end   <- Elab.liftR(rEnd)
       env   <- Elab.env(AvailabilityStartParam -> start, AvailabilityEndParam -> end)
     } yield env
+
+    case (TargetEnvironmentType, "blindOffsetTarget", Nil) =>
```
swalker2m, line 137: Do we even need an elaborator in this case?  I would think it would work without one.

PR #2005, opened 2025-09-24
Shortcut 3790: illustrative test case

swalker2m:
This PR adds a couple of test cases showing what I expect is causing the problem reported in [Shortcut 3790](https://app.shortcut.com/lucuma/story/3790/twilight-calibration-observation-sequence#activity-6827).


cquiroz (approved):
LGTM, it seems we need to handle this client-side

PR #2025, opened 2025-10-03
core update

cquiroz:


modules/service/src/main/scala/lucuma/odb/util/Codecs.scala
```diff
@@ -91,6 +92,13 @@ trait Codecs {
       .array[A](ev.tag, s => ev.fromTag(s).toRight(s"${tpe.name}: no such element '$s'"), tpe)
       .imap[List[A]](_.flattenTo(List))(Arr.fromFoldable)
 
+  private def codecFromFormat[A](format: Format[String, A], tpe: Type): Codec[A] =
+    Codec.simple(
+      format.reverseGet,
+      s => format.getOption(s).toRight(s"Invalid: $s"),
+      tpe
+    )
```
swalker2m, line 100: Might be able to return

```scala
codecFromPrism[A](Prism[String, A](format.getOption)(format.reverseGet))
```

swalker2m (approved):
Looks great. :+1:

PR #2029, opened 2025-10-05
Add a field for f2 spectroscopy observations

cquiroz:
We need a new field to hold a telluric type field. This will be used to create calibrations
The field is stored as json as it may contain an extra parameter and it seems overkill to add another table for that

### DIFF
--- a/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -7981,7 +7981,7 @@ enum Flamingos2Reads {
 }
 
 """
-Telluric calibration tag
+Telluric calibration type
 """
 enum TelluricTag {
   HOT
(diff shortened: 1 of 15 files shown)


modules/service/src/main/resources/db/migration/V1041__flamingos2_longslit_telluric_type.sql
```diff
@@ -0,0 +1,11 @@
+
+ALTER TABLE t_flamingos_2_long_slit
+ADD COLUMN c_telluric_type jsonb NOT NULL DEFAULT '{"tag":"HOT","starTypes":null}'::jsonb;
```
swalker2m, line 3: Json fields always seem a bit frightening to me because they seem like they have maintenance implications. We do use them occasionally though.  I don't know what the best approach is though.  If we added a telluric tag enum and a tag field of that type, at least that bit would be easier to track?  I'm not opposed to this though, just unsure.

swalker2m: Are there any rules about which long slit observations can and cannot have a telluric type?

swalker2m (approved):
I think this looks great.  I raised the `jsonb` type for discussion but I'm fine with this as is.  :+1:

PR #2119, opened 2025-11-21
Startup Diagmostics 3

tpolecat:
This adds startup checks to ensure that every GraphQL enum is consistent with its corresponding Scala enum (if any).


swalker2m (approved):
Looks great.  :+1:

toddburnside:
It would be nice to have some convenience methods like `checkPostgresAndSchemaEnum` and `checkPostgresLookupTableAndSchemaEnum` (with better names) so we would only need to have one validation method call per enum.

PR #2346, opened 2026-04-09
Shortcut 8293: GHOST Observing Mode (Part 3)

swalker2m:
Another piecemeal GHOST update.  This one is concerned with hooking up the ITC.

* Creates an `ItcGhostIfu` result type to manage ITC results for GHOST. 
* Updates the `GeneratorParamsService` to create a `ScienceOnlySpectroscopy` input (renamed from `Igrins2Spectroscopy` in order to also use it for GHOST).
* Updates `ItcService` to obtain ITC results for GHOST. Notably, it doesn't actually call the ITC.  Instead it just strips out the time and count and pairs from the input and lines them up with the proper channel.  I think the ITC needs to provide red and blue results before we can actually call it?

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/ItcService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/ItcService.scala
@@ -36,12 +36,14 @@ import lucuma.core.model.Target
 import lucuma.core.util.TimeSpan
 import lucuma.itc.AsterismIntegrationTimes
 import lucuma.itc.IntegrationTime
+import lucuma.itc.ItcGhostDetector
 import lucuma.itc.TargetIntegrationTime
 import lucuma.itc.client.ClientCalculationResult
 import lucuma.itc.client.ImagingInput
 import lucuma.itc.client.InstrumentMode
 import lucuma.itc.client.ItcClient
 import lucuma.itc.client.SpectroscopyInput
+import lucuma.itc.client.SpectroscopyParameters
 import lucuma.odb.data.Itc
 import lucuma.odb.data.Md5Hash
 import lucuma.odb.data.OdbError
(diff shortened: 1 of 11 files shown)


modules/service/src/main/scala/lucuma/odb/service/ItcService.scala
```diff
@@ -385,6 +387,31 @@ object ItcService {
...
+        //   that has the time and count we need per channel.
+        def ghost(
+          ghost:   InstrumentMode.GhostSpectroscopy,
+          targets: NonEmptyList[ItcInput.TargetDefinition]
+        ): EitherT[F, OdbError, Itc] =
+          def resultSet(d: ItcGhostDetector): Zipper[Itc.Result] =
+            Zipper.fromNel:
+              targets.map: t =>
+                val it = IntegrationTime(d.timeAndCount.time, d.timeAndCount.count)
+                Itc.Result(t.targetId, it, none)
```
swalker2m, line 407: Note, S/N is always empty because we aren't actually calling the ITC.  We don't need it at this point for sequence generation since we're given the time and count.

modules/service/src/main/scala/lucuma/odb/service/ItcService.scala
```diff
@@ -385,6 +387,31 @@ object ItcService {
               .handleError: t =>
                 OdbError.RemoteServiceCallError(s"Error calling ITC service: ${t.getMessage}".some).asLeft
 
+        // A placeholder result for GHOST, at least for now.
+        //
+        // * The ITC accepts the red and blue exposure time mode data, but
+        //   only returns a single result set?  It isn't clear how to map this
+        //   back onto the red and blue channels.
```
toddburnside, line 394: It isn't clear to me right now exactly what the ITC should return for GHOST. Do we need S/N for each detector (which would  probably require some changes that would affect all instruments to accommodate this), or does the ITC just return a single S/N somehow derived from the 2 channels? I'm looking at the data types and trying to understand them better before I talk to Andy about this. Hopefully today...

toddburnside (approved):
LGTM

PR #2350, opened 2026-04-10
Calibrations trace

cquiroz:
This PR adds more tracing to the calibrations daemon.
Studying the traces I got I was able to identify an n+1 query situation and a fix is included


modules/obscalc/simulate-obscalc-event.sh
```diff
@@ -0,0 +1,35 @@
...
+
+usage() {
+  cat <<EOF
+Usage: $0 <observation-id>
+
+Triggers an obscalc recalculation by touching the observation time and setting it to now
+avoiding going through explore.
+Changing the time fires an DB trigger that invalidates obscalc, which in turn
+sends the ch_obscalc_update notification.
+Use it to e.g. test the calibration loop
```
swalker2m, line 12: Would it not be sufficient to just mark the calc state pending:

```
UPDATE t_obscalc
SET c_obscalc_state = 'pending',
    c_last_invalidation = now()
WHERE c_observation_id = '${OBS_ID}';
```

swalker2m (approved):
Looks good.  Maybe pinpoint the actual query change for the problem you mentioned?

modules/obscalc/simulate-obscalc-event.sh
cquiroz: I wasn't sure, but I guess you're right

modules/service/src/main/scala/lucuma/odb/service/CalibrationsUtils.scala
```diff
@@ -109,36 +112,47 @@ trait WorkflowStateQueries[F[_]: Monad: Services] {
...
-        AND c_obscalc_state = 'ready'
-    """.query(observation_workflow_state.opt)
-
-  private val WorkflowStateAnyQuery =
-    sql"""
-      SELECT c_workflow_state
-      FROM t_obscalc
-      WHERE c_observation_id = $observation_id
-    """.query(observation_workflow_state.opt)
+  private def workflowStates(
```
cquiroz, line 123: This is one change where we query workflow states for several oids at once rather than doing an n+1

modules/service/src/main/scala/lucuma/odb/service/CalibrationsUtils.scala
```diff
@@ -109,36 +112,47 @@ trait WorkflowStateQueries[F[_]: Monad: Services] {
...
+        val af = Statements.selectWorkflowStates(oids, onlyReady)
+        session
+          .prepareR(af.fragment.query(observation_id *: observation_workflow_state))
+          .use(_.stream(af.argument, 1024).compile.toList)
+          .map(_.map((oid, state) => oid -> state).toMap)
+
+  private def haveVisits(
+    oids: List[Observation.Id]
+  ): F[Set[Observation.Id]] =
+    NonEmptyList.fromList(oids) match
```
cquiroz, line 139: Also visits were queried an n + 1 fashion, now they can be done in one call

PR #2351, opened 2026-04-10
Fix GHOST graphs

toddburnside:
This works in the playground, but I'll wait until GHOST works properly in the ODB so I can test it more thoroughly before merging. But, I though I could get any PR feedback before then.


itc/model/src/main/scala/lucuma/itc/math.scala
```diff
@@ -10,7 +10,6 @@ import scala.math.*
 // This will not work for extremely small numbers but we don't expect such
 def roundToSignificantFigures(num: BigDecimal, n: Int): Double =
   if num == 0 then 0
-  else if num.isExactDouble then num.toDouble
```
toddburnside, line 13: This was breaking the MathSuite test I added because 11111.5 is an `exactDouble`. I think this was an optimization gone awry, so I removed it. (I also moved MathSuite so it will actually run! 😆 )

cquiroz (approved):
LGTM

PR #2360, opened 2026-04-12
More otel4s porting

cquiroz:
This PR ports most of the usages of natchez on odb to otel4s only remaining the parts related to skunk which is still on natchez

It is fairly mechanical, most of the larger changes are about setting up the metrics an tracing middleware for http4s


toddburnside (approved):
LGTM

PR #2364, opened 2026-04-13
Remove zprime as an option for acquisition filter for GN and GS

cquiroz:


modules/service/src/main/resources/db/migration/V1121__remove_zprime_acq_filter.sql
```diff
@@ -0,0 +1,17 @@
+-- reset exisiting observation with z' explicitly selected as an acquisition
+-- filter to null
+UPDATE t_gmos_north_long_slit
```
cquiroz, line 3: @swalker2m This is not actually necessary and maybe even a problem for existing sequences. Should we keep it?

swalker2m: I think it's fine.  Executed observations, and even just visited but not finished observations, have had their steps recorded.  Changing anything in the configuration in this way won't impact those steps.  I suppose if there are any recoded but as yet unexecuted acquisition steps with `ZPrime` we could change them to whatever the default otherwise would have been.  Hopefully that's not necessary.

swalker2m (approved):
Looks good. :+1:

PR #2366, opened 2026-04-13
Shortcut 8293: Fix mapping issue.

swalker2m:
Todd uncovered a Grackle mapping issue with ETMs that took me some time to fix.  Previously, the SQL produced by the mapping would filter out non-Ghost observing modes when Ghost-specific exposure time modes were included in a query.  There is a lot of churn evidenced in the change list, so I will focus on the important bits:

* Generalizes the `ExposureTimeModeView` such that the same columns can be applied to multiple different Postgres view definitions.

```scala
  class BaseExposureTimeModeView(
    viewName: String
  ) extends TableDef(viewName):

    val Id: ColumnRef            = col("c_exposure_time_mode_id", exposure_time_mode_id)
    val ObservationId: ColumnRef = col("c_observation_id",        observation_id)
    val Role: ColumnRef          = col("c_role",                  exposure_time_mode_role)
    ...

  object ExposureTimeModeView          extends BaseExposureTimeModeView("v_exposure_time_mode")
```

* Adds exposure time mode views specific to GHOST in order to separate the red and blue channels.  

```scala
  object GhostBlueExposureTimeModeView extends BaseExposureTimeModeView("v_ghost_blue_exposure_time_mode")
  object GhostRedExposureTimeModeView  extends BaseExposureTimeModeView("v_ghost_red_exposure_time_mode")
```

* Updates the `ExposureTimeModeMapping` to be path-based, accepting the corresponding view to use with the mapping.

```scala
  def etmMappings(
    base: Path,
    view: BaseExposureTimeModeView
  ): List[ObjectMapping] =
    val path = base / "exposureTimeMode"

    val TimeAndCount: ObjectMapping =
      ObjectMapping(path / "timeAndCount")(
        SqlField("id", view.TimeAndCount.SyntheticId, key = true, hidden = true),
        SqlObject("time"),
        SqlField("count", view.TimeAndCount.Count),
        SqlObject("at")
      )
    ...

  lazy val ExposureTimeModeMappings: List[ObjectMapping] =
    List(
      // Flamingos 2
      etmMappings(Flamingos2LongSlitType,            ExposureTimeModeView),
      etmMappings(Flamingos2LongSlitAcquisitionType, ExposureTimeModeView),

      // Ghost
      etmMappings(GhostIfuType / "blue",  GhostBlueExposureTimeModeView),
      etmMappings(GhostIfuType / "red",   GhostRedExposureTimeModeView),
      ...
```


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -8627,6 +8621,22 @@ input GhostDetectorConfigInput {
   explicitReadMode: GhostReadMode
 }
 
+input GhostArmInput {
```
swalker2m: I wanted to separate the ETM from the detector config, for a particular mapping reason.  As far as I can tell from looking at GHOST information, there are red and blue "arms" each of which has a detector.  So I used this term though it reads a bit weird tbh.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -8704,6 +8710,40 @@ type GhostDetectorConfig {
 
 }
 
+"""
+GHOST blue arm configuration.
+"""
+type GhostBlueArm {
```
swalker2m: So I think I need two different `TypeRef`s to facilitate the mapping so I defined `GhostBlueArm` and `GhostRedArm` though they are otherwise identical.  This seems unfortunate but I am thoroughly beaten. They share a `GhostDetector` type.  (The actual arms have their own detectors of course.)

modules/service/src/main/resources/db/migration/V1120__ghost.sql
```diff
@@ -0,0 +1,15 @@
+
+-- A view that facilitates the GHOST IFU mapping by adding embedded ids that are
+-- null or non-null depending on camera type.
+CREATE VIEW v_ghost_exposure_time_mode_link AS
+  SELECT
+     e.c_exposure_time_mode_id,
+     CASE WHEN g.c_blue_exposure_time_mode_id = e.c_exposure_time_mode_id THEN e.c_observation_id END AS c_blue_observation_id,
```
swalker2m: This is the kind of trick we play over and over: making a view in which we have nullable ids to make the Grackle `Join`s work.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
cquiroz: Not sure if it is better maybe `GhostFiberInput`?

modules/service/src/main/resources/db/migration/V1120__ghost.sql
cquiroz: Should we mention this on the readme?

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
cquiroz: Actually that is wrong fibers and detectors are not one to one

swalker2m:
I may be able to simplify this.  Converting to a draft.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
swalker2m: In the end, the view and mapping update were sufficient and I didn't need an extra layer of hierarchy here.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
swalker2m: I was able to merge red and blue as it was before.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -8606,13 +8606,14 @@ enum GhostIfu2FiberAgitator {
...
+input GhostDetectorInput {
 
   """
   Exposure time mode for the detector.  If not specified, the exposure time mode
-  of the observation's science requirements are used.
+  of the observation's science requirements are used.  Note, only TimeAndCount
+  exposure time modes are accepted at this time.
   """
-  timeAndCount: TimeAndCountExposureTimeModeInput
+  exposureTimeMode: ExposureTimeModeInput
```
swalker2m, line 8616: To remove one layer of complication and inconsistency, I switched back to `ExposureTimeMode` instead of requiring `TimeAndCount`.  The final ETM per channel is validated before creation to ensure it is in fact `TimeAndCount`. This is suboptimal but I tell myself it will be ready when ITC supports Ghost S/N mode.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -8663,12 +8664,13 @@ input GhostIfuInput {
 
 }
 
-type GhostDetectorConfig {
+
+type GhostDetector {
```
swalker2m, line 8668: Gratuitous. Sorry.

modules/service/src/main/resources/db/migration/V1120__ghost.sql
```diff
@@ -0,0 +1,19 @@
...
+-- A view that facilitates the GHOST IFU mapping by separating the blue and
+-- red channels.  Otherwise, the views are identical to v_exposure_time_mode.
+
+CREATE VIEW v_ghost_blue_exposure_time_mode AS
+  SELECT
+    e.*,
+    CASE WHEN e.c_exposure_time_mode = 'signal_to_noise' THEN e.c_exposure_time_mode_id END AS c_signal_to_noise_id,
+    CASE WHEN e.c_exposure_time_mode = 'time_and_count'  THEN e.c_exposure_time_mode_id END AS c_time_and_count_id
+  FROM t_exposure_time_mode e
+  INNER JOIN t_ghost_ifu g ON g.c_blue_exposure_time_mode_id = e.c_exposure_time_mode_id;
```
swalker2m, line 11: The `INNER JOIN` ensures that only GHOST blue channel ETMs are present in the view.

modules/service/src/main/scala/lucuma/odb/graphql/mapping/ExposureTimeModeMapping.scala
```diff
@@ -3,15 +3,73 @@
...
-      SqlField("role", ExposureTimeModeView.Role, hidden = true),
-      SqlObject("signalToNoise"),
-      SqlObject("timeAndCount")
-    )
\ No newline at end of file
+  def etmMappings(
+    base: Path,
+    view: BaseExposureTimeModeView
+  ): List[ObjectMapping] =
+    val path = base / "exposureTimeMode"
```
swalker2m, line 17: Switching to path-based with a corresponding view enables Grackle to align the `Join`s and produce the expected SQL.

modules/service/src/main/scala/lucuma/odb/graphql/OdbMapping.scala
```diff
@@ -300,6 +298,7 @@ object OdbMapping {
             GraphQLParser
               .defaultConfig
               .copy(maxInputValueDepth = 16)
+              .copy(terseError = false)
```
swalker2m, line 301: I'm not sure what, if anything, this does but if there's a way to get more error information out of Grackle it seems worth doing.  I could switch this back if there's any objection.

toddburnside (approved):
LGTM. Great job getting this to work. Sorry it was so painful.

PR #2374, opened 2026-04-15
Add pending phase i flags to the backend

cquiroz:


modules/service/src/main/resources/db/migration/V1122__add_pending_phase_i_flags.sql
```diff
@@ -0,0 +1,27 @@
...
+ALTER TABLE t_proposal ADD COLUMN c_aeon_multi_facility boolean DEFAULT false NOT NULL;
+
+-- JWST Synergy flag (for LP and Queue)
+ALTER TABLE t_proposal ADD COLUMN c_jwst_synergy boolean DEFAULT false NOT NULL;
+
+-- US Long Term flag (Queue only)
+ALTER TABLE t_proposal ADD COLUMN c_us_long_term boolean DEFAULT false NOT NULL;
+
+-- Consider for Band 3 flag (Queue only)
+ALTER TABLE t_proposal ADD COLUMN c_consider_for_band_3 boolean DEFAULT NULL;
```
swalker2m: If this is intended to be a 3-value flag (`UNSET`, `MAYBE_BAND3`, `NEVER_BAND3` or something), maybe we should add an enum type for it and default to the `UNSET` value.

modules/service/src/main/scala/lucuma/odb/graphql/input/ProposalTypeInput.scala
```diff
@@ -228,25 +258,33 @@ object ProposalTypeInput {
...
+    minPercentTime:     Option[IntPercent]                 = None,
+    minPercentTotal:    Nullable[IntPercent]               = Nullable.Null,
+    totalTime:          Nullable[TimeSpan]                 = Nullable.Null,
+    partnerSplits:      Nullable[Map[Partner, IntPercent]] = Nullable.Null,
+    reviewerId:         Nullable[ProgramUser.Id]           = Nullable.Null,
+    mentorId:           Nullable[ProgramUser.Id]           = Nullable.Null,
+    aeonMultiFacility:  Nullable[Boolean]                  = Nullable.Absent,
+    jwstSynergy:        Nullable[Boolean]                  = Nullable.Absent,
+    usLongTerm:         Nullable[Boolean]                  = Nullable.Absent,
+    considerForBand3:   Nullable[Boolean]                  = Nullable.Null
```
swalker2m: As is, `considerForBand3` is nullable but the other three don't seem nullable?  What would it mean to set `jwstSynergy: null` in an edit?  It looks like it will reset it to the default value?  Is that intended or it should just be an `Option[Boolean]` and it would ignore explicitly `null` values.

swalker2m (approved):
Can't have too many flags.  Are they really all independent?

modules/service/src/main/scala/lucuma/odb/graphql/input/ProposalTypeInput.scala
swalker2m: Also, the default is to set the `considerForBand3` to `null` in an edit?

cquiroz: all flags (except considerForBand3) default to false.

cquiroz: for `considerForBand3` we want to start with `null` such that users have to select an option before submitting. But after selection it cannot return to `null`

swalker2m: I'm probably reading it wrong, but it looked like if you don't include a `considerForBand3` field in your edit input it will reset to the default value.

cquiroz:
@swalker2m you were right, the way edits were not correct. I'm using now the style of `tooAction`. 
I also added a new enum for `considerForBand3` that I should move to core to use in explore too

modules/service/src/main/scala/lucuma/odb/graphql/input/ProposalTypeInput.scala
cquiroz: Does this test shows that case working?

https://github.com/gemini-hlsw/lucuma-odb/pull/2374/changes/d98fe431f2dec875ef2009639e6e2fafb31bb4fd#diff-a27cc3afd5f15365874f40955f501fe37227428b1c67b811c4fb400ac5a87ac1R1281

toddburnside (approved):
Seems fine to me.

modules/service/src/main/scala/lucuma/odb/graphql/input/ProposalTypeInput.scala
swalker2m: > Does this test shows that case working?

:+1: Perfect.

PR #2386, opened 2026-04-20
Add support for setting the f2 read mode on imaging

cquiroz:
Also some improvements to the schema stitching scripts


modules/service/src/main/scala/lucuma/odb/service/GeneratorParamsService.scala
```diff
@@ -343,11 +344,18 @@ object GeneratorParamsService {
 
           case f2 @ flamingos2.longslit.Config(disperser, filter, fpu, sci, acq, _, _, _, _, _, _, _, _) =>
             val sciMode   = InstrumentMode.Flamingos2Spectroscopy(sci, disperser, filter, fpu)
+
+            val readMode = acq.exposureTimeMode match
```
cquiroz: @swalker2m does this seem fine?

swalker2m: If we compute the value to send, why couldn't the ITC do the same thing?  Is there ever a case where we take it directly from user input?

swalker2m (commented):
This code looks great.  I'm just not clear on why we need to send the value at all if the ITC could do the same thing we're doing to come up with it.  Or is it about setting us up for further improvements in the future?

modules/service/src/main/scala/lucuma/odb/service/GeneratorParamsService.scala
cquiroz: I think for imaging we want to allow overrides, not sure if we want them for acq too.
a question for @andrewwstephens

andrewwstephens: For F2 imaging the best readmode is usually "Bright" but the ITC should allow manually setting it so that the users can decide for themselves.

swalker2m (approved):

PR #2388, opened 2026-04-21
Add field for telluric type to igrins2

cquiroz:
This is a mirror of the same features for flamingos 2. The telluric type is a parameter used for telluric target searches


swalker2m (approved):
Looks great. :+1:

PR #2391, opened 2026-04-21
Shortcut 8293: Add step count and slit viewing

swalker2m:
Adds step count and slit viewing camera exposure time to the GHOST configuration.  Step count is in the observing mode and used to generate the expected number of steps.  The slit viewing camera exposure time is optional and is taken from the observing mode and written into the GHOST static config.


modules/service/src/main/resources/db/migration/V1124__ghost.sql
```diff
@@ -0,0 +1,18 @@
+ALTER TABLE t_ghost_ifu
+  ADD COLUMN c_step_count                        int4     NOT NULL DEFAULT 1 CHECK (c_step_count > 0),
```
cquiroz, line 2: is this step count per ifu? SV camera time is independent of the IFU

swalker2m: It's the number of science steps in the sequence (each of which will have the prescribed time and count for the red and blue detectors).  The table is called `t_ghost_ifu` after the name of the observing mode.

cquiroz: Ok, I thought it was a per arm value, is that step count a paraemeter we calculate or provided by the user?

swalker2m: For now, provided by the user.  Someday, in theory, if we support S/N mode the ITC might figure out the optimal step count and per-step, per-detector time and count.

cquiroz (approved):
LGTM

PR #2392, opened 2026-04-21
Otel client spans

cquiroz:
This PR updates lucuma-graph-routes to support client parent span traces.

A second part are updates to the otel collector to allow web clients to send traces without including a token and with CORS supports. No LLMs were harmed writing the docker confe


toddburnside (approved):
LGTM

PR #2393, opened 2026-04-21
Add `stepIndex` and `stepCount` to `AtomDigest`

swalker2m:
This task comes from JIRA for some reason: https://noirlab.atlassian.net/browse/GSCHED-955

The scheduler team requested that the `AtomDigest` output include the step index of the first step of each atom along with the number of steps.


cquiroz (approved):
LGTM

PR #2394, opened 2026-04-21
Extend the tellurics service for igrins2

cquiroz:
It makes the calibration service generate tellurics for igrins2.
The PR also include some instrumentation and performance improvements

image


modules/service/src/main/scala/lucuma/odb/service/PerScienceObservationCalibrationsService.scala
```diff
@@ -257,9 +257,9 @@ object PerScienceObservationCalibrationsService:
                                     yield (c, deletable)
                                   else
                                     (List.empty, List.empty).pure[F]
-            // sync configuration on all deletable tellurics
-            allTellurics       <- findAllTelluricObservations(gid)
-            toSync             <- excludeFromDeletion(allTellurics, identity)
+            // Only sync existing tellurics that are note deleted or recreated.
+            toSync              = if (created.nonEmpty) List.empty
+                                  else existing.filterNot(deleted.contains)
```
swalker2m, line 262: If any were created then we ignore the existing ones?

swalker2m (approved):
Looks great, I think.  :+1:

modules/service/src/main/scala/lucuma/odb/service/PerScienceObservationCalibrationsService.scala
cquiroz: This is only to synchronize changes on the science observations. We were synchronizing new observations which are created already synchronized so it wasted some cycles

PR #2395, opened 2026-04-21
Call the ocs itc once to get both integration times and graphs

toddburnside:
[sc-8428]

For some reason, it seems that it was assumed that in order to get graphs from the OCS ITC we needed time and count. This is not the case because all calls to the OCS ITC do the same thing, including creating graphs. It is just that in some cases the graphs are thrown across the OCS-Lucuma boundary and in other cases are thrown away.

Anyway, in order to get both the time and the graphs, the OCS ITC was being called twice. Once to get the integration times, and once to get the graphs. The second call used the time and count from the first call (even if time and count exposureTimeMode had been specified). If anyone knows of a reason why we should only create graphs with time and count, let me know...

This PR makes a single call to the OCS to get both the time and the graphs. This will reduce both the memory and processing load on the ITC. It should make the calls to get graphs a little quicker.

Now, it seems to me that the lucuma ITC API call for `spectroscopyGraphs` is kind of useless. One, there seems to be no reason to require the time and count vs an ExposureTimeMode. Two, the integration time data is fairly small and needs to be computed anyway, so there is no real savings in excluding it. The ItcClient includes the call to `spectroscopyGraphs`, but I'd like to remove it. It doesn't appear to be used anywhere. This would also allow deleting a fair amount of code and a number of types to simplify things. Any objections to this?

UPDATE: `spectroscopyGraphs` has been removed from the ITC API, and the return type of `spectroscopyIntegrationTimesAndGraphs` has been simplified.


cquiroz:
in the beggining of time we'd do two graphql calls to get first the time and then the chart, that was then reduced into a single call but under the hood we'd still call ocs twice. Improving that would be great

I'm ok dropping `spectroscopyGraphs` as it stands now

swalker2m:
> all calls to the OCS ITC do the same thing, including creating graphs

The ODB sequence generation doesn't care about the graphs but just needs the time and count.  Does it need to pay the cost of creating a graph that is ultimately not used?

toddburnside:
> > all calls to the OCS ITC do the same thing, including creating graphs
> 
> The ODB sequence generation doesn't care about the graphs but just needs the time and count. Does it need to pay the cost of creating a graph that is ultimately not used?

You guessed it! The only difference is that it is not serialized and passed to the lucuma ITC. It is replaced by Nil at that point. This is another thing I'm hoping to fix. We should be able to do the required calculations directly with the `Spectra` that get created instead of the graphs which are derived from the spectra. GHOST is high enough resolution that creating the graphs adds a lot of memory pressure to the ITC.

cquiroz:
> > > all calls to the OCS ITC do the same thing, including creating graphs
> > 
> > 
> > The ODB sequence generation doesn't care about the graphs but just needs the time and count. Does it need to pay the cost of creating a graph that is ultimately not used?
> 
> You guessed it! The only difference is that it is not serialized and passed to the lucuma ITC. It is replaced by Nil at that point. This is another thing I'm hoping to fix. We should be able to do the required calculations directly with the `Spectra` that get created instead of the graphs which are derived from the spectra. GHOST is high enough resolution that creating the graphs adds a lot of memory pressure to the ITC.

I bet you are dying to avoid calculating the full spectra and instead only a the section where we care, around the requested wavelength

toddburnside:
> > > > all calls to the OCS ITC do the same thing, including creating graphs
> > > 
> > > 
> > > The ODB sequence generation doesn't care about the graphs but just needs the time and count. Does it need to pay the cost of creating a graph that is ultimately not used?
> > 
> > 
> > You guessed it! The only difference is that it is not serialized and passed to the lucuma ITC. It is replaced by Nil at that point. This is another thing I'm hoping to fix. We should be able to do the required calculations directly with the `Spectra` that get created instead of the graphs which are derived from the spectra. GHOST is high enough resolution that creating the graphs adds a lot of memory pressure to the ITC.
> 
> I bet you are dying to avoid calculating the full spectra and instead only a the section where we care, around the requested wavelength

Well, that might affect the totalSNRatio and singleSNRatio values, so I'm not sure we can do that? Of course, we don't always need those, either, so 🤷

cquiroz:
> Anyway, in order to get both the time and the graphs, the OCS ITC was being called twice. Once to get the integration times, and once to get the graphs. The second call used the time and count from the first call (even if time and count exposureTimeMode had been specified). If anyone knows of a reason why we should only create graphs with time and count, let me know...
I think it's just a historical artifact as in s/n you need this two stage process and it got carried over ro TxC

> 
> This PR makes a single call to the OCS to get both the time and the graphs. This will reduce both the memory and processing load on the ITC. It should make the calls to get graphs a little quicker.
👏 
> 
> Now, it seems to me that the lucuma ITC API call for `spectroscopyGraphs` is kind of useless. One, there seems to be no reason to require the time and count vs an ExposureTimeMode. Two, the integration time data is fairly small and needs to be computed anyway, so there is no real savings in excluding it. The ItcClient includes the call to `spectroscopyGraphs`, but I'd like to remove it. It doesn't appear to be used anywhere. This would also allow deleting a fair amount of code and a number of types to simplify things. Any objections to this?
> 
> UPDATE: `spectroscopyGraphs` has been removed from the ITC API, and the return type of `spectroscopyIntegrationTimesAndGraphs` has been simplified.
🎉

cquiroz (approved):
LGTM

PR #2397, opened 2026-04-22
Shortcut 8293: GHOST time estimates

swalker2m:
GHOST setup and detector time estimates.

### DIFF
--- a/modules/service/src/main/resources/db/migration/V1126__ghost.sql
+++ b/modules/service/src/main/resources/db/migration/V1126__ghost.sql
@@ -8,37 +8,43 @@ CREATE TABLE t_ghost_readout (
   c_blue_time interval    NOT NULL
 );
 
--- https://www.gemini.edu/instrumentation/ghost/components Table 3
+-- Taken from:
+-- https://github.com/gemini-hlsw/seqexec/blob/main/modules/server/src/main/scala/seqexec/server/ghost/GhostLUT.scala#L145
+
+-- The values published at:
+-- https://www.gemini.edu/instrumentation/ghost/components (Table 3)
+-- are reportedly fake news.
+
 COPY t_ghost_readout (
   c_read_mode,
   c_binning,
   c_red_time,
   c_blue_time
 ) FROM STDIN;
-slow	one_by_one	97.4 seconds	45.6 seconds
-slow	one_by_two	49.6 seconds	24.8 seconds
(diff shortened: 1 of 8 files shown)


modules/service/src/main/resources/db/migration/V1126__ghost.sql
```diff
@@ -0,0 +1,74 @@
...
+);
+
+-- https://www.gemini.edu/instrumentation/ghost/components Table 3
+COPY t_ghost_readout (
+  c_read_mode,
+  c_binning,
+  c_red_time,
+  c_blue_time
+) FROM STDIN;
+slow	one_by_one	97.4 seconds	45.6 seconds
```
cquiroz: It seems these numbers don't match what we have on the seqexec, in there one_by_one, slow gives red 100.75 and blue 45.957

cquiroz: As a reference:
https://github.com/gemini-hlsw/seqexec/blob/main/modules/server/src/main/scala/seqexec/server/ghost/GhostLUT.scala#L141

swalker2m: That says 

```
// Taken from OCS GhostCameras.scala
```

These numbers come from https://www.gemini.edu/instrumentation/ghost/components

cquiroz: The numbers on the seqexec were refined after multiple iterations testing with the instruments. Perhaps the one of the web page are `original` ones?

swalker2m: Why does the seqexec have readout time estimates?

cquiroz: it uses them to estimate the duration of the progress bar as we are emulating them. I recall the original numbers didn't match and they were refined over multiple testing sessions

swalker2m: Alright, I switched to your numbers.  They can be adjusted at any point in a simple, albeit tedious, flyway migration.

The API provides the time estimates in the steps so hopefully Observe doesn't need to recalculate these but instead can read it from there.

cquiroz: This was long a request that observe doesn't have its own numbers

cquiroz (approved):
LGTM

PR #2402, opened 2026-04-23
Primary migration step to support Gnirs

rpiaggio:
This PR is primarily to make the ODB compile with the latest core which introduces `GnirsLongSlit` mode and `InstrumentExecutionConfig.Gnirs`~, as well as adding the `coadds` parameter to `ExposureTimeMode.TimeAndCountMode`~.

None of these things are persisted yet and nothing has been touched in the API. But we need the ODB to compile again ASAP, so that will be done in future PRs.

The Gnirs config codecs were straightforward and so are already added in this PR.


modules/service/src/test/scala/lucuma/odb/unimplemented.scala
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2016-2025 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.odb
+
+import lucuma.core.enums.ObservingModeType
+
+val UnImplementedModes: Set[ObservingModeType] = Set(ObservingModeType.GhostIfu, ObservingModeType.GnirsLongSlit)
```
swalker2m, line 8: Thanks, this is a great idea.  I think `GhostIfu` should work now, but I'll check it in another PR.

swalker2m (approved):
Looks good. :+1:

PR #2406, opened 2026-04-24
Shortcut 8212: GHOST smart gcal

swalker2m:
Initial version of Smart GCal for GHOST.  There will likely be updates because:

* The slit viewing camera isn't yet in the dynamic config
* the ARC and FLAT files have additional configuration that doesn't exist in model yet
* it isn't clear when we need to generate gcals in the first place
* the binning mapping seems overly restrictive to me, or else the files are missing entries.


toddburnside (approved):
LGTM

PR #2412, opened 2026-04-28
Shortcut 8471: orphaned observing mode

swalker2m:
While implementing the GHOST observing mode one piece at a time, I added a create method but had not yet included a clone. We suspect that this resulted in a cloned GHOST observation which had a `c_observing_mode_type` discriminator set to GHOST, but no matching observing mode table entry.  In general the rule should be that when `t_observation.c_observing_mode_type` is `null`, then there should be no entries in any observing mode table.  If it is not `null`, there should be one row in the corresponding observing mode table.

To prevent this type of inconsistency from occurring in the future, we can add some consistency check triggers.  Is this worth the complication?


modules/service/src/main/resources/db/migration/V1131__observing_mode_consistency.sql
```diff
@@ -0,0 +1,138 @@
...
+
+-- Register each observing mode so that the consistency check will be
+-- performed when they are updated.
+SELECT register_observing_mode('flamingos_2_long_slit', 't_flamingos_2_long_slit');
+SELECT register_observing_mode('ghost_ifu',             't_ghost_ifu');
+SELECT register_observing_mode('gmos_north_imaging',    't_gmos_north_imaging');
+SELECT register_observing_mode('gmos_north_long_slit',  't_gmos_north_long_slit');
+SELECT register_observing_mode('gmos_south_imaging',    't_gmos_south_imaging');
+SELECT register_observing_mode('gmos_south_long_slit',  't_gmos_south_long_slit');
+SELECT register_observing_mode('igrins_2_long_slit',    't_igrins_2_long_slit');
```
cquiroz, line 138: I guess we need to remember to update this list when adding new modes

cquiroz (approved):
LGTM

modules/service/src/main/resources/db/migration/V1131__observing_mode_consistency.sql
swalker2m: Right, but if you forget at least there will be a quick failure instead of leaving an inconsistent observation.

toddburnside: Should we have a comment or documentation somewhere reminding devs that this needs to be done and how to do it?

swalker2m: I added a comment to `ObservingModeServices` along the lines of "if that mode ain't listed, it never existed".  You pretty much update that code whenever a new mode is added so maybe someone will see it.  In reality what is most likely to happen is the message won't be seen, the trigger will abort the transaction with a message, the developer will it look up and only _then_ realize what needs to be done.

I think this is better than leaving the observation in a bad state and preventing Explore from opening the program, but I could also be convinced to just close the PR.

swalker2m: It is a lot of fuss for something that should only happen while developing a new mode.

cquiroz: True but i don't think it hurts to have an extra safeguard

PR #2415, opened 2026-04-29
Shortcut 8293: Validate ETM update for GHOST

swalker2m:
Adds input validation for GHOST IFU mode updates.  When the ETM for a single channel (red or blue) is specified, but not the other, this requires performing a lookup to find the existing ETM for the other channel for comparison.  We now require that both ETMs be `TimeAndCount` and that the wavelength `at` is equivalent.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/GhostIfuService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/GhostIfuService.scala
@@ -3,10 +3,11 @@
 
 package lucuma.odb.service
 
+import cats.Order.catsKernelOrderingForOrder
 import cats.data.NonEmptyList
 import cats.effect.Concurrent
 import cats.syntax.applicative.*
-import cats.syntax.apply.*
+import cats.syntax.eq.*
 import cats.syntax.flatMap.*
 import cats.syntax.foldable.*
 import cats.syntax.functor.*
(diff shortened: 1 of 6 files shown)


modules/service/src/main/scala/lucuma/odb/service/GhostIfuService.scala
```diff
@@ -63,23 +64,45 @@ object GhostIfuService:
...
+
+    def etmIdColumn: String =
+      s"c_${color}_exposure_time_mode_id"
+
+  private object validateEtms:
+    def errorMessage(oids: List[Observation.Id]): String =
+      val oidsString = oids match
+        case Nil => ""
+        case _   => s"${oids.sorted.mkString("(", ", ", ")")} "
+      s"GHOST observations ${oidsString}require red and blue channel TimeAndCount exposure time modes with equivalent wavelength values."
```
cquiroz, line 80: a space before `require`?

cquiroz (approved):
LGTM

modules/service/src/main/scala/lucuma/odb/service/GhostIfuService.scala
swalker2m: Kind of a hack, but if there are no `oids` then I don't want _two_ spaces between "observations" and "require" so I just add it to the non-`Nil` case.

PR #2416, opened 2026-04-30
SC-8459 visitor mode support in odb

tpolecat:
Part 1 of support for visitor modes, which are "real" modes but they all use the same configuration table that contains only the central wavelength and max guidestar offset (this can be defaulted in the UI for permanent visitor instruments). There is no AGS, nor will workflow behave as expected (visitor mode observations generate no steps and thus will always be marked as complete). Existing tests exercise these modes.


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -11152,9 +11183,14 @@ Instrument
...
-  Instrument AcqCam
+  Instrument AcqCam at Gemini North
   """
-  ACQ_CAM
+  ACQ_CAM_NORTH
+
+  """
+  Instrument AcqCam at Gemini South
+  """
+  ACQ_CAM_SOUTH
```
swalker2m, line 11193: :+1:

modules/service/src/main/scala/lucuma/odb/service/ObservationWorkflowService.scala
```diff
@@ -136,7 +136,7 @@ case class ObservationValidationInfo(
     )
 
   def site: Option[Site] =
-    instrument.flatMap(_.site.headOption) // TODO: fix this so there's never more than one instrument
+    instrument.map(_.site)
```
swalker2m, line 139: 🎉

swalker2m (approved):
It all seems perfectly cromulent.

PR #2421, opened 2026-05-02
Add same night fields and logic to the api

cquiroz:
The ODB side of adding a `sameNight` field to `AND` groups


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -10864,6 +10864,13 @@ type Group {
   minimumInterval: TimeSpan
   maximumInterval: TimeSpan
 
+  """
+  If true, all observations in this group must be scheduled on the same night.
+  Mutually exclusive with `maximumInterval`.
+  Only valid for AND groups.
+  """
+  sameNight: Boolean!
```
swalker2m, line 10910: Why do we need this if we have `maximumInterval`?  It can conflict with the intervals as noted in the comment.

modules/service/src/main/resources/db/migration/V1130__group_same_night.sql
```diff
@@ -0,0 +1,17 @@
+-- Add Same Night option to groups
+ALTER TABLE t_group
+ADD COLUMN c_same_night bool NOT NULL DEFAULT false;
+
+-- c_same_night is mutually exclusive with c_max_interval and only valid for
+-- AND groups (c_min_required IS NULL).
+ALTER TABLE t_group
+ADD CONSTRAINT group_same_night_check
+CHECK (c_same_night = false OR (c_max_interval IS NULL AND c_min_required IS NULL));
```
swalker2m: If `c_min_required` is set to the number of observations in the group, is that not an `AND` group as well?

Why does it matter anyway that it be an `AND` group?  It seems like it could apply equally to completing 2 out of 4 observations.

swalker2m (approved):
The code looks great.  I don't understand the motivation for the flag, but :+1: on the code.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
andrewwstephens: We need a way to tell the scheduler that all the observations in a group must be observed on the same night.  One use case is to require an on-sky calibration that could be observed at any point during the night - before or after the science observation(s).  Using `maximumInterval` for this purpose gets complicated when there are more than 2 observations in the group.

modules/service/src/main/resources/db/migration/V1130__group_same_night.sql
andrewwstephens: Yes, this could apply to `OR` groups, but we couldn't think of any use-cases.  If you prefer, we could allow it in the database but not provide the UI in OR groups.

cquiroz: Seems good, I"ll allow it for either type of groups

PR #2425, opened 2026-05-04
`executableOnly` flag in obscalc event subscriptions

swalker2m:
The ODB provides clients with an obscalc event subscription that triggers whenever it finishes updating an observation's calculated values.  This PR adds a flag to the subscription input, `executableOnly`, which when set to `true` limits events to only observations whose workflow state is now (or was previously) executable (i.e., `READY` or `ONGOING`).

This is available generally but was included with the scheduler in mind.  The scheduler is not interested in tracking updates to observations still in the planning stages or that have already been completed.  Instead, it can set the `executableOnly` flag and filter out updates that cannot possibly impact its schedule.


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -5343,6 +5343,13 @@ input ObscalcUpdateInput {
   oldState:             WhereOptionEqCalculationState @deprecated(reason: "Use `oldCalculationState` instead.")
   newCalculationState:  WhereOptionEqCalculationState
   newState:             WhereOptionEqCalculationState @deprecated(reason: "Use `newCalculationState` instead.")
+
+  """
+  When set to `true`, events will be limited to those that impact executable
+  observations only.  Changes to observations that are not yet `READY` or
+  `ONGOING`, or which have already completed, will not trigger events.
+  """
+  executableOnly:       Boolean = false
```
cquiroz, line 5352: do we have this `filter` also for queries? I'm thinking of navigate and observe

cquiroz (approved):
LGTM

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
swalker2m: In `WhereObservation` I suppose you could get the same filtering using the `workflow` field, albeit less conveniently.

PR #2426, opened 2026-05-05
Add camera and prism to Gnirs rows

rpiaggio:
We need these specific values to invoke ITC on the qualifying rows.

Also, the matrix table is updated from the spreadsheet.


modules/phase0/src/main/resources/phase0/Phase0_Instrument_Matrix - Spectroscopy.tsv
```diff
@@ -558,8 +558,44 @@ GNIRS	LC	singleslit	1.0	1.000	49	111	H	1.49	1.8	1.645	0.0475	1780	yes		n
 GNIRS	LC	singleslit	1.0	1.000	49	111	K	1.91	2.49	2.2	0.0633	1780	yes		n		
 GNIRS	LC	singleslit	1.0	1.000	49	111	L	2.8	4.2	3.5	0.0944	1900	no		n		
 GNIRS	LC	singleslit	1.0	1.000	49	111	M	4.4	6	4.8	0.192	1280	no		n		
-GNIRS	SC	singleslit	0.3	0.300	7	32	XD	0.85	2.5	1.675	2.5	1700	no		n		
-GNIRS	LC	singleslit	0.1	0.100	5	32	XD	0.85	2.5	1.675	2.5	5000	yes		n		
+GNIRS	SC XD	singleslit	0.3	0.300	7	32	XD	0.85	2.5	1.675	2.5	1700	no		n
```
cquiroz, line 561: did we add an extra column? I'd expect a header too

cquiroz: This is TSV, I guess that allows spaces inside columns?

modules/service/src/main/resources/db/migration/V1134__gnirs_prism_camera.sql
```diff
@@ -0,0 +1,29 @@
+--- GNIRS Prism
+CREATE TABLE t_gnirs_prism (
+  c_tag        d_tag   NOT NULL PRIMARY KEY,
+  c_short_name varchar NOT NULL,
+  c_long_name  varchar NOT NULL
+);
+
+INSERT INTO t_gnirs_prism VALUES ('Mirror', 'Mirror',   'Mirror');
```
cquiroz, line 8: I think we have used tag in lower case and with underscores, not sure if it is important though

modules/phase0/src/main/resources/phase0/Phase0_Instrument_Matrix - Spectroscopy.tsv
rpiaggio: Right, it's not an extra column but a space in the 2nd column.

modules/service/src/main/resources/db/migration/V1134__gnirs_prism_camera.sql
rpiaggio: We use the `Enumerated`'s `tag` in the table. I see some enums following the lower_case convention, but some of them don't (eg the `t_gmos_*` ones). I just went with the existing `tag` defined in core. We should change all the non-conforming enums in core if we want to follow the convention.

cquiroz: seems fair, i guess they should match the tag in core, over whatever non existing convention

cquiroz (approved):
LGTM

PR #2429, opened 2026-05-05
Ghost Central Wavelength

swalker2m:
Adds a `centralWavelength` to the `GhostDynamicConfig` schema, which is just a fixed value.  This also changes said fixed value to 655 nm to match the `ocs`.  We were using the wavelength value as the AGS wavelength as well.  Is this correct?

If this seems acceptable I'll do a similar fix for IGRINS.

### DIFF
--- a/modules/sequence/src/main/scala/lucuma/odb/sequence/ghost/shared.scala
+++ b/modules/sequence/src/main/scala/lucuma/odb/sequence/ghost/shared.scala
@@ -6,7 +6,7 @@ package lucuma.odb.sequence.ghost
 import lucuma.core.math.Wavelength
 
 /**
- * GHOST central wavelength, which seems to be fixed at 530 nm.
+ * GHOST central wavelength, which seems to be fixed at 655 nm.
  */
 val CentralWavelength: Wavelength =
-  Wavelength.unsafeFromIntPicometers(530000)
+  Wavelength.unsafeFromIntPicometers(655000)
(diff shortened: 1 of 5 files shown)


modules/sequence/src/main/scala/lucuma/odb/sequence/ghost/shared.scala
```diff
@@ -6,7 +6,7 @@ package lucuma.odb.sequence.ghost
 import lucuma.core.math.Wavelength
 
 /**
- * GHOST central wavelength, which seems to be fixed at 530 nm.
+ * GHOST central wavelength, which seems to be fixed at 655 nm.
  */
 val CentralWavelength: Wavelength =
-  Wavelength.unsafeFromIntPicometers(530000)
\ No newline at end of file
+  Wavelength.unsafeFromIntPicometers(655000)
```
cquiroz, line 12: Should this go to core, like igrins2?

cquiroz (approved):
LGTM, we could put igrins2 as well wit the same format

modules/sequence/src/main/scala/lucuma/odb/sequence/ghost/shared.scala
swalker2m: Where is the IGRINS2 value in core?  I'll merge this and then switch it to a similar value from core in a separate PR.

cquiroz: https://github.com/gemini-hlsw/lucuma-core/blob/main/modules/core/shared/src/main/scala/lucuma/core/model/sequence/igrins2/package.scala#L33

PR #2430, opened 2026-05-06
Only receive y-data for graphs from the ocs

toddburnside:
Now only y data and x-axis information is being returned from the OCS. This should save significant time serializing and deserializing. In addition, the x data is never constructed on the server, and traversal of the y data is minimized.

### DIFF
--- a/itc/service/src/main/resources/graphql/itc_base.graphql
+++ b/itc/service/src/main/resources/graphql/itc_base.graphql
@@ -394,21 +394,29 @@ enum ItcGraphType {
   S2N_GRAPH
 }
 
-type GraphAxis {
+type GraphXAxis {
   """First value of the axis"""
   start: Float!
 
   """Last value of the axis"""
   end: Float!
 
+  """How many values for the axis"""
+  count: PosInt!
+}
+
+type GraphYAxis {
   """Max value of the axis"""
   max: Float!
+
+  """Index of the max value on the x axis"""
+  indexOfMax: Int!
 
   """Min value of the axis"""
   min: Float!
 
-  """How many values for the axis"""
-  count: PosInt!
+  """Index of the min value on the x axis"""
+  indexOfMin: Int!
 }
 
 type ItcSeries {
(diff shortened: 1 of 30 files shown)


itc/service/src/main/resources/graphql/itc_base.graphql
```diff
@@ -394,21 +429,29 @@ enum ItcGraphType {
...
-type GraphAxis {
+type GraphXAxis {
   """First value of the axis"""
   start: Float!
 
   """Last value of the axis"""
   end: Float!
 
+  """How many values for the axis"""
+  count: PosInt!
```
cquiroz, line 440: I guess this implies a constant step size, is that correct?

itc/service/src/main/resources/graphql/itc_base.graphql
```diff
@@ -505,9 +542,6 @@ type SpectroscopyTimeAndGraphsResult {
   """ITC results for each target, in case all targets succeed"""
   targetTimesAndGraphs: [TargetTimeAndGraphsOutcome!]
 
-  """In case of errors in integration times, we short cirtuit to the results from integration times"""
```
cquiroz, line 508: 👍

itc/service/src/main/resources/graphql/itc_base.graphql
toddburnside: Correct. We were actually doing this before to send the axis data to explore, I just moved that calculation to the ocs.

itc/service/src/main/scala/lucuma/itc/service/ItcCacheOrRemote.scala
```diff
@@ -28,7 +28,7 @@ import scala.concurrent.duration.*
  */
 trait ItcCacheOrRemote extends Version:
   // Manual version to manually force a flush
-  private val ManualCacheVersion = 1
+  private val ManualCacheVersion = 2
```
cquiroz: since the jar files were updated this may not be needed

toddburnside: Oh, does that break the cache as well?

toddburnside: Ah, I see that below. I can revert this.

cquiroz (approved):
LGTM

PR #2432, opened 2026-05-06
GHOST and IGRINS2 Central Wavelength

swalker2m:
* Switches the `GhostDynamicMapping` to use a new `CentralWavelength` value from core
* Adds `centralWavelength` for the `Igrins2DynamicMapping`, taking the value from core


cquiroz (approved):
LGTM

PR #2433, opened 2026-05-06
Fix guideEnvironment query for GHOST

toddburnside:
Since I'll be out for a few days and I don't want to block GHOST testing, I set the PositionAngleConstraint of the observations in the `guideEnvironmentGhost` tests to `Fixed(0)` so that the tests will pass. This doesn't explain why, when they were `Unbonded` I was getting posAngles from the AGS 180 different on CI than locally. I'll look into that more next week. But, it IS selecting the expected guide star.


modules/service/src/test/scala/lucuma/odb/graphql/DatabaseOperations.scala
```diff
@@ -813,6 +813,9 @@ trait DatabaseOperations { this: OdbSuite =>
   def createIgrins2LongSlitObservationAs(user: User, pid: Program.Id, offsets: Option[String], tids: Target.Id*): IO[Observation.Id] =
     createObservationWithSpatialOffsets(user, pid, ObservingModeType.Igrins2LongSlit, ImageQuality.Preset.PointEight, offsets, tids*)
 
+  def createGhostIfuObservationAs(user: User, pid: Program.Id, offsets: Option[String], tids: Target.Id*): IO[Observation.Id] =
```
swalker2m, line 816: This also exists in `ObservingModeSetupOperations`.  Someday, when we have time, we'll straighten all of this out.

toddburnside: Yeah. Pretty much all of them are in both places with just slightly different parameter lists. It caused me a bit of  confusion, so I just went with the flow.

modules/service/src/test/scala/lucuma/odb/graphql/query/guideEnvironmentGhost.scala
```diff
@@ -0,0 +1,131 @@
...
+      expect(pi, guideEnvironmentQuery(oid), expected = ghostPwfs2Result("V1647 Orionis", BigDecimal("130.000000")))
+
+  test("nonsidereal target with PWFS2"):
+    val setup: IO[Observation.Id] =
+      for
+        p   <- createProgramAs(pi)
+        eph  = createNonsiderealEphemeris
+        t   <- createNonsiderealTargetWithUserSuppliedEphemerisAs(pi, p, eph, name = "Nonsidereal Target")
+        o   <- createObservationAs(pi, p, List(t))
+        _   <- setObservationTimeAndDuration(pi, o, gaiaSuccess.some, fullTimeEstimate.some)
```
cquiroz, line 178: what is the obstime for this observation?

toddburnside: The same as for all of the other guideEnvironmentXXX tests:

> val gaiaSuccess: Timestamp = Timestamp.FromString.getOption("2023-08-30T00:00:00Z").get

swalker2m (approved):
Looks great. :+1:

PR #2434, opened 2026-05-06
Include resident visitor modes for imaging and spectroscopy

cquiroz:


swalker2m:
Are these distinct observing modes or just different types of visitor?  It makes sense for the mode selection table but what happens next after one is selected.

modules/phase0/src/main/scala/lucuma/odb/phase0/ConfigurationRow.scala
```diff
@@ -160,7 +161,7 @@ case class ImagingRow(
   fov:            Angle,
   filter:         String,
   ao:             Boolean,
-  capability:     Option[Capability],
+  capability:     Option[ImagingCapabilities],
```
swalker2m: The plurality of "capabilities" is throwing me. Can there be more than one in an ImagingRow?

cquiroz:
> Are these distinct observing modes or just different types of visitor? It makes sense for the mode selection table but what happens next after one is selected.

For the resident vistiors we want to record more details about the mode but at the end the will all be stored as Visitors I think

modules/phase0/src/main/scala/lucuma/odb/phase0/ConfigurationRow.scala
cquiroz: This follows `SpectrosocpyCapabilities` enum. at the current state they have only one capability

modules/phase0/src/main/resources/phase0/Phase0_Instrument_Matrix - Imaging.tsv
```diff
@@ -9,8 +9,10 @@ NIRI f/14	50	J,H,K	yes		n
...
-Zorro Wide Field	35	u,g,r,i,z,EO466,EO562,EO716,EO832,Halpha	no		s
-SCORPIO	180	g,r,i,z,J,H,Ks	no		s
\ No newline at end of file
+ALOPEKE	3	u,g,r,i,z,EO466,EO562,EO716,EO832	no	speckle	n
+ZORRO	3	u,g,r,i,z,EO466,EO562,EO716,EO832,Halpha	no	speckle	s
+ALOPEKE	35	u,g,r,i,z,EO466,EO562,EO716,EO832	no	wide field	n
+ZORRO	35	u,g,r,i,z,EO466,EO562,EO716,EO832,Halpha	no	wide field	s
+SCORPIO	180	g,r,i,z,J,H,Ks	no		s
+Visitor @ GN	-	-	yes		n
+Visitor @ GS	-	-	no		s
```
swalker2m, line 18: This file came from science, correct?  Remind me, do they maintain it?  That is, do these updates come from them or do they need to go back into wherever they keep this?

modules/service/src/main/resources/db/migration/V1136__imaging_capability.sql
```diff
@@ -0,0 +1,14 @@
...
+  c_long_name  varchar(64) NOT NULL
+);
+
+INSERT INTO t_imaging_capabilities VALUES
+  ('speckle',    'Speckle',    'Speckle'),
+  ('wide_field', 'Wide Field', 'Wide Field');
+
+-- Add capability column to imaging config options
+ALTER TABLE t_imaging_config_option
+  ADD COLUMN c_capability d_tag DEFAULT NULL REFERENCES t_imaging_capabilities(c_tag);
```
swalker2m: Should this be singular everywhere it is used?

swalker2m (approved):
I'm not clear on what attributes will be kept with the visitor instrument mode but this just adds it to the Phase 0 options table and it looks good to me.

modules/phase0/src/main/resources/phase0/Phase0_Instrument_Matrix - Imaging.tsv
cquiroz: Science maintains it, having said that I sometimes update it rather, for example the code expect instrument names to be all capitals

tpolecat (approved):
This looks good. We want the visitor modes to pop up in the phase 0 list when appropriate, but as you say for now they are just stored as visitors instruments with no extra config other than central wavelength and guidestar max separation.

PR #2437, opened 2026-05-07
Friendlier Gnirs codecs, update core

rpiaggio:
Allows easier usage of Gnirs codecs downstream.


modules/schema/src/main/scala/lucuma/odb/json/gnirs.scala
```diff
@@ -64,6 +64,14 @@ trait GnirsCodec:
         "wavelength" -> a.wavelength.asJson
       )
 
+  given Decoder[GnirsAcquisitionMirrorMode] =
```
cquiroz, line 67: Why not use the enumerated encoder/decoders?

rpiaggio: This is not an enumerated.

cquiroz (approved):

PR #2439, opened 2026-05-07
Shortcut 8595: GHOST IFU `skyPosition` option

swalker2m:
Adds an optional `skyPosition` coordinate to the GHOST observing mode.  At the moment this is unused but will be written to the `GhostStatic` configuration as part of the IFU mapping in a proximate PR.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/GhostIfuService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/GhostIfuService.scala
@@ -17,6 +17,7 @@ import grackle.ResultT
 import lucuma.core.enums.GhostBinning
 import lucuma.core.enums.GhostReadMode
 import lucuma.core.enums.GhostResolutionMode
+import lucuma.core.math.Coordinates
 import lucuma.core.model.ExposureTimeMode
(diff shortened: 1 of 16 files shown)


toddburnside (approved):
LGTM

modules/service/src/main/scala/lucuma/odb/service/GhostIfuService.scala
```diff
@@ -229,6 +230,7 @@ object GhostIfuService:
         ghost_resolution_mode         *:
         ghost_detector_red            *:
         ghost_detector_blue           *:
+        (right_ascension *: declination).to[Coordinates].opt *:
```
cquiroz, line 233: don't we have a `Coordinate` encoder?

cquiroz (approved):
LGTM

modules/service/src/main/scala/lucuma/odb/service/GhostIfuService.scala
swalker2m: I couldn't find it, but you're right I should probably have added one. I'll do it in the next installment.

PR #2441, opened 2026-05-08
Don't allow setting non long-slit fpus

cquiroz:
on the bug report a user set NS on a longslit observation, this is allowed on the ODB but breaks on AGS.

We now have geometries for the NS fpus, but still we may not want to allow them selected on longslit observations for now. This PR basically doesn't allow them selected. 

I checked the dbs and only dev has offending observations and they use NS. Therefore I prefer not to migrate them and now they won't fail if selected as the geometry is present.


swalker2m (approved):
I wonder if we should have made a separate type for each kind, but at any rate this looks fine to me.

Ideally we'd prevent or restrict the `c_fpu` in `t_gmos_north_long_slit` as well.  But I think for that to be feasible we'd need different enums in the database.

cquiroz:
> I wonder if we should have made a separate type for each kind, but at any rate this looks fine to me.
> 
> Ideally we'd prevent or restrict the `c_fpu` in `t_gmos_north_long_slit` as well. But I think for that to be feasible we'd need different enums in the database.

I agree but we will enable those fpus eventually. What is open to me is whether we should only allow longslit fpus with longslit mode

swalker2m:
> > I wonder if we should have made a separate type for each kind, but at any rate this looks fine to me.
> > Ideally we'd prevent or restrict the `c_fpu` in `t_gmos_north_long_slit` as well. But I think for that to be feasible we'd need different enums in the database.
> 
> I agree but we will enable those fpus eventually. What is open to me is whether we should only allow longslit fpus with longslit mode

Right, that's what I mean.  `t_gmos_north_long_slit` is the database representation of GMOS North long slit observations.

PR #2443, opened 2026-05-09
Telluric disappears when science

cquiroz:
This is a fix for a bug that would remove the tellurics when an observation started executing. This is because the science moves to ongoing and that case wasn't being considered.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/CalibrationsUtils.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/CalibrationsUtils.scala
@@ -9,6 +9,7 @@ import cats.effect.Concurrent
 import cats.syntax.all.*
 import grackle.Result
 import lucuma.core.enums.CalibrationRole
+import lucuma.core.enums.ExecutionState
 import lucuma.core.enums.ObservationWorkflowState
 import lucuma.core.enums.ScienceBand
 import lucuma.core.enums.Site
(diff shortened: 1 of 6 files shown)


modules/service/src/main/scala/lucuma/odb/service/CalibrationsUtils.scala
```diff
@@ -177,6 +198,19 @@ object WorkflowStateQueries:
...
         void")"
 
+    def selectTelluricScienceWorkflowStates(oids: NonEmptyList[Observation.Id]): AppliedFragment =
+      void"""
+        SELECT t.c_observation_id, sci_obscalc.c_workflow_state
+        FROM t_observation t
+        JOIN t_observation sci
+          ON sci.c_group_id = t.c_group_id
+         AND sci.c_calibration_role IS NULL
+        JOIN t_obscalc sci_obscalc
```
swalker2m: We might consider switching to a join on `v_generator_params` instead of `t_obscalc`.  Getting the execution state doesn't require running obscalc or generating the sequence now.  You can just check the `c_execution_state` (which maps to `lucuma.core.enums.ExecutionState`) directly.

swalker2m (approved):
Looks fine, though I think getting the execution state from `v_generator_params` might be better. It should change immediately when appropriate from `ready` to `ongoing` whereas the obscalc will do all of its calculations in due time.

modules/service/src/main/scala/lucuma/odb/service/CalibrationsUtils.scala
cquiroz: Good point I didn't realize that, actually I. Can use it in other bits as well where we avoid deleting calibrations after started

swalker2m: Yes, I think that would make sense.

PR #2447, opened 2026-05-10
Rename guideMinStarSep to scienceFov

cquiroz:
Discussed with science and it makes more sense to use the sci fov as a param for ags

### DIFF
--- a/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -4805,13 +4805,13 @@ input ObservingModeInput @oneOf {
 input VisitorInput {
   mode: VisitorObservingModeType
   centralWavelength: WavelengthInput
-  guideStarMinSep: AngleInput
+  scienceFov: AngleInput
 }
 
 type Visitor {
   mode: VisitorObservingModeType!
   centralWavelength: Wavelength!
-  guideStarMinSep: Angle!
+  scienceFov: Angle!
 }
 
 """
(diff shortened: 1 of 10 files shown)


swalker2m (approved):
This seems fine, though it sort of implies that the `scienceFov` is always circular?  It makes me wonder whether the value should be the radius of the FOV or is it the diameter?  Presumably the "minimum separation" would be the angular separation between the base and the guide star whereas if the FOV value is the diameter it would be one half the value?

cquiroz:
yes we assume a circular fov which may not be exact in all cases, eventually we could allow a flag to pick a square shape if needed

This is the original science justification:
```It may be more intuitive to specify the science field of view. The minimum guide star separation must take into account the size of the PWFS arm, which is information that we would need to provide in the documentation.```

**Comment by swalker2m** [comment]
> yes we assume a circular fov which may not be exact in all cases, eventually we could allow a flag to pick a square shape if needed

When doing the guide star selection, would you use half of this value then?

**Comment by cquiroz** [comment]
For AGS we define circular areas where we don't allow vignetting passing a radius value, It is probably worth specifing that in the docs

**Comment by cquiroz** [comment]
> > yes we assume a circular fov which may not be exact in all cases, eventually we could allow a flag to pick a square shape if needed
> 
> When doing the guide star selection, would you use half of this value then?

Yes, that's correct

File: modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql [hunk 1]
```diff
@@ -4805,13 +4805,13 @@ input ObservingModeInput @oneOf {
 input VisitorInput {
   mode: VisitorObservingModeType
   centralWavelength: WavelengthInput
-  guideStarMinSep: AngleInput
+  scienceFov: AngleInput
```
swalker2m, line 4812: Maybe put a comment here on whether it is the diameter or radius?  Is this something we should look up from a table?  Visitors are named at least and the FOV won't change across usages of the same visitor instrument.  Or maybe it will, hmm.

cquiroz: We do know the fov of resident visitors. for alien visitors you can put whatever so from our point of view you can defined multiple obs with the same name visitor but different fovs. Same applies to central wavelength

PR #2449, opened 2026-05-11
SC-8464 update workflow for visitor modes

tpolecat:
This updates the workflow to allow staff (or better) users to transition visitor observations between `Ready` and `Ongoing`. It does this by adding `t_observation.c_declared_state` which can be nullable or contain `declared_ongoing` or `declared_complete`. The old `c_declared_complete` column is now computed and may be removable but I'm leaving it for now.

@swalker2m Is this going to cause downstream issues when there are [declared] ongoing observations that generate no steps?


tpolecat:
This is ready, pending build issues.

modules/service/src/main/resources/db/migration/V1139__visitor_workflow_status.sql
```diff
@@ -0,0 +1,134 @@
...
+  e.c_exposure_time,
+  e.c_exposure_count,
+  o.c_observing_mode_type,
+  o.c_science_band,
+  o.c_declared_complete,
+  o.c_declared_state,
+  CASE
+    -- The observation is explicitly marked complete -> completed.
+    WHEN o.c_declared_complete
+      THEN 'completed'::e_execution_state
```
swalker2m: I would guess that we want to say that when `c_declared_state` isn't null, then use it directly and don't try to compute the execution state any further.

```sql
WHEN o.c_declared_state IS NOT NULL THEN o.c_declared_state
```

or something.

tpolecat: Well, good point. I will change that.

modules/service/src/main/resources/db/migration/V1141__visitor_workflow_status 3.sql
```diff
@@ -0,0 +1,114 @@
...
+-- This is a multi-step process that requires re-creating two views.
+
+-- Continued from previous migration
+
+-- Drop `c_declared_complete` and re-create as computed.
+ALTER TABLE t_observation
+  DROP c_declared_complete;
+ALTER TABLE t_observation
+  ADD c_declared_complete BOOLEAN NOT NULL
+  GENERATED ALWAYS AS (c_declared_state IS NOT DISTINCT FROM 'declared_complete'::e_execution_state) STORED;
```
swalker2m: I can't think of what we need this for, given the new `c_declared_state` and the update to `v_generator_params` to use it.  Is there anything else?

swalker2m (approved):
This seems good to me. 👍

modules/service/src/main/resources/db/migration/V1141__visitor_workflow_status 3.sql
tpolecat: I did it initially to prevent breaking everything but it may not be necessary anymore. I will see if I can remove it.

tpolecat: Yep seems fine to remove.

PR #2453, opened 2026-05-12
Some improvements to speed up the build on CI

cquiroz:
* Disable coverage
* We were doing static checks on each test shard, instead now we have a separate task to do it making shards only do the actual tests
* test shards were doing lfs checkout. I thought this would be fast as we are caching that, but sometimes it would do a full checkout in some of the shards making the whole process slower. Now we don checkout lfs as the shards don't run the legacy itc tests

CI has gone from about 17m to 13m on a few runs I did and it should be more deterministic.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,9 +36,9 @@ jobs:
     timeout-minutes: 60
     steps:
       - name: Checkout current branch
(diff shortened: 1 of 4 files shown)


swalker2m (approved):
Sounds reasonable to me. :+1:

PR #2456, opened 2026-05-12
Add name and requested time fields

cquiroz:
alien visitors need to provided a name an the total time for proposals

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/VisitorService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/VisitorService.scala
@@ -7,7 +7,10 @@ import cats.data.NonEmptyList
 import cats.effect.Concurrent
 import cats.implicits.*
 import grackle.Result
+import lucuma.core.enums.VisitorObservingModeType
 import lucuma.core.model.Observation
+import lucuma.odb.data.OdbError
+import lucuma.odb.data.OdbErrorExtensions.*
 import lucuma.odb.graphql.input.VisitorInput
 import lucuma.odb.sequence.visitor.Config
 import lucuma.odb.service.Services.Syntax.session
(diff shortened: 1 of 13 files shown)


modules/service/src/main/scala/lucuma/odb/service/VisitorService.scala
```diff
@@ -121,6 +129,19 @@ object VisitorService:
...
+  // Alien visitor require both a name and a total request time.
+  private def validateFields[A, B](
+    mode:      VisitorObservingModeType,
+    name:      Option[A],
+    totalTime: Option[B]
+  ): Either[OdbError, Unit] =
+    mode match
+      case VisitorObservingModeType.VisitorNorth | VisitorObservingModeType.VisitorSouth if !name.isDefined || !totalTime.isDefined =>
+        OdbError.InvalidArgument(
+          s"Visitor mode $mode requires both `name` and `totalRequestTime` to be provided.".some
```
toddburnside, line 141: When we create a visitor mode in explore, will we have the name and totalRequestTime info? It seems like we won't so this validation would fail. And, if it is required should it be validated in `VisitorInput.Edit.toCreate`?

toddburnside (approved):
LGTM with a question.

modules/service/src/main/scala/lucuma/odb/service/VisitorService.scala
cquiroz: Yes, the UI will require them.
I'll add validation in `toCreate`

toddburnside: Don't we need to create an observing mode when the user `Accept`s a mode from the table? At that point we don't have the name and totalRequestTime, do we?

cquiroz: You're right, for alien visitors we won't use the table but a separate form where you enter the fields and only then you can accept it

toddburnside: Oh. Cool. I think.

toddburnside: If that is the case, should they be Options in `Config` and non-required in the API result for the queries?

cquiroz: They are optional

toddburnside: Right. But should they be? It seems like they are required.

cquiroz: they are required for alien visitors but not for resident visitors. One could argue we should have that separation in the schema

cquiroz: Maybe a question for @tpolecat

tpolecat: I think this approach is fine for now. The non-alien visitors may get split off into their own tables, leaving this one behind (and making more sense).

tpolecat (approved):
👍🏻

PR #2457, opened 2026-05-13
Core OffsetMode rename to make it generic

rpiaggio:


cquiroz (approved):
LGTM, we shoulld have the explore side ready before merging

rpiaggio:
> LGTM, we shoulld have the explore side ready before merging

Yeah, I have it ready, but we need to release the libs from here and update the version in explore before sending the PR.

PR #2460, opened 2026-05-15
Alopeke and zorro modes should not expand

cquiroz:
Since we are not storing filter selection we may rather not show them as separate rows

### DIFF
--- a/modules/phase0/src/main/scala/lucuma/odb/phase0/ConfigurationRow.scala
+++ b/modules/phase0/src/main/scala/lucuma/odb/phase0/ConfigurationRow.scala
@@ -196,4 +196,19 @@ object ImagingRow extends RowParsers {
     }
   }
 
+  /**
+   * A single line in the .tsv file becomes exactly one row.
+   * Used for instruments that can't (yet) store details like filters (Alopeke, Zorro)
+   */
+  val singleRow: Parser[List[ImagingRow]] = (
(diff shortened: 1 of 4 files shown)


modules/phase0/src/main/scala/lucuma/odb/phase0/ConfigurationRow.scala
```diff
@@ -196,4 +196,19 @@ object ImagingRow extends RowParsers {
...
+   */
+  val singleRow: Parser[List[ImagingRow]] = (
+    (instrument        <* htab) ~
+    (arcsec            <* htab) ~
+    (string            <* htab) ~ // hold all filters
+    (ao                <* htab) ~
+    (imagingCapability <* htab) ~
+    site
+  ).map { case (((((inst, fov), filter), ao), capability), site) =>
+    List(ImagingRow(inst, fov, filter, ao, capability, site))
```
toddburnside, line 211: Does this mean the row displayed in explore will contain a long comma separated list of filters?

toddburnside (approved):
LGTM

modules/phase0/src/main/scala/lucuma/odb/phase0/ConfigurationRow.scala
cquiroz: yes, science was ok with that idea

PR #2461, opened 2026-05-15
Enable resource schema validation in CI

hugo-vrijswijk:


hugo-vrijswijk:
Not sure how to handle the resource import statements. Closing for now...

PR #2462, opened 2026-05-15
Use SchemaStitcher for ITC and SSO and only load SSO Schema once at startup

hugo-vrijswijk:


cquiroz:
Could you ellaborate on what does this PR do? how does it work with respect to the dynamic enums?

hugo-vrijswijk:
> Could you ellaborate on what does this PR do? how does it work with respect to the dynamic enums?

It doesn't change anything for dynamic enums. The stitching script still remains for ITC.

It does do two other things:
- Uses the `SchemaStitcher` (originally from navigate) for ITC and SSO. It:
  - Loads schemas in a single non-blocking way in `F[_]`, to reduce code duplication
  - Allows import statements like `#import UserId from "lucuma/odb/graphql/OdbSchema.graphql"` which loads types from other schemas in a project's resources, to further reduce code usage. It _could_ also be used to replace the `itc.graphql`/`itc_base.graphql` stitching script, but then the schema validator action doesn't work anymore (as it doesn't understand these import statements). It'd be nice if this did work, but I don't know how (yet)
- Moves the schema loading up to the application initialization in the SSO service. Before, it would be read (and parsed) for every request. Now it just happens once during startup.

Main reasoning behind the Scala SchemaStitcher is to be able to use the import statements. I noticed the Node stitching script doesn't fully prune types that are not used. So this seemed like a cleaner solution. The import syntax is also standard (and used by the navigate schema already, for example)

toddburnside:
> I noticed the Node stitching script doesn't fully prune types that are not used. So this seemed like a cleaner solution.

Maybe a future task could be to use the scala stitcher for stitching the ITC schema. There is a lot of unused types left in that schema. 

Never mind. I see you already mentioned that in your elaboration.

toddburnside:
> * but then the schema validator action doesn't work anymore (as it doesn't understand these import statements). It'd be nice if this did work, but I don't know how (yet)

Maybe we could still stitch it "manually" to start with, using the scala schema stitcher via scala-cli?

cquiroz:
It would be great to have 2 things:

* No dynamic enums such we could avoid having to start a server to get the schema
* Proper imports, not sure if there is a graphql standard for that

This comment is not really related to this PR more of a general wish list

itc/service/src/main/scala/lucuma/itc/service/ItcMapping.scala
```diff
@@ -22,28 +23,34 @@ import lucuma.itc.service.config.*
...
-  def loadSchema[F[_]: Sync]: F[Schema] =
-    Sync[F]
-      .blocking:
-        Using(Source.fromResource("graphql/itc.graphql", getClass().getClassLoader())): src =>
-          Schema(src.mkString).toEither
-            .fold(
-              x => sys.error(s"Invalid schema: ${x.toList.mkString(", ")}"),
-              identity
+  def loadSchema[F[_]: Sync: Logger]: F[Schema] =
+    SchemaStitcher[F](Path("graphql/itc.graphql"), SchemaSource.fromResource).build
```
cquiroz, line 36: itc.graphql doesn't use `#import` in that case what is the benefit of schema stitcher?
This doesn't mean that it shouldn't it would be probably better to use `#import` if it were standard rather than the manual step

modules/sso-service/src/main/resources/Sso.graphql
```diff
@@ -1,3 +1,4 @@
+#import UserId, Partner, UserProfile from "lucuma/odb/graphql/OdbSchema.graphql"
```
cquiroz, line 1: Did you say this breaks the validation action for schema changes?

modules/sso-service/src/main/scala/graphql/SsoMapping.scala
```diff
@@ -28,13 +29,13 @@ import lucuma.core.model.OrcidId
 import lucuma.core.model.StandardRole
 import lucuma.core.model.StandardUser
 import lucuma.core.model.User
+import lucuma.odb.graphql.schema.SchemaSource
+import lucuma.odb.graphql.schema.SchemaStitcher
```
cquiroz, line 33: Was `SchemaStitcher` already present? I thought it was done for `navigate`

itc/service/src/main/scala/lucuma/itc/service/ItcMapping.scala
hugo-vrijswijk: Mostly so that the schema is loaded in the same way in most projects, and in a non-blocking way in F[_] with slightly clearer errors

modules/sso-service/src/main/resources/Sso.graphql
hugo-vrijswijk: Yes it _can_ handle import statements like these (they are standard), but it doesn't know how to resolve from other resource folders

modules/sso-service/src/main/scala/graphql/SsoMapping.scala
hugo-vrijswijk: I added it to this project when working on Resource

hugo-vrijswijk:
> It would be great to have 2 things:
> 
>     * No dynamic enums such we could avoid having to start a server to get the schema
> 
>     * Proper imports, not sure if there is a graphql standard for that
> 
> 
> This comment is not really related to this PR more of a general wish list

The second wish is implemented in #2477. With that (and this) we could remove the schema stitching script

cquiroz:
so itc instead of doing a "manual" stitching would import the odb file? that sounds great
if itc doesn't use any of the dynamic imports it seems we should go a ahead and adopt this solution

cquiroz (approved):
LGTM

PR #2463, opened 2026-05-15
Check generator params to avoid deleting per program calibrations

cquiroz:
This is a follow up of https://github.com/gemini-hlsw/lucuma-odb/pull/2443

This PR applies the same fix we did for per science calibrations to per program calibrations.


toddburnside (approved):
Looks good to me. 👨‍🦯

PR #2468, opened 2026-05-17
Phantom groups

cquiroz:
The phantom groups are telluric groups that for some reason have lost their science observation but the group still exists.

The calibrations daemon is supposed to clean them up and it seems pretty tight to me. However user reports inidcate that some of these events maybe related to troubles in drag and dropping.

Moving a science obs out of a telluric group is allowed, and that combined with the dnd problem may explain this.

This PR closes that option, forbiding observations to be removed from system groups (except by the daemon itself)

A migration hides the existing telluric groups to avoid having to delete every reference.


cquiroz:
As a side note. This is a solution of a possible reason for the phantom groups but there maybe other cases we haven't identified yet

toddburnside (approved):
LGTM

PR #2471, opened 2026-05-19
Fix visitor observing mode updates

cquiroz:
Missed a path for updating visitors and also fixes  a bad sql column name


rpiaggio (approved):
LGTM

PR #2472, opened 2026-05-19
Shortcut 8667: Fix observing night selection query

swalker2m:
This PR fixes the query to obtain the current observing night from a visit.  The current version works only for GMOS and was written, I believe, before the visit table included the visit site.

The failure to obtain the observing night related to a GNIRS visit resulted in not applying a daylight discount to an observation which lasted past the morning twilight.


toddburnside (approved):
LGTM

PR #2473, opened 2026-05-19
Telluric observation offsets are always the default ones

cquiroz:
This PR essentially adds a stage after sync where we can reset some properties of the tellurics. specifically offsets for this case

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/PerScienceObservationCalibrationsService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/PerScienceObservationCalibrationsService.scala
@@ -365,6 +365,17 @@ object PerScienceObservationCalibrationsService:
         def cloneSourceMode(sm: Option[ObservingModeType]): F[Unit] =
           sm.traverse(mode => obsModeService.clone(mode, sourceOid, targetOid)).void
 
+        // Reset any mode-specific telluric overrides on the cloned obs.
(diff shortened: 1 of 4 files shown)


modules/service/src/main/scala/lucuma/odb/service/PerScienceObservationCalibrationsService.scala
```diff
@@ -365,6 +365,17 @@ object PerScienceObservationCalibrationsService:
...
           sm.traverse(mode => obsModeService.clone(mode, sourceOid, targetOid)).void
 
+        // Reset any mode-specific telluric overrides on the cloned obs.
+        // At the moment only offsets but eventually other properties may need a reset,
+        def resetTelluricConfig(sm: Option[ObservingModeType]): F[Unit] =
+          sm match
+            case Some(ObservingModeType.Igrins2LongSlit) =>
+              S.igrins2LongSlitService.resetTelluricConfig(targetOid)
+            case Some(ObservingModeType.Flamingos2LongSlit) =>
+              S.flamingos2LongSlitService.resetTelluricConfig(targetOid)
```
swalker2m: This is fine, but just in case services also come into scope with importing `Services.Syntax.*` so they could be accessed un-prefixed.

swalker2m (approved):
Looks great, seems like a good idea to me. :+1:

PR #2474, opened 2026-05-19
Fixes a bug when cloning a telluric group science

cquiroz:
This is a negative side effect of a previous PR to avoid creating phantom groups

You may want to check this hiding whitespace


rpiaggio (approved):
LGTM

PR #2475, opened 2026-05-20
Gnirs Long Slit observing mode

rpiaggio:
Adds GNIRS long slit to DB, GraphQL API.

Igrins2OffsetMode was generalized to SlitOffsetMode.


modules/schema/src/main/scala/lucuma/odb/format/TelescopeConfigsFormat.scala
```diff
@@ -0,0 +1,49 @@
...
+ * JSON blob (column name varies by instrument, e.g. `c_telescope_configs` for GNIRS).
+ */
+trait TelescopeConfigsFormat:
+  import lucuma.odb.json.offset.transport.given
+  import lucuma.odb.json.stepconfig.given
+  import lucuma.odb.json.telescopeConfigAlongSlit.transport.given
+
+  val AlongSlitFormat: Format[String, NonEmptyList[TelescopeConfigAlongSlit]] =
+    Format(
+      s => io.circe.parser.decode[List[TelescopeConfigAlongSlit]](s).toOption.flatMap(NonEmptyList.fromList),
```
cquiroz: why using `fqn`?

rpiaggio: > why using `fqn`?

Because Claude apparently doesn't know how to `import`.

modules/service/src/main/resources/db/migration/V1146__gnirs_long_slit.sql
```diff
@@ -0,0 +1,339 @@
...
+    m.c_mode_key,
+    o.c_program_id,
+    max(m.c_observation_id) as c_observation_id
+  FROM
+    t_igrins_2_long_slit m
+  JOIN t_observation o ON m.c_observation_id = o.c_observation_id
+  GROUP BY
+    m.c_mode_key,
+    o.c_program_id
+-- GNIRS LongSlit
```
cquiroz: no ghost?

modules/service/src/main/scala/lucuma/odb/service/CalibrationConfigSubset.scala
```diff
@@ -52,6 +53,10 @@ object CalibrationConfigSubset:
   case object GhostConfigs extends CalibrationConfigSubset derives Eq:
     def modeType: ObservingModeType = ObservingModeType.GhostIfu
 
+  // TODO: What do we need here?
```
cquiroz: In the case gnirs I think we have tellurics thus this doesn't really matter a lot, you may want to make it equivalent to igrins2

modules/service/src/main/scala/lucuma/odb/service/GnirsLongSlitService.scala
```diff
@@ -0,0 +1,543 @@
...
+        // Read mode default + explicit
+        gnirs_obs_read_mode              *: // c_read_mode_default
+        gnirs_obs_read_mode.opt          *: // c_read_mode (explicit)
+        // Well depth default + explicit
+        gnirs_well_depth                 *: // c_well_depth_default
+        gnirs_well_depth.opt             *: // c_well_depth (explicit)
+        // Focus motor steps (nullable = Best)
+        int4.opt                         *:
+        // Telescope configs: explicit (both nullable) + default
+        slit_offset_mode.opt   *: // c_slit_offset_mode (explicit)
```
cquiroz: missaligned 😄 
if we just had a tool to do that

modules/service/src/main/scala/lucuma/odb/service/GnirsLongSlitService.scala
```diff
@@ -0,0 +1,543 @@
...
+              acqReadMode *: acqCoadds *: acqFilter *: acqOffP *: acqOffQ *:
+              acqExpTime *: acqExpCount *: acqExpAt *: EmptyTuple) =>
+          SlitTelescopeConfigsFormat.getOption((slitOffsetModeDef, tcDef))
+            .toRight(s"Could not parse default telescope configs from '$tcDef'")
+            .flatMap: defaultTC =>
+              val explicitTC = (slitOffsetModeExp, tcExp).mapN: (mode, json) =>
+                SlitTelescopeConfigsFormat.getOption((mode, json))
+                  .toRight(s"Could not parse explicit telescope configs from '$json'")
+              explicitTC.sequence.flatMap { explicitTCOpt =>
+                PosInt.from(coadds)
```
cquiroz, line 124: Just a matter of taste. I'd put all the validations in one line and then do patter matching to detect the bad cases

modules/CLAUDE.md
```diff
@@ -0,0 +1,264 @@
+# lucuma-odb Contributor Guide (CLAUDE)
```
cquiroz, line 1: was this fully generated? maybe it should go to the `services` dir and be only odb orientedu/

cquiroz (commented):
In general seems good to me, I'm not sure if we need a second look at the migrations, there is a lot going on there

modules/service/src/main/scala/lucuma/odb/service/ConfigurationService.scala
```diff
@@ -411,6 +417,10 @@ object ConfigurationService {
                         flamingos2LongSlit {
                           disperser
                         }
+                        gnirsLongSlit {
```
toddburnside, line 420: Some of the queries have `grating`, `camera`, and `prism`, and others don't have the `grating`. Is this correct? There's a bit too much here for me to follow completely.

toddburnside (approved):
Too big to fully review, but it seems fine to me.
Could you add some replies to the copilot review comments as to whether they have been addressed or don't apply for some reason?

modules/service/src/main/resources/db/migration/V1146__gnirs_long_slit.sql
toddburnside: Good question. 
Also, this seems much more complicated to follow than the `v_all_modes` and `v_observing_mode_group` implementation in the V1119 migration. This will be much harder to maintain for new modes. Is there a reason for the change?

modules/service/src/main/resources/db/migration/V1146__gnirs_long_slit.sql
```diff
@@ -0,0 +1,339 @@
...
+  JOIN t_observation o ON m.c_observation_id = o.c_observation_id
+  GROUP BY
+    m.c_mode_key,
+    o.c_program_id;
+
+-- Register mode for observing mode consistency trigger
+SELECT register_observing_mode('gnirs_long_slit', 't_gnirs_long_slit');
+
+-- Update check_etm_consistent to handle gnirs_long_slit:
+-- acquisition ETM is stored inline in t_gnirs_long_slit; only one science ETM in t_exposure_time_mode
```
toddburnside: Why is the acquistion ETM stored `inline`?

cquiroz: Good question, don't we have a separate table for etms?

rpiaggio: Didn't realize that, will change.

modules/service/src/main/resources/db/migration/V1146__gnirs_long_slit.sql
rpiaggio: I fixed this now to follow the same pattern as in previous migrations

modules/service/src/main/scala/lucuma/odb/service/GnirsLongSlitService.scala
rpiaggio: I'm not sure what you mean. Can you point to an example?

modules/CLAUDE.md
rpiaggio: Thing is, we should restructure the repo to separate `odb`/`itc`/`sso` into separate folders. Right now, `odb` and `sso` are mixed in `modules` (with `sso` projects having a prefix). I instructed to generate this file ignoring `itc` and `sso`, only for `odb`. Still, I'm moving it into `modules`

modules/service/src/main/scala/lucuma/odb/service/ConfigurationService.scala
rpiaggio: They should all have `grating`, thanks for the catch. Adding.

modules/service/src/main/resources/db/migration/V1146__gnirs_long_slit.sql
```diff
@@ -0,0 +1,339 @@
+-- GNIRS Long Slit observing mode: lookup tables, main table, view, mode registration,
+-- configuration request columns, and ETM trigger update.
+
+-- Lookup table for GnirsDecker
+CREATE TABLE t_gnirs_decker (
```
toddburnside: Do these need to be lookup tables? Or could they be postgres enums? We use scala enums for the short and long names, don't we?

rpiaggio: I think they can be enums, yes. Let me check.

rpiaggio:
> Too big to fully review, but it seems fine to me. Could you add some replies to the copilot review comments as to whether they have been addressed or don't apply for some reason?

Yes, copilot caught:
- `coadds` appreaed as nullable in the API but it wasn't nullable in the DB.
- The ObservingMode.Acquisition was being ignored in an edit and not applied.
- A test misname and outdated comments.
- A bug in the logic to clear up the offsets and return them to the default.

All of these have been addressed.

modules/schema/src/main/scala/lucuma/odb/format/TelescopeConfigsFormat.scala
```diff
@@ -0,0 +1,50 @@
...
+import lucuma.core.enums.SlitOffsetMode
+import lucuma.core.model.SlitTelescopeConfigs
+import lucuma.core.model.sequence.TelescopeConfig
+import lucuma.core.model.sequence.TelescopeConfigAlongSlit
+import lucuma.core.optics.Format
+
+/**
+ * Serialization layer between `SlitTelescopeConfigs` (the lucuma-core domain type) and the two
+ * columns used to persist it in the DB: a discriminant tag (e.g. `c_slit_offset_mode`) and a
+ * JSON blob (column name varies by instrument, e.g. `c_telescope_configs` for GNIRS).
```
swalker2m, line 18: Could they be made the same everywhere?  Are all the long slit modes the same wrt offsetting along the slit?

modules/service/src/main/resources/db/migration/V1146__gnirs_long_slit.sql
```diff
@@ -0,0 +1,319 @@
...
+          '[{"q":{"microarcseconds":2000000},"guiding":"ENABLED"},{"q":{"microarcseconds":-4000000},"guiding":"ENABLED"},{"q":{"microarcseconds":-4000000},"guiding":"ENABLED"},{"q":{"microarcseconds":2000000},"guiding":"ENABLED"}]'
+        WHEN ls.c_filter IN ('Order2', 'Order1', 'PAH') THEN
+          '[{"q":{"microarcseconds":-3000000},"guiding":"ENABLED"},{"q":{"microarcseconds":3000000},"guiding":"ENABLED"},{"q":{"microarcseconds":3000000},"guiding":"ENABLED"},{"q":{"microarcseconds":-3000000},"guiding":"ENABLED"}]'
+        ELSE
+          '[{"q":{"microarcseconds":-1000000},"guiding":"ENABLED"},{"q":{"microarcseconds":5000000},"guiding":"ENABLED"},{"q":{"microarcseconds":5000000},"guiding":"ENABLED"},{"q":{"microarcseconds":-1000000},"guiding":"ENABLED"}]'
+      END AS c_telescope_configs_default
+  ) d;
+
+-- Dedicated views joining t_exposure_time_mode with t_gnirs_long_slit by FK.
+-- Same pattern as V1120__ghost.sql; these support per-role Grackle mappings.
```
swalker2m: This was necessary in the case of GHOST because there were two ETMs for science (red vs blue).  For GNIRS, the exposure time mode role acquisition vs science is probably fine?  I don't think we do anything like this for GMOS or F2 long slit?

swalker2m (approved):
I didn't check every line carefully but it all looks great.  The only question is whether we really need separate views for the science vs. acquisition ETMs for GNIRS.

modules/schema/src/main/scala/lucuma/odb/format/TelescopeConfigsFormat.scala
rpiaggio: The idea is to eventually use this approach for all longslit modes, yes.

modules/service/src/main/resources/db/migration/V1146__gnirs_long_slit.sql
rpiaggio: Ah, hadn't noticed the role column in the ETM table. I'll rework to use that.

PR #2477, opened 2026-05-21
Use custom schema loader to support import statements for graphql schema validations

hugo-vrijswijk:
Replaces the GitHub action used for validating GraphQL schema changes with a similar implementation with a custom "loader" (`LucumaResourceLoader`). This loader supports the import syntax we use in JVM resource .graphql files. Most of the code for checks still comes from the original action, which is used as a dependency.

Adds some code similar to the action that will add annotations to the pull request on breaking schema changes, and also supports the "expected-breaking-change" label to allow merging breaking changes when the label is present.

### DIFF
--- a/.github/validate-schema.mjs
+++ b/.github/validate-schema.mjs
@@ -1,0 +1,257 @@
+//
+// This file has _most_ of the functionality of the validate-schema action, but the main difference with the original action is that it supports loading schema files from JVM resource directories (in the same project). This is done by the LucumaResourceLoader
+// Annotations are supported and added during GitHub actions check using GHA annotations.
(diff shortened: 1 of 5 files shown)


cquiroz:
This is very cool, seems to be in the right direction. would it work even without you previous #2462 pr?

cquiroz (approved):
LGTM

toddburnside (approved):
Great idea!

hugo-vrijswijk:
> This is very cool, seems to be in the right direction. would it work even without you previous #2462 pr?

Yes, it does. Resource already imports from OdbSchema with import statements. But with _both_ of them the schema stitcher script could be removed for itc. 

The whole thing turned out to be a lot more code than I expected, but I am glad it works

PR #2479, opened 2026-05-21
Remove schema stitching scripts from ITC and use import statements instead

hugo-vrijswijk:
Expected-breaking-change because of all the removed (unused) types in the itc schema. If anything is removed that was used, things would break


cquiroz (approved):
i like it. -10k loc 😄

PR #2480, opened 2026-05-21
Update basePosition to return fields appropriate for navigate tracking

cquiroz:
This is a breaking change that changes `basePosition` such it can be used by navigate to slew and track targets that would be unfied for single/muti targets as well as supporting non-sidereals

I couldn't find any usages of the existing basePosition in `explore` so far.


toddburnside (approved):
LGTM

PR #2481, opened 2026-05-22
Add macro to validate and stitch schema at compile-time

hugo-vrijswijk:
Adds a macro that validates the schema at compile time, checking that the resource exists, that it can be loaded and stitched together without errors, and if it can be loaded, it embeds the final schema string in the code, so it doesn't have to be stitched again at runtime.

Also adds a utility method to load a schema with logging and error handling.

Examples:

image

image

### DIFF
--- a/.scalafix.conf
+++ b/.scalafix.conf
@@ -1,3 +1,3 @@
 include ".scalafix-common.conf"
-OrganizeImports.removeUnused = false
+OrganizeImports.removeUnused = true
 OrganizeImports.targetDialect = Scala3
--- a/modules/binding/src/main/scala/lucuma/odb/graphql/schema/SchemaSource.scala
+++ b/modules/binding/src/main/scala/lucuma/odb/graphql/schema/SchemaSource.scala
@@ -3,34 +3,23 @@
 
 package lucuma.odb.graphql.schema
 
-import cats.effect.Sync
 import cats.syntax.eq.*
-import fs2.RaiseThrowable
-import fs2.Stream
-import fs2.io
 import fs2.io.file.Path
-import fs2.text
 
-trait SchemaSource[F[_]]:
-  def resolve(name: Path): Stream[F, String]
+import java.lang.System.lineSeparator
+import scala.io.Source
+
+trait SchemaSource:
+  def resolve(name: Path): List[String]
 
 object SchemaSource:
-  def fromResource[F[_]: Sync]: SchemaSource[F] = (name: Path) =>
-    Stream
-      .eval(Sync[F].blocking(Option(getClass.getClassLoader.getResourceAsStream(name.toString))))
-      .flatMap:
-        case Some(resource) =>
-          io.readInputStream(Sync[F].pure(resource), 8192)
-            .through(text.utf8.decode)
-            .through(text.lines)
-        case None           => Stream.raiseError[F](new io.IOException(s"Resource $name not found"))
+  def fromResource(classloader: ClassLoader): SchemaSource = (name: Path) =>
+    Source.fromResource(name.toString, classloader).getLines().toList
 
-  def fromString[F[_]: RaiseThrowable](name: Path, content: String): SchemaSource[F] = (n: Path) =>
-    if (n === name) Stream.emit(content).through(text.lines)
-    else Stream.raiseError[F](new Error(s"Unknown source $n"))
+  def fromString(name: Path, content: String): SchemaSource = (n: Path) =>
+    if (n === name) content.split(lineSeparator).toList
+    else throw new RuntimeException(s"Unknown source $n")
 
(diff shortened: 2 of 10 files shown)


.scalafix.conf
```diff
@@ -1,3 +1,3 @@
 include ".scalafix-common.conf"
-OrganizeImports.removeUnused = false
+OrganizeImports.removeUnused = true
```
cquiroz, line 2: Seems like it should be a default everywhere

hugo-vrijswijk: I think it was broken for Scala 3 for a while, but yes I agree

modules/binding/src/main/scala/lucuma/odb/graphql/schema/SchemaSource.scala
```diff
@@ -3,34 +3,24 @@
...
-    Stream
-      .eval(Sync[F].blocking(Option(getClass.getClassLoader.getResourceAsStream(name.toString))))
-      .flatMap:
-        case Some(resource) =>
-          io.readInputStream(Sync[F].pure(resource), 8192)
-            .through(text.utf8.decode)
-            .through(text.lines)
-        case None           => Stream.raiseError[F](new io.IOException(s"Resource $name not found"))
+  def fromResource(classloader: ClassLoader): SchemaSource = (name: Path) =>
+    classloader.getResource(name.toString) match
```
cquiroz: `Option(...).fold`?

modules/binding/src/main/scala/lucuma/odb/graphql/schema/SchemaSource.scala
```diff
@@ -3,34 +3,24 @@
...
+      case null => throw new Error(s"Resource $name not found")
+      case str  => Source.fromResource(name.toString).getLines().toList
 
-  def fromString[F[_]: RaiseThrowable](name: Path, content: String): SchemaSource[F] = (n: Path) =>
-    if (n === name) Stream.emit(content).through(text.lines)
-    else Stream.raiseError[F](new Error(s"Unknown source $n"))
+  def fromString(name: Path, content: String): SchemaSource = (n: Path) =>
+    if (n === name) content.split("\n").toList
```
cquiroz: I'm always a bit nervous when separators are hardcoded, though I guess we never intend to run this in windows?

modules/binding/src/main/scala/lucuma/odb/graphql/schema/SchemaSource.scala
```diff
@@ -3,34 +3,24 @@
...
+    else throw new Error(s"Unknown source $n")
 
-  def fromStringMap[F[_]: RaiseThrowable](m: Map[Path, String]): SchemaSource[F] = (name: Path) =>
+  def fromStringMap(m: Map[Path, String]): SchemaSource = (name: Path) =>
     m.get(name)
-      .fold(Stream.raiseError[F](new Error(s"Unknown source $name")))(x =>
-        Stream.emit(x).through(text.lines)
-      )
+      .fold(throw new Error(s"Unknown source $name"))(_.split("\n").toList)
```
cquiroz: `RuntimeException`?

cquiroz (approved):
Seems very good though I can't really review the macro itself

Added a few comments inline

rpiaggio:
I like automating and checking all of this.

- Will this run in every compilation?
- Was a scalafix rule considered instead of a macro? If so, why was the macro approach favored?

hugo-vrijswijk:
> Will this run in every compilation?

Yes, but sbt 2 will fix everything ✨. Before, it ran on every startup (or even every request).

> Was a scalafix rule considered instead of a macro? If so, why was the macro approach favored?

I did not consider it, but I don't know how to write scalafix rules. It to me also seems a lot more heavyweight than a macro. Most of the macro code is the same as the runtime code before, but returning a piece of code to be inserted instead of a schema. IMO the macro itself is quite understandable (and you can CTRL-click to everything, which I'm not sure would work with a scalafix rule).

rpiaggio (approved):
LGTM

modules/binding/src/main/scala/lucuma/odb/graphql/schema/SchemaSource.scala
hugo-vrijswijk: I've updated it to use System.lineSeparator

modules/binding/src/main/scala/lucuma/odb/graphql/schema/SchemaSource.scala
hugo-vrijswijk: Good suggestion

modules/binding/src/main/scala/lucuma/odb/graphql/schema/SchemaSource.scala
hugo-vrijswijk: Source.fromResource does the same check, so I've removed the option match

PR #2483, opened 2026-05-22
ProposalStatus and FilterType no longer dynamic

toddburnside:
Also removes the metadata endpoint because it is no longer used. Explore already uses the scala enums


modules/service/src/main/scala/lucuma/odb/service/ProposalService.scala
```diff
@@ -201,26 +193,23 @@ object ProposalService {
...
         hasProposal:       Boolean,
         piEmailStr:        Option[NonEmptyString],
         title:             Option[NonEmptyString],
         reference:         Option[ProposalReference],
         semester:          Option[Semester],
         scienceSubtype:    Option[ScienceSubtype],
         splitsSum:         Long,
-        availablePartners: Set[Tag],
-        requestedPartners: Set[Tag],
+        availablePartners: Set[Partner],
```
toddburnside, line 204: Getting the partners from the DB as Partners instead of Tags is an unrelated change that will be useful when we need to send proposal notifications to the partners. (soon...)

modules/service/src/main/scala/lucuma/odb/service/ObservationWorkflowService.scala
```diff
@@ -391,66 +383,62 @@ object ObservationWorkflowService {
...
-              case Completed  => if info.isDeclaredComplete then List(Ongoing) else Nil
-
-          (state, allowedTransitions)
-
-        }
+      ): (ObservationWorkflowState, List[ObservationWorkflowState]) =
+        // A special ordering where codes are ordered as they would occur in a typical lifecycle.
+        given Ordering[ObservationValidationCode] =
+          Ordering.by:
+            case ObservationValidationCode.CallForProposalsError => 1
```
cquiroz, line 390: we coud automatically align code...

cquiroz (approved):
Great work!

modules/service/src/main/scala/lucuma/odb/service/ObservationWorkflowService.scala
toddburnside: That's crazy talk. 😜

PR #2484, opened 2026-05-22
Set obsclass for tellurics and specphoto correctly

cquiroz:
I took this approach of updating the obs class after the generation but could be also done setting the class at creation time. This way seems to me that is easier to apply to all instruments that need it


swalker2m (approved):
This is clever and reminds me of a similar post-processing step in the `ocs` sequence code.  I think it's well done.  Since this is a map applied to a `Stream` I don't think it adds overhead.  I'll approve it, but I was hesitant as it does introduce a new place to look and a source of confusion about where values are being set.  For GMOS for example we set the `ObserveClass` here instead for twilights: https://github.com/gemini-hlsw/lucuma-odb/blob/main/modules/sequence/src/main/scala/lucuma/odb/sequence/gmos/longslit/Science.scala#L291-L293

PR #2485, opened 2026-05-22
Allow introspection without authorization

toddburnside:
It turns out the metadata service was doing double duty for allowing introspection without authorization... 😢 

This creates an even more minimal service that only allows introspection. There may be a better way of doing this, but this is what I know how to do and it gets us back up and running for downloading the schema and using the playground.


cquiroz (approved):
LGTM
Did you try with the playground?

toddburnside:
> Did you try with the playground?

The playground is working for me, so... I was also able to use the schema fetching script in lucuma-schemas.

PR #2486, opened 2026-05-23
Provide dummy metadata endpoint for explore service workers

toddburnside:
Unfortunately, the older explore service workers seem to need to be able to download this metadata `before` they can update themselves. So we will have an endpoint that simply sends the old data. I guess someday in the future we can delete this...


cquiroz (approved):
LGTM

PR #2489, opened 2026-05-23
Igrins2 name and maroonx sky fiber area

cquiroz:
Update odb to bring two changes on lucuma core.

* Add geometric model and ags params for Maroon-X
* Add a dash to the public name of igrins 2


rpiaggio (approved):
LGTM

PR #2491, opened 2026-05-24
Initial Gnirs science sequence generation

rpiaggio:
- Generates a simple science sequence with varying offsets. No instrument configuration is changed throughout the sequence. I don't know if Gnirs requires this. I tried to determine it from OCS, where there doesn't seem to be time estimates for configuration changes in Gnirs (at least Claude couldn't find them).
- Sequence is not persisted yet.
- Neither is there visit recording capability yet.
- Initial configuration implementation tries to mimic Gmos so that code can be reused once we get to implement imaging and IFU.


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
swalker2m, line 6663: At least in my view, these GraphQL schema warnings don't seem to line up with the actual content of the message.  Here it notes `estimate` was added to `GnirsStep` and attaches it to `BandNormalizedSurface`.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -8107,6 +8113,214 @@ type Igrins2Step {
...
+type GnirsExecutionSequence {
+  """
+  Next atom to execute.
+  """
+  nextAtom: GnirsAtom!
+
+  """
+  (Prefix of the) remaining atoms to execute, if any.
+  """
+  possibleFuture: [GnirsAtom!]!
```
swalker2m, line 8148: This is the correct way to do it for now.  Someday, I'd like to combine `nextAtom` and `possibleFuture` into a single list of atoms since it no longer changes without user intervention.

modules/service/src/main/scala/lucuma/odb/logic/GeneratorStreaming.scala
```diff
@@ -250,6 +268,24 @@ object GeneratorStreaming:
           gen <- EitherT.fromEither(LongSlit.instantiate(context.oid, calculator.igrins2Step, context.namespace, cfg, itc))
         yield gen.covary[F]).value
 
+      override def selectOrGenerateGnirsLongSlit(
+        context: GeneratorContext
+      )(using Transaction[F]): F[Either[OdbError, StreamingExecutionConfig[F, GnirsStatic, GnirsDynamic]]] =
+        // GNIRS does not yet have DB-materialized sequences (no
+        // selectGnirsSequence on SequenceService).  Until that lands, we always
+        // generate from scratch — equivalent to "select returns None, generate".
+        generateGnirsLongSlit(context)
```
swalker2m, line 277: Seems fine for now.  Manual sequence editing of course will not work.

swalker2m (approved):
This all looks great to me. :+1:

PR #2492, opened 2026-05-24
Require staff access to modify gnirs focus

rpiaggio:


modules/service/src/main/scala/lucuma/odb/graphql/mapping/AccessControl.scala
```diff
@@ -459,7 +461,10 @@ trait AccessControl[F[_]] extends Predicates[F] {
 
   def selectForUpdate(
     input: CreateObservationInput,
-  )(using Services[F]): F[Result[AccessControl.CheckedWithId[ObservationPropertiesInput.Create, Program.Id]]] = {
+  )(using Services[F]): F[Result[AccessControl.CheckedWithId[ObservationPropertiesInput.Create, Program.Id]]] =
+    if input.SET.exists(_.needsStaffAccess) && user.role.access < Access.Staff then
```
cquiroz, line 465: I wonder if there is an `isStaff` method in role or access, or maybe add it via extension

cquiroz (approved):
LGTM

modules/service/src/main/scala/lucuma/odb/graphql/mapping/AccessControl.scala
rpiaggio: Since `Access` is ordered, we would need something like `isAtLeastStaff`. But I don't think that gives better legibility than `<` (which is also more general).

PR #2494, opened 2026-05-25
Gnirs acquisiton type

rpiaggio:
Although very similar, Gnirs acquisition type is a distinct enum from read mode. This PR fixes the former confusion.


modules/service/src/main/resources/db/migration/V1150__gnirs_acq_type.sql
```diff
@@ -0,0 +1,81 @@
...
+-- New enum: classifies acquisition target brightness (VeryBright / Bright / Faint)
+CREATE TYPE e_gnirs_acquisition_type AS ENUM ('VeryBright', 'Bright', 'Faint');
+
+-- Drop the view first; it expands ls.* at creation time so it holds c_acq_read_mode implicitly.
+DROP VIEW v_gnirs_long_slit;
+
+-- Swap the column on the underlying table.
+ALTER TABLE t_gnirs_long_slit DROP COLUMN c_acq_read_mode;
+ALTER TABLE t_gnirs_long_slit
+  ADD COLUMN c_acq_type e_gnirs_acquisition_type NOT NULL DEFAULT 'Faint';
```
swalker2m, line 13: Should we use the existing `c_acq_read_mode` to initialize the new `c_acq_type`?  Or does it get initialized based on the asterism elsewhere?

swalker2m (approved):
This looks fine, but what will be the purpose of the acquisition type?  Will it change something about the acquisition sequence generation?

modules/service/src/main/resources/db/migration/V1150__gnirs_acq_type.sql
rpiaggio: From what I gathered after talking to Andy, for initial acquisition we will use S/N query to ITC and then initialize this parameter from there.

rpiaggio:
> This looks fine, but what will be the purpose of the acquisition type? Will it change something about the acquisition sequence generation?

That's a good question for Andy, I'll forward it.

PR #2495, opened 2026-05-25
Fix using TagBinding for ProposalStatus

cquiroz:


modules/service/src/main/scala/lucuma/odb/graphql/input/WhereProgram.scala
```diff
@@ -24,7 +24,7 @@ object WhereProgram {
     val WhereTypeBinding             = WhereEq.binding[ProgramType](path / "type", ProgramTypeBinding)
     val WhereProgramReferenceBinding = WhereProgramReference.binding(path / "reference")
     val WherePiBinding               = WhereProgramUser.binding(path / "pi", ProgramUserRole.Pi.some)
-    val WhereEqProposalStatus        = WhereUnorderedTag.binding(path / "proposalStatus", TagBinding)
+    val WhereEqProposalStatus        = WhereUnorderedTag.binding(path / "proposalStatus", ProposalStatusBinding)
```
swalker2m: What does `WhereUnorderedTag` give us?  Why not just `WhereEq`?

swalker2m (approved):

modules/service/src/main/scala/lucuma/odb/graphql/input/WhereProgram.scala
swalker2m:
```scala
object WhereEq:

  def binding[A: Eq](path: Path, binding: Matcher[A]): Matcher[Predicate] =
    ObjectFieldsBinding.rmap {
      case List(
        binding.Option("EQ", rEQ),
        binding.Option("NEQ", rNEQ),
        binding.List.Option("IN", rIN),
        binding.List.Option("NIN", rNIN)
      ) =>
        (rEQ, rNEQ, rIN, rNIN).parMapN {
          (EQ, NEQ, IN, NIN) =>
            and(List(
              EQ.map(a => Eql(path, Const(a))),
              NEQ.map(a => NEql(path, Const(a))),
              IN.map(as => In(path, as)),
              NIN.map(as => Not(In(path, as)))
            ).flatten)
        }
    }
```

Everything that has an `Order` instance has an `Eq` instance so given the above it doesn't seem like we need this:

```scala
object WhereUnorderedTag {

  def binding[A: Order](path: Path, binding: Matcher[A]): Matcher[Predicate] =
    ObjectFieldsBinding.rmap {
      case List(
        binding.Option("EQ", rEQ),
        binding.Option("NEQ", rNEQ),
        binding.List.Option("IN", rIN),
        binding.List.Option("NIN", rNIN)
      ) =>
        (rEQ, rNEQ, rIN, rNIN).parMapN {
          (EQ, NEQ, IN, NIN) =>
            and(List(
              EQ.map(a => Eql(path, Const(a))),
              NEQ.map(a => NEql(path, Const(a))),
              IN.map(as => In(path, as)),
              NIN.map(as => Not(In(path, as)))
            ).flatten)
        }
    }
}
```

cquiroz: Right I just renamed the bindig but there is no need for the Unorderd part. I'll remove it

swalker2m: :+1:

PR #2496, opened 2026-05-26
Add plan time calculation for resident visitors

cquiroz:


modules/sequence/src/main/scala/lucuma/odb/sequence/visitor/VisitorTimeCalculator.scala
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * Computes the time estimate for a resident visitors (Alopeke, Zorro, maroon-x).
+ * There is no sequence for these modes. We get the teotal time as:
+ *
+ * total time is `setup + count * (exposureTime + readout)` derived
+ * from the single science ExposureTimeMode row.
+ *
+ * Acquisition and reacquisition times are zero for resident visitors.
+ */
+object VisitorTimeCalculator:
```
swalker2m: Since this thing creates an `ExecutionDigest`, maybe it should be the `VisitorExecutionDigestCalculator`?

swalker2m (approved):
Looks good.

PR #2499, opened 2026-05-26
Update skunk to v1

hugo-vrijswijk:
Only possible with a Skunk snapshot for now unfortunately.


cquiroz:
Why a snapshot? i thought v1 was already released

hugo-vrijswijk:
> Why a snapshot? i thought v1 was already released

It is, but 1.0 has a bug that is fixed in 1.1.0-RC1. Even with that fix, Skunk v1 is built against Otel4s 0.16. Since we're already using Otel4s 1 we get compatibility scheme errors. So until there is a Skunk version that is built against Otel4s 1 we can't merge this (probably Skunk v2)

PR #2501, opened 2026-05-26
Fix line mapping on schema validation

hugo-vrijswijk:


itc/service/src/main/resources/graphql/itc.graphql
```diff
@@ -53,6 +53,10 @@ input GmosSSpectroscopyInput {
   Exposure time mode.
   """
   exposureTimeMode: ExposureTimeModeInput!
+  """
+  Exposure time mode.
+  """
+  exposureTimeMode2: ExposureTimeModeInput!
```
cquiroz: what is this?

hugo-vrijswijk: Adding a breaking change to test if the line gets reported correctly. I'll put this back on draft until I've confirmed it works

cquiroz: i think adding a field shouldn't be flagged as a breaking change, renaming or deleting should

hugo-vrijswijk: Adding a field to a regular type is not, but adding a required field to an input type is (since existing inputs will break):

> Adding a required input field to an existing input object type is a breaking change because it will cause existing uses of this input object type to error.

PR #2507, opened 2026-05-27
Allow change observing mode even with an edited sequence

toddburnside:
This addresses the explicit request in https://app.shortcut.com/lucuma/story/8094/error-reverting-configuration-with-edited-sequence?vc_group_by=day&ct_workflow=all&cf_workflow=500000071 of being able to revert a configuration even if the sequence has been edited, with the added bonus of allowing a different observing mode as long as it is a different instrument.

A caveat: Updating with the same instrument but different mode (spectroscopy vs imaging) will result in the new mode having a sequence for the old mode. Changes I am making in explore will prevent this, but it could still happen via the API. I figure we can address this if/when it becomes an issue.

@swalker2m I'm not really familiar with the edited sequence thing. Could you tell me if I'm missing something or doing something wrong?

### DIFF
--- a/modules/service/src/main/resources/db/migration/V1152__clear_sequence_on_instrument_change.sql
+++ b/modules/service/src/main/resources/db/migration/V1152__clear_sequence_on_instrument_change.sql
@@ -1,0 +1,36 @@
+-- When an observation's instrument changes, the previously materialized
+-- sequence and any manually-defined atoms become invalid.
+
+-- Make the FK from t_atom back to t_observation deferrable so the trigger
+-- can run within the same transaction without a constraint violation.
+ALTER TABLE t_atom
+  DROP CONSTRAINT t_atom_c_observation_id_c_instrument_fkey,
+  ADD  CONSTRAINT t_atom_c_observation_id_c_instrument_fkey
+    FOREIGN KEY (c_observation_id, c_instrument)
+    REFERENCES t_observation(c_observation_id, c_instrument)
+    ON DELETE CASCADE DEFERRABLE INITIALLY DEFERRED;
+
+
+CREATE OR REPLACE FUNCTION clear_sequence_on_instrument_change()
+  RETURNS TRIGGER AS $$
+BEGIN
+  DELETE FROM t_sequence_materialization
+   WHERE c_observation_id = OLD.c_observation_id;
+
+  -- Other foreign key constraints will prevent deletion of executed sequences.
+  DELETE FROM t_atom
+   WHERE c_observation_id = OLD.c_observation_id
+     AND c_instrument      = OLD.c_instrument;
+
+  RETURN NEW;
+END;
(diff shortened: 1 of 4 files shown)


modules/service/src/main/resources/db/migration/V1152__clear_sequence_on_instrument_change.sql
```diff
@@ -0,0 +1,36 @@
...
+  DELETE FROM t_atom
+   WHERE c_observation_id = OLD.c_observation_id
+     AND c_instrument      = OLD.c_instrument;
+
+  RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER clear_sequence_on_instrument_change_trigger
+AFTER UPDATE OF c_instrument ON t_observation
```
swalker2m, line 30: Should it perhaps be `AFTER UPDATE OF c_observing_mode_type ON t_observation`?  That would handle changing modes but not instruments?

modules/service/src/main/resources/db/migration/V1152__clear_sequence_on_instrument_change.sql
```diff
@@ -0,0 +1,36 @@
...
+  RETURNS TRIGGER AS $$
+BEGIN
+  DELETE FROM t_sequence_materialization
+   WHERE c_observation_id = OLD.c_observation_id;
+
+  -- Other foreign key constraints will prevent deletion of executed sequences.
+  DELETE FROM t_atom
+   WHERE c_observation_id = OLD.c_observation_id
+     AND c_instrument      = OLD.c_instrument;
+
```
swalker2m, line 24: I think these are the things that doing.   I guess we're relying on permissions blocking you from even attempting this if the observation is ongoing or executed?

Because multiple tables are being manipulated and I wanted to avoid ending up in an inconsistent state, I would lock observations for update in all the manual sequence manipulation code.

```
  PERFORM 1 FROM t_observation WHERE c_observation_id = p_observation_id FOR UPDATE;
```

I think if we do this, we should also add that here at the top.

swalker2m:
I think I might be inclined to add a mutation to abandon manual edits and then explicitly call that from the client instead of automatically wiping it out.  We could expose in the API, as we were discussing, whether the sequence has been materialized (and without execution data).

toddburnside:
> I think I might be inclined to add a mutation to abandon manual edits and then explicitly call that from the client instead of automatically wiping it out. We could expose in the API, as we were discussing, whether the sequence has been materialized (and without execution data).

I think that would be a better solution. It is also actually better in explore from a UI perspective:

- The user can undo the edits without entirely reverting the observing mode.
- We can better handle selecting an existing configuration from the dropdown to overwrite the current one.

I'll close this one.

toddburnside:
Closing in deference to a different solution as discussed above.

PR #2509, opened 2026-05-28
Add enum modes to the schema

cquiroz:
some enum values were not included in the schema, This doesn't break explore nor the odb itself as we know the enum values but affects python clients


swalker2m (approved):
:+1:

PR #2512, opened 2026-05-28
Add Gnirs record test

rpiaggio:


modules/service/src/test/scala/lucuma/odb/graphql/mutation/recordVisit.scala
```diff
@@ -445,6 +445,45 @@ class recordVisit extends OdbSuite with query.GenerationTestSupport with Executi
...
+      oid => s"""
+        mutation {
+          recordVisit(input: {
+            observationId: "$oid",
+          }) {
+            visit {
+              gnirs {
+                wellDepth
+              }
+              igrins2 {
```
cquiroz, line 462: why are we checking igrins2 and gmos?

rpiaggio: It's the same pattern than for other instruments. I guess it's to check that they actually return null.

cquiroz (approved):
LGTM

PR #2513, opened 2026-05-28
Include a groupId in group events

cquiroz:
This is useful to tell which group was actually deleted on the UI


toddburnside:
Isn't the id already part of `Group`?

cquiroz:
> Isn't the id already part of `Group`?

Yes but on hard deletes value is `null`

toddburnside:
> > Isn't the id already part of `Group`?
> 
> Yes but on hard deletes value is `null`

Ah. Fun.

toddburnside (approved):
LGTM

PR #2514, opened 2026-05-28
Unify Gnirs with F2 logic for automatic ReadMode handling

rpiaggio:
I had implemented an extra enum that just added `Automatic` option, but that isn't necessary if we follow the F2 path, making the logic consistent in both instruments.


swalker2m (approved):
Looks great. :+1:

PR #2516, opened 2026-05-29
Classical proposal api updates

cquiroz:
don't merge until explore is ready


modules/service/src/main/scala/lucuma/odb/service/ProposalService.scala
```diff
@@ -240,10 +240,8 @@ object ProposalService {
                  (s === ScienceSubtype.Queue))
               )
             },
-            missingConsiderForBand3(pid).asFailure.whenA(
-              (scienceSubtype.contains(ScienceSubtype.Queue) || scienceSubtype.contains(ScienceSubtype.Classical))
-              && considerForBand3.contains(ConsiderForBand3.Unset)
-            ),
```
swalker2m, line 246: If we were to switch from queue to classical, does the `c_consider_for_band_3` field in `t_proposal` get set to `null`?

swalker2m (approved):
Looks great. :+1:

modules/service/src/main/scala/lucuma/odb/service/ProposalService.scala
cquiroz: Good question, i'll add that

PR #2518, opened 2026-05-29
Gnirs acquisition sky offset

rpiaggio:
Remodel the Gnirs acquisition sky offset as a single attribute, renames and adds db checks to ensure both axes are set or neither. Also, more tests.


swalker2m (approved):
:+1:

PR #2519, opened 2026-05-29
Gnirs acquisition type can now be null (automatic)

rpiaggio:
Just like for read mode, we need the exposure time to compute the acquisition type. We defer this to seq generation time by allowing it to be set to null, which will be interpreted as Automatic.


swalker2m (approved):
Looks great :+1:

PR #2520, opened 2026-05-29
Gnirs central wavelength fix model

rpiaggio:
Gnirs centralwavelength was erroneously modeled as an independent value instead of derived from filter/grating.


swalker2m (approved):
:+1:

PR #2523, opened 2026-05-29
Shortcut 8595: GHOST IFU Mapping

swalker2m:
Adds a `GhostIfuMapping` to the static configuration.  This is derived from the targets and observation properties, if possible.  It uses the geometry from core to accomplish this.

I settled on a suboptimal API instead of an interface-based one because GraphQL (or maybe just Grackle?) doesn't allow two or more interface implementations to have the same name for a field (even though it doesn't appear in the interface itself) and yet have different types.

A future option would be to make automatic adjustments to the observation (position angle mostly) in order find a sky position and/or IFU mapping, but for now this gives Observe something to work with.


modules/sequence/src/main/scala/lucuma/odb/sequence/ghost/ifu/GhostIfuMappingSyntax.scala
```diff
@@ -0,0 +1,128 @@
...
+  extension (g: GhostIfuMapping.type)
+    def derive(
+      ctx:     IfuMappingContext,
+      targets: List[Target]
+    ): Either[String, GhostIfuMapping] =
+      targets match
+        case Nil          => "Cannot derive a GHOST IFU mapping until targets are defined.".asLeft
+        case List(t)      => deriveOneTarget(ctx, t)
+        case List(t1, t2) => deriveDualTarget(ctx, t1, t2)
+        case _            => "Cannot derive a GHOST IFU mapping with more than two targets.".asLeft
```
swalker2m, line 122: In the next PR, I'll need to add an additional check that the two positions are not too close together.

toddburnside (approved):
LGTM. Great work figuring this all out!

modules/binding/src/main/scala/lucuma/odb/data/PosAngleConstraintMode.scala
```diff
@@ -5,9 +5,6 @@ package lucuma.odb.data
 
 import lucuma.core.util.Enumerated
 
-// There is something similar in lucuma-core.  It needs to be updated and then
-// we can switch to that.
-
 enum PosAngleConstraintMode(val dbTag: String):
```
cquiroz, line 8: We do have https://github.com/gemini-hlsw/lucuma-core/blob/main/modules/core/shared/src/main/scala/lucuma/core/model/PosAngleConstraint.scala

modules/sequence/src/main/scala/lucuma/odb/sequence/ghost/ifu/GhostIfuMappingSyntax.scala
```diff
@@ -0,0 +1,128 @@
...
+import lucuma.core.model.sequence.ghost.GhostIfuMapping
+import lucuma.core.util.Timestamp
+
+/**
+ * Syntax for the GhostIfuMapping companion that adds derivation and validation
+ * from observation information.  Particularly, the resolution mode, the
+ * (possible) sky position, the position angle, the observation time, and the
+ * targets.
+ */
+object GhostIfuMappingSyntax:
```
cquiroz, line 26: Not sure if we'd need this logic in explore/observe. if so we may need to move this to core

modules/sequence/src/main/scala/lucuma/odb/sequence/ghost/ifu/IfuMappingContext.scala
```diff
@@ -0,0 +1,27 @@
...
+ * GhostIfuMapping or validate that a GHOST observation has a mapping.
+ */
+case class IfuMappingContext(
+  resolutionMode:     GhostResolutionMode,
+  sky:                Option[Coordinates],
+  posAngleConstraint: PosAngleConstraint,
+  explicitBase:       Option[Coordinates],
+  when:               Timestamp
+):
+  // Maybe we should limit GHOST to Fixed?
```
cquiroz, line 23: why?

modules/service/src/main/resources/db/migration/V1158__ghost.sql
```diff
@@ -0,0 +1,65 @@
...
+  ADD COLUMN  c_ifu2_rv        numeric             NULL,
+  ADD COLUMN  c_ifu2_parallax  d_angle_µas         NULL,
+
+  ADD CONSTRAINT t_ghost_static_consistent_ifu_mapping CHECK (
+    CASE
+      WHEN c_ifu_mapping = 'nonsidereal'     THEN
+        num_nulls(
+          c_ifu1_ra, c_ifu1_dec, c_ifu1_epoch, c_ifu1_pm_ra, c_ifu1_pm_dec, c_ifu1_rv, c_ifu1_parallax,
+          c_ifu2_ra, c_ifu2_dec, c_ifu2_epoch, c_ifu2_pm_ra, c_ifu2_pm_dec, c_ifu2_rv, c_ifu2_parallax
+        ) = 14
```
cquiroz, line 29: 👍

modules/service/src/main/scala/lucuma/odb/service/ObservationWorkflowService.scala
```diff
@@ -332,6 +367,7 @@ object ObservationWorkflowService {
               .flatMap(addCfpInfos)
               .flatMap(addProgramAllocations)
               .flatMap(addCoordinates)
+              .flatMap(addOtherConfigErrors)
```
cquiroz, line 370: Are these going to be exposed as errors to explore? do they block workflow state changes?

cquiroz (approved):
Looks great, some comments inline

modules/binding/src/main/scala/lucuma/odb/data/PosAngleConstraintMode.scala
swalker2m: Yes, and we use it.  We store this as an enum (discriminator) + an angle in the database though and we need the enum somewhere.  I think there used to be one in core but I couldn't find it.  Maybe that's for the best.

modules/sequence/src/main/scala/lucuma/odb/sequence/ghost/ifu/GhostIfuMappingSyntax.scala
swalker2m: Right that'd be fine if we find we need it.

modules/sequence/src/main/scala/lucuma/odb/sequence/ghost/ifu/IfuMappingContext.scala
swalker2m: To derive a mapping and check validity I need an angle.  We will probably have to support a mutation that selects a sky position and automatically adjusts the PA.  The guide star will be a PWFS option so reachability shouldn't be a concern, but I suppose the PA impacts vignetting.  I'm not sure how to trade these off.

modules/service/src/main/scala/lucuma/odb/service/ObservationWorkflowService.scala
swalker2m: Yes, if there isn't an IFU mapping then the observation isn't considered defined.  It should appear as an error just as if there were some other missing bit of configuration.

PR #2524, opened 2026-05-29
Gnirs acquisiton model adjustments

rpiaggio:
- Acquisition filter follows default/explicit pattern like the Flamingos2 one.
- Use `_effective` `COALESCE` view column for all default/explicit columns.
- Infer default acquisition filter as the closest to the science wavelength from a select list, again mimicking Flamingos2 logic.


rpiaggio:
Companion explore PR: https://github.com/gemini-hlsw/lucuma-apps/pull/884

modules/service/src/main/resources/db/migration/V1159__gnirs_acq_filter.sql
```diff
@@ -0,0 +1,102 @@
...
+
+-- The acquisition filter is now an optional explicit override (NULL = use default).
+ALTER TABLE t_gnirs_long_slit ALTER COLUMN c_acq_filter DROP NOT NULL;
+
+CREATE VIEW v_gnirs_long_slit AS
+  SELECT
+    ls.*,
+    -- grating wavelength default + effective
+    d.c_grating_wavelength_default,
+    COALESCE(ls.c_grating_wavelength, d.c_grating_wavelength_default) AS c_grating_wavelength_effective,
```
swalker2m, line 29: I like this `_effective` field idea for the view.  Wish I'd done it elsewhere.

swalker2m (approved):
Looks great, no notes. :+1:

modules/service/src/main/resources/db/migration/V1159__gnirs_acq_filter.sql
rpiaggio: Credit where it's due: it was Claude's idea

PR #2525, opened 2026-05-29
Allow determining if sequences have been materialized and deleting them

toddburnside:
I think this is ready for review now.

Final state:
- AccessControl prevents `deleteSequence` for executed observations.
- An additional check prevents `deleteSequence` if the observation has any visits.

Because of those 2 limitations, we don't have to worry about whether we should delete the static config or not, because the static config is created on the first visit.


modules/service/src/main/scala/lucuma/odb/graphql/mapping/AccessControl.scala
```diff
@@ -757,4 +758,27 @@ trait AccessControl[F[_]] extends Predicates[F] {
                   case o       => Result.internalError(s"Checked one id '$o', but got a list of ids back: $os")
       yield c).value
 
+  def selectForUpdate(
+    input: DeleteSequenceInput
+  )(using Services[F], NoTransaction[F]): F[Result[AccessControl.CheckedWithId[Unit, Observation.Id]]] =
+    val allowedStates: Set[ObservationWorkflowState] =
+      if user.role.access <= Access.Pi then ObservationWorkflowState.preExecutionSet // ok prior to execution
+      else ObservationWorkflowState.allButComplete
```
toddburnside: I wasn't sure about the access limitations, but it seemed reasonable to not allow deleting a sequence for completed observations.

swalker2m (approved):
This looks great to me. :+1:

PR #2530, opened 2026-06-01
Shortcut 8595: proximity error

swalker2m:
Adds a proximity test to the GHOST IFU mapping such that target + sky or target + target configurations in which the two IFU probes are required to be too close are flagged.

### DIFF
--- a/modules/sequence/src/main/scala/lucuma/odb/sequence/ghost/ifu/GhostIfuMappingSyntax.scala
+++ b/modules/sequence/src/main/scala/lucuma/odb/sequence/ghost/ifu/GhostIfuMappingSyntax.scala
@@ -25,12 +25,20 @@ import lucuma.core.util.Timestamp
  */
 object GhostIfuMappingSyntax:
 
+  private enum AssignmentResult:
+    case TargetAtIfu1 extends AssignmentResult
+    case TargetAtIfu2 extends AssignmentResult
+    case OutOfRange   extends AssignmentResult
+    case TooClose     extends AssignmentResult
+
+  import AssignmentResult.*
+
   private def ifuAssignment(
     base:          Coordinates,
-    c1:            Coordinates,
-    c2:            Option[Coordinates],
+    target:        Coordinates,
+    skyOrTarget:   Option[Coordinates],
     positionAngle: Option[Angle]
-  ): Int =
+  ): AssignmentResult =
     val angle  = positionAngle.getOrElse(Angle.Angle0)
 
     val field1 = GhostIfuPatrolField.ifu1PatrolFieldAt(angle, Offset.Zero)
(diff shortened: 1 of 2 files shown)


modules/sequence/src/main/scala/lucuma/odb/sequence/ghost/ifu/GhostIfuMappingSyntax.scala
```diff
@@ -25,12 +25,20 @@ import lucuma.core.util.Timestamp
  */
 object GhostIfuMappingSyntax:
 
+  private enum AssignmentResult:
+    case TargetAtIfu1 extends AssignmentResult
```
cquiroz, line 29: I was going to suggest on the last PR to use an enum rather than numbers.
Should we use these on the frontend for presenting errors?

modules/sequence/src/main/scala/lucuma/odb/sequence/ghost/ifu/GhostIfuMappingSyntax.scala
```diff
@@ -39,12 +47,30 @@ object GhostIfuMappingSyntax:
...
+      if shape1.contains(pos1) && pos2.forall(shape2.contains)      then TargetAtIfu1
+      else if pos2.forall(shape1.contains) && shape2.contains(pos1) then TargetAtIfu2
+      else OutOfRange
+
+    extension (a: Angle)
+      def µas: Long =
+        Angle.signedMicroarcseconds.get(a).abs
+
+    // Check whether the positions are too close.
+    pos2.fold(preliminaryResult): p =>
```
cquiroz, line 63: are these measured from the center of each ifu?

cquiroz (approved):
LGTM

modules/sequence/src/main/scala/lucuma/odb/sequence/ghost/ifu/GhostIfuMappingSyntax.scala
swalker2m: Would it be useful?  I just made it a private enum to communicate which message to send.  I'm not sure how we'd communicate the enum back, unless we expose this information in the API itself I guess.

modules/sequence/src/main/scala/lucuma/odb/sequence/ghost/ifu/GhostIfuMappingSyntax.scala
swalker2m: Just from the two coordinates.  It just checks whether the two coordinates are within 102 arcseconds of one another.  Is this correct?

cquiroz: i was just wondering, i guess it is the same as in the OT?

cquiroz (approved):
LGTM

PR #2531, opened 2026-06-01
Gnirs preliminary acquisition sequence generation

rpiaggio:
- Remove the special cases for no-acquisition Gnirs spectroscopy.
- Implement preliminary acquisition sequence generation for Gnirs. Tried to mimick F2's logic with the new Gnirs requirements. It may need some iteration:
  - The specs show duplicated steps. I don't know if this was done in OCS to have a generated step handy in case it needed repetition or if we actually want the duplicated step in the sequence. I'm assuming the former. Since in GPP we can edit the sequence and clone steps, this wouldn't be necessary.
  - The specs show a different filter in the first step in some cases. I'm not sure what the logic is behind this. For the moment, the same filter is used throughout.

I've already asked these things to Andy. I'd like to merge this PR so it doesn't diverge too much and continue work on the explore logic while we wait for answers, we can fine tune later.

### DIFF
--- a/ITC-README.md
+++ b/ITC-README.md
@@ -1,6 +1,5 @@
 # lucuma-itc
 
-
 This is a graphql server acting as a proxy for the old ocs2-based itc server
 
 # Run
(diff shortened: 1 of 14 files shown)


swalker2m (approved):
Seems plausible to me.  :+1:

PR #2533, opened 2026-06-01
Shortcut 7072: GMOS Acquisition Read Mode

swalker2m:
Uses `LOW` read mode for long acquisition exposure times.

### DIFF
--- a/modules/sequence/src/main/scala/lucuma/odb/sequence/gmos/longslit/Acquisition.scala
+++ b/modules/sequence/src/main/scala/lucuma/odb/sequence/gmos/longslit/Acquisition.scala
@@ -109,10 +109,12 @@ object Acquisition:
           _  <- optics.fpu           := GmosFpuMask.Builtin(fpu).some
           _  <- optics.xBin          := GmosXBinning.One
           _  <- optics.yBin          := GmosYBinning.One
+          _  <- optics.ampReadMode   := GmosAmpReadMode.Fast
           _  <- optics.roi           := acqConfig.roi.slitRoi
           s1 <- scienceStep(10.arcsec, 0.arcsec, ObserveClass.Acquisition)
 
           _  <- optics.exposure      := lastExpTime(exposureTime)
+          _  <- optics.ampReadMode   := readMode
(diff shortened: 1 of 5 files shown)


modules/sequence/src/main/scala/lucuma/odb/sequence/gmos/longslit/Acquisition.scala
```diff
@@ -97,6 +101,7 @@ object Acquisition:
           _  <- optics.grating       := none[(G, GmosGratingOrder, Wavelength)]
           _  <- optics.xBin          := GmosXBinning.Two
           _  <- optics.yBin          := GmosYBinning.Two
+          _  <- optics.ampReadMode   := readMode
```
cquiroz, line 104: Will this apply to all steps in the acq?

swalker2m: Yes, any settings that aren't explicitly changed apply to the remaining steps.  One question is whether we're supposed to switch back to `FAST` for the 20 second step 2.  I'll assume not and correct if necessary.

modules/service/src/test/scala/lucuma/odb/graphql/query/executionAcqGmosNorth_slow.scala
```diff
@@ -0,0 +1,102 @@
...
+                          "exposure": { "seconds": 61.000000 },
+                          "readout": {
+                            "ampReadMode": "SLOW",
+                            "ampGain": "LOW"
+                          }
+                        }
+                      },
+                      {
+                        "instrumentConfig": {
+                          "exposure": { "seconds": 20.000000 },
```
cquiroz, line 79: Like this step should also be slow/low if exp time is under 60?

cquiroz (approved):
LGTM

modules/sequence/src/main/scala/lucuma/odb/sequence/gmos/longslit/Acquisition.scala
swalker2m: I got an answer on this.  We do need to switch back to `FAST` for the 20 second step.

PR #2534, opened 2026-06-01
Add Flamingos2 imaging observing mode enum and stubs

cquiroz:
This a bare minimum to unblock after the lucuma-core release


rpiaggio (approved):
LGTM, thank you!

PR #2537, opened 2026-06-02
Add support to create and read f2 imaging

cquiroz:
This is more or less a one-to-one copy of gmos imaging.


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -9051,6 +9020,155 @@ input Flamingos2LongSlitInput {
...
+
+  """
+  Optional explicitly specified F2 ReadMode. If set it overrides the default.
+  """
+  explicitReadMode: Flamingos2ReadMode
+
+  """
+  Optional explicitly specified F2 Reads. If set it overrides the default.
+  """
+  explicitReads: Flamingos2Reads
```
swalker2m, line 9128: Should there be `defaultReads` and `reads` fields to accompany this?  Same for `defaultReadMode` and `readMode`?

cquiroz: right how did I miss them

modules/service/src/main/scala/lucuma/odb/graphql/input/ImagingFilterCheck.scala
```diff
@@ -0,0 +1,22 @@
...
+
+package lucuma.odb.graphql.input
+
+import cats.data.NonEmptyList
+import cats.syntax.all.*
+import grackle.Result
+import lucuma.odb.data.OdbError
+import lucuma.odb.data.OdbErrorExtensions.*
+
+trait ImagingFilterCheck:
```
swalker2m, line 12: Seems like a good idea to extract this :+1:

swalker2m (approved):
Looks great. :+1:

PR #2538, opened 2026-06-03
Shortcut 8836: Zero minimum time estimate bug

swalker2m:
The time estimate service builds a tree of groups and their observations which it traverses to calculate the minimum and maximum planned time.   Unfortunately it includes deleted groups.

In the issue report, a group (with minimumRequired = 1) contained an observation and three sub-groups, one of which was deleted (and empty).  The empty group had a minimum time estimate of 0.  Since only one is required and it is empty, the minimum time estimate is 0.

The solution is to filter out deleted groups, since they shouldn't be included in the estimate.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/GroupService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/GroupService.scala
@@ -42,7 +42,7 @@ trait GroupService[F[_]] {
   def updateGroups(SET: GroupPropertiesInput.Edit, which: AppliedFragment)(using Transaction[F]): F[Result[List[Group.Id]]]
   def selectGroups(
     programId: Program.Id,
-    groupFilter: AppliedFragment = void"",
+    groupFilter: AppliedFragment = sql"c_existence = $existence"(Existence.Present),
     obsFilter: AppliedFragment = sql"c_existence = $existence"(Existence.Present)
   )(using Transaction[F]): F[GroupTree]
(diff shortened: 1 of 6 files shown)


modules/service/src/main/scala/lucuma/odb/service/GroupService.scala
```diff
@@ -42,7 +42,7 @@ trait GroupService[F[_]] {
   def updateGroups(SET: GroupPropertiesInput.Edit, which: AppliedFragment)(using Transaction[F]): F[Result[List[Group.Id]]]
   def selectGroups(
     programId: Program.Id,
-    groupFilter: AppliedFragment = void"",
+    groupFilter: AppliedFragment = sql"c_existence = $existence"(Existence.Present),
```
cquiroz, line 45: What happens if the caller passes another filter? should the existence filter be appended to the passed param?

swalker2m: I take it that if you supply your own filter it would decide whether to include deleted elements, just as was done for the observation filter in the next line down.

cquiroz (approved):
LGTM

modules/service/src/main/scala/lucuma/odb/service/GroupService.scala
swalker2m: I checked the actual usages outside of tests and (a) we never pass any filters and (b) all are in contexts in which we should have been excluding deleted groups.

PR #2540, opened 2026-06-03
Gnirs acquisition fine tuning: sky offsets are typed to acquisition type faint

rpiaggio:
Also, first step when VeryBright should always use H filter.

This PR requires https://github.com/gemini-hlsw/lucuma-core/pull/1432, it will fail to compile until that core PR is released. But it can still be reviewed.


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -9726,20 +9726,16 @@ input Igrins2LongSlitInput {
...
-  filter: GnirsFilter!
+  coadds: PosInt!
 
-  "The acquisition filter used by default (closest acquisition filter to the science wavelength) when no explicit filter is specified."
-  defaultFilter: GnirsFilter!
+  explicitAcquisitionType: GnirsAcquisitionType
 
-  "An explicitly specified acquisition filter to use instead of the default (if any)."
+  "An explicitly specified acquisition filter. When null, the filter is determined automatically from the acquisition mode at sequence-generation time."
   explicitFilter: GnirsFilter
```
swalker2m, line 9736: `filter` and `defaultFilter` are removed, but we keep `explicitFilter`?   Also, shouldn't there be an `acquisitionType` and `defaultAcquisitionType`?

rpiaggio: The thing is that the default acquisition type depends on the exposure time, which we don't have until we call ITC. And the default acquisition filter depends on the acquisition type, so it also depends transitively on the exposure time.

So I decided to implement with the same strategy as read mode (also used for F2) where these values can be `null`, and in that case they are computed automatically during sequence generation after invoking ITC. The UI will show an `Auto` option in the respective dropdowns. That's why there are no defaults, the default is simply `null`. 

The alternative to this is to invoke ITC when creating the observing mode and on every relevant update.

swalker2m (approved):
I'm not really following this one, but it seems plausible.

PR #2541, opened 2026-06-03
Shortcut 8595: Switch to target ids in GHOST IFU mapping

swalker2m:
These are the database changes that would accompany [the core update](https://github.com/gemini-hlsw/lucuma-core/pull/1434).  Needs that PR merged and a core release in order to compile and work.


modules/service/src/main/resources/db/migration/V1166__ghost.sql
```diff
@@ -0,0 +1,111 @@
+-- Migration: Replace IFU coordinate/proper-motion columns with target_id
+-- foreign keys; remove 'nonsidereal' from e_ghost_ifu_mapping.
+--
+-- Precondition: t_ghost_static has no rows (development environment).
```
swalker2m, line 4: Migration by Claude (mostly).   This only works if there are no existing `t_ghost_static` rows, which is the case as of the time of this comment.

modules/service/src/main/resources/db/migration/V1165__ghost.sql
```diff
@@ -0,0 +1,111 @@
...
+    FOREIGN KEY (c_program_id, c_observation_id, c_ifu2_target_id)
+    REFERENCES t_asterism_target (c_program_id, c_observation_id, c_target_id);
+
+-- -----------------------------------------------------------------------
+-- 4. Remove 'nonsidereal' from e_ghost_ifu_mapping.
+--    PostgreSQL has no DROP VALUE so we rename, recreate, cast, clean up.
+-- -----------------------------------------------------------------------
+ALTER TYPE e_ghost_ifu_mapping RENAME TO e_ghost_ifu_mapping_old;
+
+CREATE TYPE e_ghost_ifu_mapping AS ENUM (
```
cquiroz, line 60: I recall you had to create the enums in one trasaction and use them on the next. But this seems to be passing tests

cquiroz (approved):
LGTM

modules/service/src/main/resources/db/migration/V1165__ghost.sql
swalker2m: Right, I'm not sure what the deal is.  I tried the migration myself of dev database data and there were no complaints. 🤔

PR #2546, opened 2026-06-04
Implement resource endpoints connected to database

hugo-vrijswijk:


resource/service/src/main/resources/db/migration/V0000__telescope_night_timeline.sql
```diff
@@ -0,0 +1,96 @@
+-- SITES
+
+create type e_site as enum('gn', 'gs');
```
cquiroz, line 3: Is this a separate db right? We have some of these enums on the common db

hugo-vrijswijk: At this point in time it is a separate db, yes

resource/service/src/main/scala/resource/server/graphql/QueryMapping.scala
```diff
@@ -3,56 +3,57 @@
...
 import lucuma.core.enums.Site
 import lucuma.odb.graphql.binding.*
-import resource.model.*
-import resource.server.DummyData
+import resource.server.graphql.table.*
 
 import java.time.LocalDate
 
-trait QueryMapping[F[_]: Sync] extends BaseMapping[F] {
+given Eq[LocalDate] = Eq.fromUniversalEquals
```
cquiroz: We may have this instance on cats time

cquiroz (commented):
In general it seems fine I wonder if there is some code that could be shared across servers. Do you think it would help?

resource/service/src/main/scala/resource/server/graphql/QueryMapping.scala
hugo-vrijswijk: I thought it must be somewhere. Thanks

hugo-vrijswijk:
> In general it seems fine I wonder if there is some code that could be shared across servers. Do you think it would help?

Yeah probably. I didn't want to depend directly on `service`. But I might end up moving some code to a separate module

PR #2547, opened 2026-06-04
Igrins2 reacquistion time should be 0

cquiroz:


modules/service/src/main/resources/db/migration/V1167__igrins2_reacquisition_zero.sql
```diff
@@ -0,0 +1,21 @@
+-- IGRINS-2 has on-instrument guiding and flexure compensation, so
+-- re-acquisitions are not required.
+-- Set the reacquisition overhead to 0.
+UPDATE t_time_estimate
+   SET c_time = '0 seconds'
+ WHERE c_tag = 'igrins2_reacquisition';
+
+-- Recalculate the time estimates for all observations
+TRUNCATE TABLE t_obscalc;
```
cquiroz: @swalker2m this maye be heavy handed but I see we have used this style for other similar adjustments.
Should I try to only delete igrins2 obscalc entries instead?

swalker2m: I think we could probably do something like:

```sql
UPDATE t_obscalc c
SET c_obscalc_state = 'pending'
FROM t_observation o
WHERE c.c_observation_id = o.c_observation_id
 AND o.c_instrument = 'igrins2'; 
```

cquiroz: done

swalker2m (approved):
Looks great. :+1:

PR #2548, opened 2026-06-04
Shortcut 7179: `isSplittable`

swalker2m:
This is part of [Shortcut 7179](https://app.shortcut.com/lucuma/story/7179/support-unsplittable-observations).  This PR only:

1. Adds a `scheduling: SchedulingPropertiesInput` to hold both the new `isSplittable` property and the existing `timingWindows`.  This corresponds to the UI suggestion grouping these items.
2. Deprecates (but does not remove) the existing `timingWindows` field in `ObservationPropertiesInput`, to ease the transition. If both `scheduling` and `timingWindows` are set, then the input is rejected with an error.

**Note: at the risk of attracting "it doesn't work" Shortcut tasks, the `isSplittable` field is not yet being used.** In the next installment I'll need to add the value to the observation table and make a scheduling properties field for `Observation` and the associated Grackle mapping.  Finally the sequence generation will need to respect the splittable field and manual edits that try to split an unsplittable observation will have to be rejected.

I'm not crazy about the term "splittable" but it was suggested even for the UI so it seemed best to go with it.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/graphql/input/ObservationPropertiesInput.scala
+++ b/modules/service/src/main/scala/lucuma/odb/graphql/input/ObservationPropertiesInput.scala
@@ -9,17 +9,24 @@ import cats.data.NonEmptyList
 import cats.syntax.all.*
 import eu.timepit.refined.types.numeric.NonNegShort
 import eu.timepit.refined.types.string.NonEmptyString
+import grackle.Result
+import grackle.syntax.*
 import lucuma.core.enums.ScienceBand
 import lucuma.core.model.Attachment
 import lucuma.core.model.Group
 import lucuma.core.model.Target
 import lucuma.odb.data.Existence
 import lucuma.odb.data.Nullable
+import lucuma.odb.data.OdbError
+import lucuma.odb.data.OdbErrorExtensions.*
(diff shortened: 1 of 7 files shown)


modules/service/src/main/scala/lucuma/odb/graphql/input/ObservationPropertiesInput.scala
```diff
@@ -95,7 +103,11 @@ object ObservationPropertiesInput {
             rPosAngleConstraint,
             rTargetEnvironment,
             rConstraintSet,
-            rTimingWindows,
+            (rTimingWindows, rScheduling).parFlatMapN {
+              case (Some(_), Some(_)) => PleaseUseSchedulingField
```
cquiroz, line 107: I guess this may soft-break clients which try to use `timingWindows`

cquiroz (approved):
LGTM

modules/service/src/main/scala/lucuma/odb/graphql/input/ObservationPropertiesInput.scala
swalker2m: It would only be a problem if they use both `timingWindows` _and_ `scheduling`.

PR #2552, opened 2026-06-05
Rename scienceFov to agsDiameter

cquiroz:
MaroonX will use the sky fiber diameter of 30 arcsecs rather than the fiber fov

Now looking at the schema changes, should we keep both fields? it would only make a difference for MaroonX @swalker2m wdyt?

Don't merge until explore is updated


swalker2m (approved):
I prefer this name and approve this change because

* it identifies or hints at the purpose of this information
* does not mislead into thinking that it limits or represents the exact field of view for science

:+1:

PR #2556, opened 2026-06-05
Shortcut 7179: stores the splittable field

swalker2m:
Adds storage, mutation, clone to the scheduling constraints (which I renamed from scheduling "properties").  They are still not actually _used_, but [the UI](https://app.shortcut.com/lucuma/story/6491/support-unsplittable-observations) could be made at this point (Shortcut 6491 should be unblocked).


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -12751,9 +12775,15 @@ type Observation {
   constraintSet: ConstraintSet!
 
   """
-  Observation timing windows
+  Deprecated (using `schedulingConstraints` instead).  Observation timing
+  windows.
   """
-  timingWindows: [TimingWindow!]!
+  timingWindows: [TimingWindow!]! @deprecated(reason: "Use `schedulingConstraints` instead. (Deprecated June 5, 2026)")
```
cquiroz, line 12781: explore can't handle deprecated fields though I think it won't break the installed version but will not let you send new PRs until fixed

swalker2m: This is a good thing, right?  At any rate, it is just moving the field down inside of `schedulingConstraints`.  Everything else about timing windows works the same.

toddburnside: With the rate PRs and releases have been happening lately, maybe I should get a PR in for Explore with the minimum changes required to support this before we merge this PR? I'll try to get that done today.

toddburnside (approved):
LGTM

I made the explore PR to make use of the new API: https://github.com/gemini-hlsw/lucuma-apps/pull/923

Unless someone wants changes, feel free to merge it along with this one.

PR #2560, opened 2026-06-07
F2 should use offsets generators

cquiroz:
By mistake i added regular offsets for F2 imaging instead of what GMOS imaging does using offset generators.

This PR corrects this and also generalizes the variants to be used by both F2 and GMOS

Don't merge until the explore PR is ready


modules/service/src/main/scala/lucuma/odb/graphql/mapping/Flamingos2ImagingMapping.scala
```diff
@@ -46,10 +35,47 @@ trait Flamingos2ImagingMapping[F[_]]
...
     )
 
+  lazy val Flamingos2GroupedImagingMapping: ObjectMapping =
+    ObjectMapping(Flamingos2ImagingType / "variant" / "grouped")(
+      SqlField("observationId", Flamingos2ImagingView.Grouped.ObservationId, key = true, hidden = true),
+      SqlField("order",         Flamingos2ImagingView.Grouped.WavelengthOrder),
+      SqlField("skyCount",      Flamingos2ImagingView.Sky.Count),
+      SqlObject("offsets",      Join(Flamingos2ImagingView.ObservationId, TelescopeConfigGeneratorView.ObjectObservationId)),
+      SqlObject("skyOffsets",   Join(Flamingos2ImagingView.ObservationId, TelescopeConfigGeneratorView.SkyObservationId)),
+    )
```
swalker2m, line 45: We might be able to generalize these mappings too at some point.

swalker2m (approved):
This all looks reasonable to me. :+1:

PR #2561, opened 2026-06-07
Enable S/N ETM on Gnirs sequence generation call to ITC

rpiaggio:


modules/service/src/main/scala/lucuma/odb/service/GeneratorParamsService.scala
```diff
@@ -426,6 +426,12 @@ object GeneratorParamsService {
                              .fold(GnirsFilter.fromSpectroscopyWavelength(gn.gratingWavelength))(_.asRight[String])
                              .leftMap(msg => Error.MisconfiguredObservation(obsParams.observationId, msg))
             yield
+              val sciReadMode = gn.exposureTimeMode match
+                                  case ExposureTimeMode.SignalToNoiseMode(_, _) =>
+                                    GnirsReadMode.Bright // In practice this will be ignored by the ITC, which derives the read mode itself in S/N mode
```
cquiroz, line 431: I think we do the same thing for f2 read mode right?

cquiroz (approved):
LGTM

modules/service/src/main/scala/lucuma/odb/service/GeneratorParamsService.scala
rpiaggio: Exactly

PR #2566, opened 2026-06-08
Gnirs tellurics

rpiaggio:
- Add telluricType like for F2 and Igrins2.
- Use different offsets depending on configuration according to the spec.


swalker2m (approved):
Looks reasonable. :+1:

modules/service/src/main/scala/lucuma/odb/service/TelluricTargetsService.scala
```diff
@@ -576,3 +588,11 @@ object TelluricTargetsService:
         WHERE  c_instrument = 'Igrins2'
         LIMIT  1
       """.query(numeric.opt *: numeric.opt)
+
+    val SelectGnirsHmin: Query[Void, (Option[BigDecimal], Option[BigDecimal])] =
+      sql"""
+        SELECT c_hmin_hot, c_hmin_solar
+        FROM   t_spectroscopy_config_option
```
cquiroz, line 595: Was this value included on gnirs phase0 data?

rpiaggio: Good catch, they are not loaded in the current spreadsheet. I'll track the issue.

PR #2573, opened 2026-06-09
Shortcut 7179: un-splittable observations (part 3)

swalker2m:
When an observation is marked un-splittable, the generated science sequence should contain a single atom containing all the steps of whatever atoms it would have otherwise contained.  This is accomplished by doing a normal sequence generation and then combining all the atoms' steps into a single atom.

Note, for the next PR(s), there is more work to do for manual sequences and for the setup count estimation.

### DIFF
--- a/modules/sequence/src/main/scala/lucuma/odb/sequence/data/UnsplittableAtom.scala
+++ b/modules/sequence/src/main/scala/lucuma/odb/sequence/data/UnsplittableAtom.scala
@@ -12,7 +12,7 @@ object UnsplittableAtom:
    * Default title for the single atom that appears in the science sequence of
    * an unsplittable observation.
    */
-  val Descripton: NonEmptyString =
+  val Description: NonEmptyString =
     NonEmptyString.unsafeFrom("Unsplittable Science Atom")
 
   /**
--- a/modules/sequence/src/main/scala/lucuma/odb/sequence/data/UnsplittableAtom.scala
+++ b/modules/sequence/src/main/scala/lucuma/odb/sequence/data/UnsplittableAtom.scala
@@ -1,0 +1,23 @@
+// Copyright (c) 2016-2025 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.odb.sequence.data
+
+import eu.timepit.refined.types.numeric.PosInt
+import eu.timepit.refined.types.string.NonEmptyString
+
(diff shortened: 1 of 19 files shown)


modules/sequence/src/main/scala/lucuma/odb/sequence/data/UnsplittableAtom.scala
```diff
@@ -0,0 +1,23 @@
...
+   */
+  val Descripton: NonEmptyString =
+    NonEmptyString.unsafeFrom("Unsplittable Science Atom")
+
+  /**
+   * Size limit for the number of steps in the single atom of an unsplittable
+   * observation.
+   */
+  val StepLimit: PosInt =
+    PosInt.unsafeFrom(2048)
```
swalker2m, line 23: This is arbitrary, but we can change it if requested.  I think there needs to be some limit though.

modules/service/src/main/scala/lucuma/odb/logic/GeneratorStreaming.scala
```diff
@@ -211,7 +222,8 @@ object GeneratorStreaming:
           itc  = requireSpectroscopyItc(context.oid, context.itcRes)
           rol  = context.params.calibrationRole
           gen <- EitherT(LongSlit.instantiate(context.oid, calculator.flamingos2Step, context.namespace, exp.flamingos2, cfg, itc, rol))
-        yield gen.covary[F]).value
+          res <- collapseIfNecessary(context, gen)
+        yield res).value
```
swalker2m, line 226: It's a bit unfortunate that you have to remember to do this, but I didn't come up with a better way.  There isn't a single point that all these generator calls get funneled into.

modules/sequence/src/main/scala/lucuma/odb/sequence/data/UnsplittableAtom.scala
cquiroz: don't we have a 1k step limit somewhere?

cquiroz (approved):
LGTM

modules/sequence/src/main/scala/lucuma/odb/sequence/data/UnsplittableAtom.scala
swalker2m: It's an atom limit.  In this case, we'll have just a single atom so it seems like we should limit the steps.  Really we should limit the steps regardless but it seems particularly important for this.

Typically an atom contains one or more science observations and some cals so 1K seemed like too few.

PR #2574, opened 2026-06-09
Add support to edit f2 imaging observations

cquiroz:
Again very much based on gmos imaging. I tried to export some of the common parts to a new `ImagingService` that can be used by both f2 and gmos


swalker2m (approved):
Thanks for generalizing all that.  Variant updates are tricky.  :+1:

PR #2577, opened 2026-06-10
Add cassRotator field to the target enviroment

cquiroz:
This is used by navigate to select the kind of cass rotation mode. I didn't add a field to the db as the rule for now is that maroonx uses `fixed` and the rest are `following`. 

Eventually this field would be user selectable when we add `Altair` support

### DIFF
--- a/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -6864,6 +6864,21 @@ type CatalogInfo {
 }
 
 """
+Cassegrain rotator tracking mode
+"""
+enum CassRotator {
+  """
+  The cass rotator is held at a fixed angle.
+  """
+  FIXED
+
+  """
+  The cass rotator follows the field.
+  """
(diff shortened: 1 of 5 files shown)


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -16243,6 +16258,11 @@ type TargetEnvironment {
   The type of blind offset (automatic or manual) if a blind offset exists.
   """
   blindOffsetType: BlindOffsetType!
+
+  """
+  The cassegrain rotator tracking mode.
+  """
+  cassRotator: CassRotator!
```
cquiroz, line 16265: @swalker2m does it seem a good location for this field? This is what we agreed with @jluhrs but could go some other place instead.

swalker2m: Seems reasonable, I think.

swalker2m (approved):
Looks good. :+1:

jluhrs (approved):
LGTM

PR #2578, opened 2026-06-10
Gnirs inline callibrations with smart gcal port

rpiaggio:
Calibrations follow ~the 90-minute~ a 3-hour min interval and 3 hours max visits, ~same as~ similar to for F2.


swalker2m (approved):
Looks great.  :+1:

PR #2579, opened 2026-06-10
Shortcut 7179: Unsplittable observation manual sequence edit

swalker2m:
Updates the handling of unsplittable observations such that:

* Attempting to replace the sequence (i.e., manual sequence edit) of an unsplittable observation is rejected if it contains more than one atom or if that atom has too many steps
* Marking a heretofore splittable observation as unsplittable is not allowed if it has a materialized sequence with multiple atoms, or with a single atom containing too many steps


cquiroz (approved):
LGTM

cquiroz (approved):
LGTM

PR #2581, opened 2026-06-10
Shortcut 7179: Estimate setup cost for unsplittable observations

swalker2m:
Unsplittable observations should have a single setup, which this PR guarantees.

There are a couple of valid criticisms that occur to me:

* We are not validating, at least not yet, that the single atom is short enough to run in a single night.
* The splittable observation setup count doesn't look at the number of atoms, though it probably should.  For example, if there is only one atom we can't have multiple setups even if it is super long.

We should look at the time estimates for the sequence atoms to come up with this estimate rather than checking the splittable property or the total time estimate.  Perhaps someday we'll address those concerns.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/logic/Generator.scala
+++ b/modules/service/src/main/scala/lucuma/odb/logic/Generator.scala
@@ -204,7 +204,9 @@ object Generator:
           for
             a <- EitherT(sequenceDigest(stream.acquisition))
             s <- EitherT(sequenceDigest(stream.science))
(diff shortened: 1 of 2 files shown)


cquiroz:
>     * We are not validating, at least not yet, that the single atom is short enough to run in a single night.
Should the scheduler take care of that?

swalker2m:
> > ```
> > * We are not validating, at least not yet, that the single atom is short enough to run in a single night.
> > ```
> 
> Should the scheduler take care of that?

I suppose the scheduler will ensure that such a long atom is never scheduled.  Ideally we'd generate some kind of error or warning as well that shows up in Explore.

cquiroz (approved):
LGTM

PR #2582, opened 2026-06-11
Add support to clone f2 imaging observations

cquiroz:
As before, some of the code was extracted to be shared between F2 and GMOS

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,7 @@ jobs:
   build:
     name: Test
     strategy:
+      fail-fast: false
       matrix:
         os: [ubuntu-22.04]
         scala: [3]
--- a/modules/service/src/main/scala/lucuma/odb/service/Flamingos2ImagingService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/Flamingos2ImagingService.scala
@@ -189,8 +189,9 @@ object Flamingos2ImagingService:
         newObservationId: Observation.Id,
         etms:             List[(ExposureTimeModeId, ExposureTimeModeId)]
       )(using Transaction[F]): F[Unit] =
-        session.exec(Statements.clone(observationId, newObservationId))                       *>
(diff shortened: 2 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -27,6 +27,7 @@ jobs:
   build:
     name: Test
     strategy:
+      fail-fast: false
```
swalker2m, line 30: This will give us all the test results instead of having to push test fixes one by one?  The cost is continuing to use CI resources even after we know the build will fail?  Is this the tradeoff or what is the thinking here?

modules/service/src/main/scala/lucuma/odb/service/Flamingos2ImagingService.scala
```diff
@@ -260,3 +276,49 @@ object Flamingos2ImagingService:
...
+    ): AppliedFragment =
+      sql"""
+        INSERT INTO #${Flamingos2ImagingService.ModeTableName} (
+          c_observation_id,
+          c_program_id,
+          c_read_mode,
+          c_reads,
+          c_decker,
+          c_readout_mode,
+          c_variant,
```
swalker2m: This seems good.  If the variant column names are shared, we might consider defining them in `ImagingStatements` and then just unpacking them here and in `GmosImagingService`?

swalker2m (approved):
Looks great. :+1:

.github/workflows/ci.yml
cquiroz: It is mostly about this docker failures we've seen. I `think` it would let all the shards try to run and could save us some time as we can restart the single ones that failed
We can test if that's helping a bit I'm not sure

modules/service/src/main/scala/lucuma/odb/service/Flamingos2ImagingService.scala
cquiroz: Right, we could do that too

PR #2590, opened 2026-06-12
Require central wavelength for Gnirs, rename

rpiaggio:
The central wavelength for GNIRS spectroscopy has to be passed when creating the observing mode, rather than inferring it from the filter. It therefore switches to a value/`initial`/`effective` pattern rather than the `explicit`/`default` one.

Also, it's renamed from grating wavelength => central wavelength. Although the former is more specific, the latter ensures consistent naming across instruments.

This is a breaking change, but I'm sending the explore PR right after this one.

Note that in the dynamic config the property is still called grating wavelength since it's only used for spectroscopy.


toddburnside (approved):
Seems fine to me.

cquiroz (approved):
LGTM

PR #2591, opened 2026-06-12
Add sequence generation for f2 imaging

cquiroz:


cquiroz:
Note that pre-imaging i setup exactly as gmos but there are some question marks on whether that is correct, we may need to tweak pre-imaging in the near future

swalker2m (approved):
Looks great :+1:

PR #2596, opened 2026-06-15
Observation workflow should not allow transition Ongoing -> Inactive

toddburnside:


modules/service/src/main/scala/lucuma/odb/service/ObservationWorkflowService.scala
```diff
@@ -452,12 +452,10 @@ object ObservationWorkflowService {
                 case Defined    => Some(Ready)
 
         // Our final state is the execution state (if any), else the user state (if any), else the validation state,
-        // with the one exception that user state Inactive overrides execution state Ongoing
         val state: ObservationWorkflowState =
           (executionState, userStatus(validationStatus)) match
             case (None, None)     => validationStatus
             case (None, Some(us)) => us
-            case (Some(Ongoing), Some(Inactive)) => Inactive
```
toddburnside, line 460: I could add a migration to set the userStatus column from inactive to null for any ongoing observations, but it won't really make a difference because the inactive flag is ignored for ongoing observations, anyway. Plus, there currently are no such rows in any of the DBs. 😄

swalker2m (approved):
Seems like we explicitly allowed the ongoing => inactive transition on user request.  At any rate, looks fine to me. :+1:

toddburnside:
> Seems like we explicitly allowed the ongoing => inactive transition on user request. At any rate, looks fine to me. 👍

Yeah. The code seems to indicate some special cases being made for that. Bookmark this PR, we may be changing it back some day.

PR #2599, opened 2026-06-16
Omit nighttime cals in GNIRS telluric sequences

rpiaggio:


cquiroz:
Could you explain a bit what's the context of this? does this need to apply for all tellurics?

swalker2m (approved):
Not to sidestep Carlos' question, but the implementation looks good. :+1:

rpiaggio:
> Could you explain a bit what's the context of this? does this need to apply for all tellurics?

It was requested in https://app.shortcut.com/lucuma/story/8995/do-not-include-flats-arcs-in-gnirs-telluric-sequences.

I called Andy's attention to the fact that F2 tellurics do include flats+arcs. He let me know that both instruments behave different in this way, and that it's correct tha F2 includes them and GNIRS doesn't.

PR #2600, opened 2026-06-17
Fix Gnirs auto-acquisition filter logic

rpiaggio:
Follow-up PR to https://github.com/gemini-hlsw/lucuma-core/pull/1459. Will compile once core gets released and the version fixed.

Bonus: fix max science time span between calibrations.


cquiroz (approved):
LGTM

PR #2602, opened 2026-06-17
Add changePrincipalInvestigator mutation

toddburnside:
Adds a GraphQL mutation that transfers the PI role of a program to one of its coinvestigators. The target program user, which must currently be a coinvestigator, becomes the new PI and the current PI is demoted to coinvestigator. Because at most one PI per program is allowed, the demotion is performed before the promotion.

Only the program's current PI or a user with staff access (or greater) is authorized to perform this operation; the access check is baked into the demotion SQL so an unauthorized caller updates no rows.


cquiroz (approved):
LGTM

PR #2603, opened 2026-06-17
Shortcut 8952: Exchange CfP

swalker2m:
This is a work-in-progress PR for adding exchange partners, calls, and proposals.  In summary, it updates the call for proposal API itself, without yet touching the proposal API, submission rules, etc.  I'm pausing for input on that part.

Except for the GraphQL API itself, it was mostly written by Claude.  API updates include:

* Creates CfP properties for each of Gemini, Keck and Subaru.  The existing `CallForProposals` type is divided into four parts:
  1. `GeminiCallProperties` - specific to CfPs for programs that will execute at Gemini.  This is every proposal and program to date.
     * `type: CallForProposalsType!` - Queue vs Classical vs other exotic Gemini proposal types
     * `coordinateLimits: SiteCoordinateLimits!` - both north and south
     * `instruments: [Instrument!]!`
     * `proprietaryMonths: NonNegInt!`
     * `allowsNonPartnerPi: Boolean!`
     * `nonPartnerDeadline: Timestamp`
     * `exchangePartners: [ExchangePartner!]!`
  2. `KeckCallProperties` - specific to CfPs that will execute at Keck
     * `instruments: [KeckInstrument!]!`
     * `coordinateLimits: CoordinateLimits!` - north only
  3. `SubaruCallProperties` - specific to CfPs that will execute at Subaru
     * `type: SubaruProposalType`
     * `instruments: [SubaruInstrument!]`
     * `coordinateLimits: CoordinateLimitsInput`- north only
  4. `CallForProposals` - Container for the three observatories
     * `id: CallForProposalsId!`
     * `title: NonEmptyString!`
     * `semester: Semester!`
     * `active: DateInterval!`
     * `partners: [CallForProposalsPartner!]!` - these are _Gemini_ partners and potentially a submission deadline.  I'm still sorting out how the time requests are supposed to work
     * `submissionDeadlineDefault: Timestamp`
     * `existence: Existence!`
     * `observatory: Observatory!` - discriminator for the observatory-specific properties, exactly one of which is set
     * `gemini: GeminiCallProperties`
     * `keck: KeckCallProperties`
     * `subaru: SubaruCallProperties`
* Changes `HasPartner` partner link type to `HasGeminiPartner` and adds `HasExchangePartner`.  This enables `ProgramUser`s that are affiliated with Keck and Subaru instead of Gemini partners.


modules/service/src/main/scala/lucuma/odb/graphql/input/WherePartnerLink.scala
```diff
@@ -8,26 +8,23 @@ import cats.syntax.parallel.*
...
 import lucuma.core.enums.PartnerLinkType
 import lucuma.odb.graphql.binding.*
 
-object WherePartnerLink {
+object WherePartnerLink:
 
   def binding(path: Path): Matcher[Predicate] =
     val WherePartnerLinkType = WhereEq.binding[PartnerLinkType](path / "linkType", PartnerLinkTypeBinding)
-    val WherePartner         = WhereOptionEq.binding[Partner](path / "partner", PartnerBinding)
+    val WherePartner         = WhereOptionEq.binding[Partner](path / "geminiPartner", PartnerBinding)
```
toddburnside, line 20: Do we want to rename this to WhereGeminiPartner?

modules/service/src/test/scala/lucuma/odb/graphql/mutation/createCallForProposals.scala
toddburnside, line 1: Are there any tests here for non-gemini CfPs? I may have glossed over them.

modules/service/src/test/scala/lucuma/odb/graphql/mutation/updateCallsForProposals.scala
toddburnside, line 1: Same question as for createCallsForProposals

toddburnside (approved):
It's a lot to take it, but it looks great so far.

modules/service/src/main/scala/lucuma/odb/graphql/input/WherePartnerLink.scala
swalker2m: That's a good point.  I debated renaming the enum itself to `GeminiPartner` to better contrast with the new `ExchangePartner`.  If I did that, then it would naturally be `WhereGeminiPartner`.  Would it be too much of a hardship?  I was hesitant to change even more fundamental stuff.

modules/service/src/test/scala/lucuma/odb/graphql/mutation/createCallForProposals.scala
swalker2m: I'll add a few.  This code is kind of pending the proposal creation tied to non-Gemini CfPs but it should be a least lightly tested for now.

modules/service/src/main/scala/lucuma/odb/graphql/input/WherePartnerLink.scala
toddburnside: Yeah. I see what you mean. Changing the enum to GeminiPartner probably would be too bad on the UI side. It's probably a lot more complicated on the API side with migrations and columns and all. But, I'm also fine with the API facing stuff calling it GeminiPartner and in code we call it Partner and say "of course it's a Gemini Partner".

swalker2m:
Close in favor of updated branch.

PR #2609, opened 2026-06-18
Add a predicate for observation mode type

cquiroz:
I want to use this feature to split explore queries on a per mode basis

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/graphql/input/WhereObservation.scala
+++ b/modules/service/src/main/scala/lucuma/odb/graphql/input/WhereObservation.scala
@@ -10,6 +10,7 @@ import grackle.Path
 import grackle.Predicate
 import grackle.Predicate.*
 import lucuma.core.enums.Instrument
+import lucuma.core.enums.ObservingModeType
 import lucuma.core.enums.ScienceBand
 import lucuma.core.enums.Site
 import lucuma.core.model.Observation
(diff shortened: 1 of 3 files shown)


toddburnside (approved):
LGTM

modules/service/src/main/scala/lucuma/odb/graphql/input/WhereObservation.scala
```diff
@@ -52,12 +53,32 @@ object WhereObservation {
...
+                NIN.map(as => Not(In(modePath, as)))
+              ).flatten)
+
     val SubtitleBinding = WhereOptionString.binding(path / "subtitle")
     val WhereOrderObservationIdBinding = WhereOrder.binding[Observation.Id](path / "id", ObservationIdBinding)
     val WhereReferenceBinding = WhereObservationReference.binding(path / "reference")
     val WhereProgramBinding = WhereProgram.binding(path / "program")
     val ScienceBandBinding = WhereOptionOrder.binding(path / "scienceBand", enumeratedBinding[ScienceBand])
     val InstrumentBinding = WhereOptionEq.binding(path / "instrument", enumeratedBinding[Instrument])
+    val ObservingModeTypeBinding = observingModeTypeBinding(enumeratedBinding[ObservingModeType])
```
swalker2m, line 81: Did `WhereOptionEq.binding(path / "observingMode" / "mode", enumeratedBinding[ObservingModeType])` not work?  I'm just wondering if we really need to define it explicitly in `observingModeTypeBinding` above.

PR #2610, opened 2026-06-18
Gnirs daytime pinhole calibrations

rpiaggio:
Implement Daytime Pinhole calibrations. These are used for cross-dispersed observations.

While working in this feature I realized some adjustments were needed so it ends up being a largish PR. I broke it up in conceptual commits, so I recommend reviewing the commits separately:

1. `migrate to new GnirsFpu enum` (`950c09a`):
   * Replace the GNIRS FPU `Either[GnirsFpuSlit, GnirsFpuOther]` with a proper `GnirsFpu` sum type (`Slit` | `Other`). This makes the code easier to read.
   * Add a proper Postgres enumeration for `GnirsFpuOther` (`e_gnirs_fpu_other postgres`), it was only a generic tag before.
   *  Split smart-gcal `c_fpu` into `c_fpu_slit`/`c_fpu_other`.
  
2. `load gnirs daytime pinhole flats into smart gcal` (`0386b18`):
   * Extend the GNIRS smart-gcal parser to read the full file including the pinhole FPU rows, wich daytime pinhole flat generation needs.
   
3. `gnirs daytime pinhole calibration` (`ef20431`):
      * Daytime Pinhole sequence generation with a single flat with data from Smart GCAL. As a `DayCal`, it's not chargeable.

4.  `generate gnirs pinhole cals, rename obs cal groups` (`dc59732`):
   * Daytime Pinhole observation creation and tracking the same way as telluric standards are: delete them if the original obs is deleted (unless it started execution), workflow-state inheritance, etc.
   * The calibration is per-obs and must go into the same group as telluric standards. This group was named "telluric group" everywhere, but that name no is no longer accurate. Therefore, it was renamed to "obs calibration group" across the board.

6. `tolerate no targets when generating daytime pinhole sequences` (`7537c07`):
   * Skip checking that a target exists when generating the sequence.

~6. `share targets with pinhole cal` (`960943a`):~
   * ~Daytime Pihole observations are done with a closed dome, but they still need a dummy target. The original observation's target environment is linked to it. This avoids having to create dummy targets for this, which we must then keep track of (eg: delete them if the original observation gets deleted).~

~NOTE: This PR will fail to compile until the new `lucuma-core` is released including https://github.com/gemini-hlsw/lucuma-core/pull/1467~


rpiaggio:
I'm converting this to draft. Turns out that linking the cal observation to the original target environment is not that good of an idea. I'll look for alternatives.

cquiroz:
> I'm converting this to draft. Turns out that linking the cal observation to the original target environment is not that good of an idea. I'll look for alternatives.

Are the targets well known in advance?

rpiaggio:
> Are the targets well known in advance?

There are no targets since it's just a flat. I'm exploring the implications of having it just be a target-less observation.

rpiaggio:
This PR now ready for review

rpiaggio:
BTW, the issue with the target was solved by tolerating the observation with no targets when generating the sequence. So, Daytime Pinholes have no target, as it should be.

modules/service/src/main/scala/lucuma/odb/service/GeneratorParamsService.scala
```diff
@@ -228,17 +228,24 @@ object GeneratorParamsService {
...
         val targetCheck =
-          params
-            .traverse: p =>
-              for
-                t <- p.targetId.toRightNel(MissingParam.forObservation("target"))
-                _ <- p.sourceProfile.toRightNel(MissingParam.forTarget(t, "source profile"))
-              yield ()
-            .void
+          if calibrationRole.contains(CalibrationRole.DaytimePinhole) then
+            ().asRight[NonEmptyList[MissingParam]]
```
swalker2m, line 240: :+1:

swalker2m (approved):
Looks excellent. :+1:

PR #2611, opened 2026-06-18
Add fullTimeEstimate to the ExecutionDigest in the graphQL API

toddburnside:


modules/schema/src/main/scala/lucuma/odb/json/sequence.scala
```diff
@@ -128,10 +128,11 @@ trait SequenceCodec {
...
       Json.obj(
-        "setup"       -> a.setup.asJson,
-        "setupCount"  -> a.setupCount.asJson,
-        "acquisition" -> a.acquisition.asJson,
-        "science"     -> a.science.asJson
+        "setup"            -> a.setup.asJson,
+        "setupCount"       -> a.setupCount.asJson,
+        "acquisition"      -> a.acquisition.asJson,
+        "science"          -> a.science.asJson,
+        "fullTimeEstimate" -> a.fullTimeEstimate.asJson
```
cquiroz, line 135: Funny that it was just a matter of exposing it

cquiroz (approved):
LGTM

modules/schema/src/main/scala/lucuma/odb/json/sequence.scala
toddburnside: Yeah. Pretty simple.

swalker2m: I think I was just being stingy about what we supply in the digest, since this is created as a fully elaborated JSON object and then matched to the query by Grackle.  I think this is fine though.

PR #2613, opened 2026-06-18
Shortcut 8952: Support exchange time at Gemini

swalker2m:
Adds support for tying a proposal from a Subaru or Keck PI to a Gemini CfP and submitting it.  This adds an `exchangePartner` field to the Queue and Classical proposal type inputs.  The `exchangePartner` and `partnerSplits` are mutually exclusive.  If you specify one, the other may not be set.  On submission, we guarantee that one is defined (and if it is partner splits, that they add to 100 as before).


swalker2m:
Note, this is a PR to the [original unmerged draft PR](https://github.com/gemini-hlsw/lucuma-odb/pull/2603).

PR #2614, opened 2026-06-19
Fix SED-type switching on Gaussian source profiles

toddburnside:
Switching a target's SED between band-normalized and emission-lines on a Gaussian profile failed with "Not a band normalized spectral definition." or "Not an emission lines spectral definition ...". Unlike Point/Uniform, GaussianInput.EditBinding applied the spectral-definition edit unconditionally, so a type switch tried to read the existing definition as the new type and failed.

Add a shared SpectralDefinition.createOrEdit helper (replace on SED-type mismatch, edit on match) in SpectralDefinitionInput, reuse the existing matches check from there, and route GaussianInput.EditBinding through it via Integrated.CreateOrEditBinding. This gives Gaussian the same in-place SED-type switching semantics Point and Uniform already had.

Add updateTargets coverage for in-place Gaussian SED switches in both directions.


cquiroz (approved):
LGTM

PR #2618, opened 2026-06-19
Remove duplicate classes by itcclient depending on schemas

rpiaggio:
ITC client had duplicate encoders from way back then when they were separate repos.

This PR unifies the code by introducing a dependency on itcClient to schemas.


cquiroz (approved):
LGTM

PR #2620, opened 2026-06-19
Return empty guidingEnvironment for twilight and pinhole calibrations

rpiaggio:
These calibrations don't guide, so we now short-circuit to return an empty guiding if it queried (it is queried from observe).


swalker2m (approved):

cquiroz:
It could be good to add a test for this case to avoid regressions in the future

rpiaggio:
> It could be good to add a test for this case to avoid regressions in the future

Good idea: https://github.com/gemini-hlsw/lucuma-odb/pull/2621

PR #2627, opened 2026-06-23
Support gnirs coadds in ITC results

rpiaggio:
- Implement support for Gnirs coadds in ITC. This is a breaking change in the ITC schema since it's a new required parameter.
- Force coadds=1 when switching to S/N ETM (which doesn't support them).


itc/service/src/main/scala/lucuma/itc/input/GnirsImagingInput.scala
```diff
@@ -33,6 +35,9 @@ object GnirsImagingInput:
             GnirsCameraBinding("camera", camera),
             GnirsReadModeBinding("readMode", readMode),
             GnirsWellDepthBinding("wellDepth", wellDepth),
+            PosIntBinding("coadds", coadds),
             PortDispositionBinding("port", portDisposition)
           ) =>
-        (exposureTimeMode, filter, camera, readMode, wellDepth, portDisposition).parMapN(apply)
+        (exposureTimeMode, filter, camera, readMode, wellDepth, coadds, portDisposition).parMapN(
```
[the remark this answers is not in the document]
rpiaggio: This is innocuous: it's ignored downstream by the ITC. We actually want to keep it this way for when ITC supports it.

itc/service/src/main/scala/lucuma/itc/input/GnirsSpectroscopyInput.scala
```diff
@@ -45,6 +47,7 @@ object GnirsSpectroscopyInput:
             GnirsCameraBinding("camera", camera),
             GnirsReadModeBinding("readMode", readMode),
             GnirsWellDepthBinding("wellDepth", wellDepth),
+            PosIntBinding("coadds", coadds),
             PortDispositionBinding("port", portDisposition)
```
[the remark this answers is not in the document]
rpiaggio: Same comment as above.

itc/legacy-tests/src/test/scala/lucuma/itc/legacy/LegacyITCGnirsImgExpTimeSuite.scala
```diff
@@ -40,6 +41,7 @@ class LegacyITCGnirsImgExpTimeSuite extends CommonITCLegacySuite:
     camera = GnirsCamera.ShortBlue,
     readMode = GnirsReadMode.Bright,
     wellDepth = GnirsWellDepth.Shallow,
+    coadds = PosInt.unsafeFrom(1),
```
cquiroz, line 44: Should the test iterate over coadds? not sure what range it has though

itc/service/src/main/scala/lucuma/itc/legacy/package.scala
```diff
@@ -29,39 +30,51 @@ case class ItcSourceDefinition(
 ):
   export target.*
 
+extension (mode: ObservingMode)
+  /** The number of coadds for instruments that support it (currently GNIRS), else None. */
```
cquiroz, line 34: Really only gnirs supports coadds? i think ghost and gpi do as well but maybe that's not relevant in this context

cquiroz (approved):
LGTM

itc/legacy-tests/src/test/scala/lucuma/itc/legacy/LegacyITCGnirsImgExpTimeSuite.scala
rpiaggio: I don't think it has a formal upper bound and this test is mostly to test transport. Not sure we should iterate: I'm fine assuming that if one integer is encoded correctly, all of them will.

itc/service/src/main/scala/lucuma/itc/legacy/package.scala
rpiaggio: Currenty, only Gnirs. We can add the new ones when they are available.

PR #2630, opened 2026-06-23
Batch execution digest resolver to avoid N+1

cquiroz:
I've been looking at the traces for the very large query that explore issues with a large program containing 258 observations. 
This PR is the result of one of the issues found where we do n+1 queries instead of batching them.

The execution.digest `EffectHandler `called `selectExecutionDigest` once per observation
On a list query that's one DB round-trip and one transaction per result row.

This PR rewrites `runEffects` to make a single `selectManyExecutionDigest(oids) call (WHERE c_observation_id IN (…)).` Per-observation

In my sample program this reduces the total set of transactions by 258 comparing traces before and after the change


toddburnside (approved):
Good discovery. LGTM

PR #2632, opened 2026-06-23
Shortcut 7036: Cap per-program resource count (observations, groups, targets, attachments, notes)

swalker2m:
## What

Adds a per-program **resource limit**: a combined cap on the number of user-created observations, groups, targets, attachments, and program notes associated with a single program. Defaults to **1000** and is raisable per-program by staff.

- New staff-gated mutation `setProgramResourceLimit(programId, limit)`.
- `Program.resourceLimit` and `Program.resourceCount` exposed for reads (no staff gating).
- Exceeding the limit fails with `OdbError.ProgramResourceLimitExceeded`.

## Why

The API lets script writers programmatically add observations/groups/targets/etc. This guards against mistakes and malicious bulk creation by capping the total per program.

## How

- **`V1182__program_resource_limit.sql`** — `t_program.c_resource_limit` (NOT NULL DEFAULT 1000) holds the cap. The running count lives in a **dedicated `t_program_resource_count` table** (not on `t_program`), maintained O(1) by immediate per-row triggers on the five contributing tables. Enforcement is a **deferrable-initially-deferred** constraint trigger that raises a custom SQLSTATE (`LU001`); the count is exposed via the `v_program` view.
- The DB exception is mapped centrally in `MutationMapping` to `OdbError.ProgramResourceLimitExceeded` (the trigger is deferred, so it surfaces at commit across all mutations).

### Counting rules
Counts only `present`, non-system rows: observations/targets exclude `c_calibration_role`; groups exclude `c_system`; program notes count when present; attachments always count (no existence/role columns). System/calibration *programs* are exempt entirely.

## Reviewer notes

- **Why the count is off `t_program`:** an earlier design kept the counter on `t_program` itself. That **deadlocked** under concurrent same-program inserts — child inserts take `FOR KEY SHARE` on the program row (FK), and the counter `UPDATE` wanted `FOR NO KEY UPDATE`, producing a mutual-upgrade cycle. It also wrote `t_program` on every resource change, polluting the chronicle/audit log and firing spurious edit notifications. Nothing FK-references the dedicated count table, so its updates only serialize. The `concurrentInsertDeadlock` test is the regression guard (it deadlocks on the old design, passes on this one).
- **Lowering the limit below current usage** (e.g. `limit: 0` to freeze a program) is allowed and succeeds **with a warning** rather than failing — enforcement only fires on resource inserts, so a frozen program simply blocks further additions.
- **Drift safety:** `program_resource_count_actual()` recomputes from scratch and is the source of truth; the `programResourceLimit` suite includes a reconciliation test asserting the maintained counter matches the recomputed value across insert / soft-delete / restore / hard-delete / cross-program-move. Any future migration that bulk-modifies a resource table with triggers disabled should re-run the recompute.
- **Migration:** introduces `V1182`.

## Tests
`programResourceLimit` (counting, attachments, soft-delete, readable fields, staff-only, below-usage warning, freeze-at-0, reconciliation) and `concurrentInsertDeadlock`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


modules/service/src/main/resources/db/migration/V1182__program_resource_limit.sql
```diff
@@ -0,0 +1,212 @@
+-- Cap the number of user-created resources (observations + groups + targets +
+-- attachments + program notes, combined) that may be associated with a single
```
cquiroz, line 2: The cost of notes has to be much lower than observations. why do we put them all in the same bucket? Should we care more of the expensive resources?

modules/service/src/main/resources/db/migration/V1182__program_resource_limit.sql
```diff
@@ -0,0 +1,212 @@
...
+$$ language plpgsql;
+
+-- ===========================================================================
+-- Counter maintenance (immediate). One trigger function shared by all
+-- contributing tables; the "counts toward the limit" predicate is selected on
+-- TG_TABLE_NAME (PL/pgSQL plans each branch lazily, so a column is only
+-- referenced for the table that has it). Handles soft-delete, restore, program
+-- moves, and role changes.
+-- ===========================================================================
+create or replace function update_program_resource_count()
```
cquiroz, line 96: very nice

modules/service/src/main/resources/db/migration/V1182__program_resource_limit.sql
swalker2m: That's a good point.  I just added everything with a GID.  Maybe we should ignore notes or heavily discount them.

cquiroz: I suppose it is not very critical, it would be odd to have a program with way more notes than observations

cquiroz (approved):
Seems good, 1000 is a good limit to start with

toddburnside (approved):
Looks great to me. The separate table approach is a nice touch. 😆 I think we should go with this as a starting point. If Notes or something else turns out to be a sticking point, we can revisit. Or, they can just bump the limit for the program. Per program limit was a good idea.

PR #2635, opened 2026-06-24
[Itc] Reinstate coadds with a bunch of diagnostic tools

rpiaggio:
1) Make the coadds optional, so current clients don't break.
2) A logging to get information of where the issue is.
3) Add contention in ITC, avoiding it from being swamped in case of slow calculations.
4) Add an env variable to opt out of actually using the coadds in the computation.
5) Add redis to flake.nix so it's easy to run locally.


cquiroz:
Seems heavy handed but we could revert after the issue is solved

cquiroz (approved):
LGTM

PR #2638, opened 2026-06-24
Restore uniqueness of static rows for observations

cquiroz:
This was discovered testing GHOST execution in observe. In case we had errors we'd endup with a duplicated static row for the ghost observations. This is because after an error a new visit is added but I think this would be triggered every time we do a new visit.


modules/service/src/main/resources/db/migration/V1186__static_config_unique.sql
```diff
@@ -0,0 +1,48 @@
+-- Each execution static table is meant to hold exactly one row per observation.
+-- V1117 deduped and dropped c_visit_id for some instrument and implicitly dropped
+-- the UNIQUE (c_observation_id, c_visit_id) constraint)
```
swalker2m, line 3: Will this just expose an underlying issue that causes us to duplicate static configurations, or were we relying on a constraint violation somewhere to prevent it?  I'm asking myself this as much as anything since I've been the one monkeying around with this.

swalker2m: Wondering why we don't just make the observation id be the primary key at this point.

cquiroz: I think the key is here
https://github.com/gemini-hlsw/lucuma-odb/blob/sc-9120-ghost-static-config-can-be-duplicated-in-odb/modules/service/src/main/scala/lucuma/odb/service/SequenceService.scala#L1308

We rely on that on `ON CONFLICT DO NOTHING` to avoid inserting duplicates with the `UNIQUE` constraing but that was dropped

swalker2m:
```sql
        INSERT INTO t_ghost_static (
          c_program_id,
          c_observation_id,
          ...
        )
        SELECT
          o.c_program_id,
          $observation_id,
          $ghost_static
        FROM t_observation o
        WHERE o.c_observation_id = $observation_id
        ON CONFLICT DO NOTHING
        RETURNING c_static_id
```
Note the `ON CONFLICT DO NOTHING RETURNING c_static_id`.  So I think we were relying on conflict to prevent this before and your update should fix it.  I think we should drop the `c_static_id` and just use `c_observation_id`, but that can be done in another update.

swalker2m (approved):
Looks great, thanks for fixing this.

PR #2639, opened 2026-06-24
Revert diagnostic mechanisms on ITC

rpiaggio:


cquiroz (approved):
LGTM, why the version bump?

rpiaggio:
Appearently, the constructor for `BinaryEffectfulCache` is public in the library, and that changed. Also a public debug member of the legacy package was removed which should never have been public.

cquiroz:
we updated the reddit library not long ago, could that be related?

rpiaggio:
> we updated the reddit library not long ago, could that be related?

I checked that. That update got into dev previously and things were working after the upgrade.

PR #2642, opened 2026-06-25
Fix Gnirs XD ITC graph and single exposure S/N

rpiaggio:
This actually required fixing in OCS and processing here. JARs are updated.


cquiroz (approved):
LGTM

PR #2646, opened 2026-06-26
Shortcut 9121: exchange observations

swalker2m:
## Exchange observations (Keck / Subaru)

Adds support for **exchange observations** — observations belonging to an exchange proposal, where a Gemini PI requests time at Keck or Subaru. Builds on the already-merged exchange *Call for Proposals* / *proposal* support.

These observations are intentionally lightweight: Keck/Subaru instruments are **not** supported by ITC or AGS, so there is no exposure-time mode, no guide-star calculation, and no generated sequence. An exchange observation captures only its target/conditions/scheduling constraints (pre-existing observation fields), an **exchange instrument** (a `KeckInstrument` or `SubaruInstrument`), and a **total requested time**.

Modeled after the existing Gemini *Visitor* family: a new observing **mode** that reveals the exchange instruments.

### Highlights
- New `ExchangeObservingModeType` with `EXCHANGE_KECK` / `EXCHANGE_SUBARU`; the mode tag encodes the observatory.
- New `t_exchange` table with nullable `c_keck_instrument` / `c_subaru_instrument` (exactly-one CHECK) and `c_total_request_time`. `c_observing_mode_type` is a **generated column** derived from whichever instrument is set, and still serves in the PK and composite FK to `t_observation`.
- Keck/Subaru are **not** `Instrument`s — `instrument` is now optional on the observing-mode/exec-config GraphQL surface, and `InstrumentExecutionConfig` gains an `Exchange` case.
- Generator produces an "alien" execution digest directly from `totalRequestTime` (no ITC/AGS), like Visitor.
- Observatory/instrument consistency (mode vs. the proposal's CfP) is a **soft workflow validation** in `ObservationWorkflowService` — mirroring how Gemini validates instrument-vs-CfP — rather than a create-time gate.
- Create / update / clone supported; tests cover all three plus the workflow validations.

### ⚠️ Depends on lucuma-core
Requires gemini-hlsw/lucuma-core#1480 to be **merged and released**. `build.sbt` currently points at a local SNAPSHOT of that branch, so **CI will be red until core is released and the version is repointed**.

Opening as a **draft** so the UI team can review the new schema (exchange observing mode, optional `instrument`, exchange instrument enums) and update their code before either PR merges.


cquiroz:
> These observations are intentionally lightweight: Keck/Subaru instruments are **not** supported by ITC or AGS, so there is no exposure-time mode, no guide-star calculation, and no generated sequence. An exchange observation captures only its target/conditions/scheduling constraints (pre-existing observation fields), an **exchange instrument** (a `KeckInstrument` or `SubaruInstrument`), and a **total requested time**.
> 
> Modeled after the existing Gemini _Visitor_ family: a new observing **mode** that reveals the exchange instruments.

Should they be unified with visitors? i guess we have a few more fields for visitors and ags does run

modules/schema/src/main/scala/lucuma/odb/json/sequence.scala
```diff
@@ -226,48 +226,44 @@ trait SequenceCodec {
...
         a match
+          case InstrumentExecutionConfig.Exchange        => "instrument" -> Json.Null // no exchange instance
           case i@InstrumentExecutionConfig.Flamingos2(_) => "flamingos2" -> i.asJson
           case i@InstrumentExecutionConfig.Ghost(_)      => "ghost"      -> i.asJson
           case i@InstrumentExecutionConfig.GmosNorth(_)  => "gmosNorth"  -> i.asJson
           case i@InstrumentExecutionConfig.GmosSouth(_)  => "gmosSouth"  -> i.asJson
           case i@InstrumentExecutionConfig.Gnirs(_)      => "gnirs"      -> i.asJson
           case i@InstrumentExecutionConfig.Igrins2(_)    => "igrins2"    -> i.asJson
-          case i@InstrumentExecutionConfig.Visitor(_)    => "visitor"    -> i.asJson
+          case InstrumentExecutionConfig.Visitor(vi)     => "instrument" -> vi.asJson  // no visitor instance
```
toddburnside, line 266: It seems like for Exchange and Visitor, the "instrument" is actually already set? But, I guess this match needs to return some Json.

modules/service/src/main/resources/db/migration/V1195__exchange_observations.sql
```diff
@@ -0,0 +1,132 @@
...
+    ) STORED,
+
+  PRIMARY KEY (c_observation_id, c_observing_mode_type),
+  FOREIGN KEY (c_observation_id, c_observing_mode_type)
+    REFERENCES t_observation(c_observation_id, c_observing_mode_type)
+    ON DELETE CASCADE DEFERRABLE INITIALLY DEFERRED
+);
+
+-- Exchange observations have no exposure time mode.
+CREATE OR REPLACE FUNCTION check_etm_consistent()
```
toddburnside, line 38: Wow. Does this function need to be updated for each new mode?

toddburnside (approved):
Seems good to me. I'll start look at what needs to be done in explore to accommodate this.

modules/schema/src/main/scala/lucuma/odb/json/sequence.scala
swalker2m: Right, it could be reworked to avoid it (e.g., using from `Json.fromFields` artfully applied) but as is there needs to be a (field name, JSON) tuple for all the cases.

modules/service/src/main/resources/db/migration/V1195__exchange_observations.sql
swalker2m: Yeah, the requirements differ in the various cases.  We could default to not complain if the mode is unknown instead, but it seems like that would be license to not consider it.

Now that I look at it though, I don't think there's any reason for visitor and exchange to be treated differently wrt exposure time modes.

toddburnside: Yeah, I think it's better to raise the exception so that it's not forgotten.

PR #2647, opened 2026-06-26
Itc failures cache

cquiroz:
I keep checking the traces for a very large program. The program has 258 observation and from those 27 show itc "deterministic" errors, meaning they are not due to connection errors but will be errors always regardless of how often we repeat them.

Measurements for this particular case show that almost 60% of the time is spent requerying the itc for the same error

This PR proposes to store these errors in db acting as a cache, equivalent to the succesful itc case. Random errors will be retried again.

itc version changes will clear the table as it is already done for the successful case

A question for @swalker2m. Should itc version changes invalidate obscalc? I don't see that happening for the succesful case either

### DIFF
--- a/modules/service/src/main/resources/db/migration/V1187__itc_result_failure.sql
+++ b/modules/service/src/main/resources/db/migration/V1187__itc_result_failure.sql
@@ -1,28 +1,9 @@
--- Persistent cache for deterministic ITC failures.
--- Keyed by (program_id, observation_id) with a hash column matching the ITC input hash.
---
--- A hash mismatch means the observation changed, so the cached failure is ignored and
--- ITC is re-called.
-CREATE TABLE t_itc_result_failure (
-  c_program_id      d_program_id     NOT NULL,
-  c_observation_id  d_observation_id NOT NULL,
-  c_hash            bytea            NOT NULL,
-  c_error_message   text             NOT NULL,
-  FOREIGN KEY (c_program_id, c_observation_id)
-    REFERENCES t_observation(c_program_id, c_observation_id)
-    ON DELETE CASCADE,
-  CONSTRAINT t_itc_result_failure_pkey PRIMARY KEY (c_program_id, c_observation_id)
-);
+-- Truncate to allow schema change (c_asterism_results was NOT NULL).
+TRUNCATE TABLE t_itc_result;
 
--- Extend the ITC version trigger to also clear failures when ITC is updated.
-CREATE OR REPLACE FUNCTION itc_version_update()
-  RETURNS trigger AS $$
-BEGIN
-  NEW.c_last_update = NOW();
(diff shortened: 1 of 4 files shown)


modules/service/src/main/resources/db/migration/V1187__itc_result_failure.sql
```diff
@@ -0,0 +1,28 @@
...
+CREATE TABLE t_itc_result_failure (
+  c_program_id      d_program_id     NOT NULL,
+  c_observation_id  d_observation_id NOT NULL,
+  c_hash            bytea            NOT NULL,
+  c_error_message   text             NOT NULL,
+  FOREIGN KEY (c_program_id, c_observation_id)
+    REFERENCES t_observation(c_program_id, c_observation_id)
+    ON DELETE CASCADE,
+  CONSTRAINT t_itc_result_failure_pkey PRIMARY KEY (c_program_id, c_observation_id)
+);
```
swalker2m: Is there an advantage to having a separate table instead of adding a `c_error_message` column to the existing `t_itc_result`?

swalker2m (approved):
Seems like a great idea, as long as we're sure we don't cache transient or spurious errors.

modules/service/src/main/resources/db/migration/V1187__itc_result_failure.sql
cquiroz: `t_itc_result` has a not null `jsonb` column `c_asterism_result`. Maybe set it to `{}`? otherwise it would have to be nullable

swalker2m: I'd say nullable with a constraint that exactly one of `c_asterism_result` or `c_error_message` is not null.

cquiroz: Implemented, now it will use a single table

modules/service/src/main/scala/lucuma/odb/service/ItcService.scala
```diff
@@ -184,43 +185,61 @@ object ItcService {
...
+                if h === inputHash
+                result              <- itcOpt.map(_.asRight).orElse(errOpt.map(msg => OdbError.ItcError(msg.some).asLeft))
+              yield result
+
+      override def selectOne(
+        pid:    Program.Id,
+        oid:    Observation.Id,
+        params: GeneratorParams
+      )(using Transaction[F]): F[Option[Itc]] =
+        selectOneCached(pid, oid, params).map(_.flatMap(_.toOption))
```
swalker2m: `selectOne` seems to only be used externally by the `GeneratorContext`.  I wonder if we should just change the signature to the `selectOneCached` return value: `F[Option[Either[OdbError, Itc]]]`.  In other words, it seems like this will benefit the context lookup too.

PR #2651, opened 2026-06-27
Support GNIRS IFU in ITC

rpiaggio:
As specified in https://app.shortcut.com/lucuma/story/9117/gnirs-ifu-itc-default-analysis-method,  the default analysis method is set to Sum of 2x2 elements at the center.

This is an incremental PR atop of https://github.com/gemini-hlsw/lucuma-odb/pull/2649.

PR #2653, opened 2026-06-28
Gnirs IFU observing mode

rpiaggio:
This PR requires a release of core that includes https://github.com/gemini-hlsw/lucuma-core/pull/1482.

This PR sits atop https://github.com/gemini-hlsw/lucuma-odb/pull/2651.

Since LongSlit and IFU modes differ only on their FPU, I decided to unify in a `GnirsSpectroscopy` observing mode with a wider definition for FPU. The same criterion was applied for the DB and the schema. This implies a mechanical rename of instances named "long slit" -> "spectroscopy".

After this PR, GNIRS IFU modes generate a sequence. The Acquisition must be adjusted though, which will be done in a future PR.


swalker2m (approved):
A cursory review, admittedly, but it looks good.  I think it was a good idea to combine the spectroscopy cases.  Presumably we'll do that for GMOS too.

swalker2m (approved):

PR #2655, opened 2026-06-29
Add missing tag in db

rpiaggio:
I'm merging this since I forgot to include it in the last PR and it broke ODB.


rpiaggio:
Turns out the migration was correctly left out after all. Closing this.

PR #2658, opened 2026-06-29
Consider the sky position when calculating AGS for ghost

cquiroz:
This requires a core release at https://github.com/gemini-hlsw/lucuma-core/pull/1483

To be exact the feature in core is not needed for the fix but it would change the test results. Unfortunately it can be merged as there is a set of warnings around gnirs ifu that need to be fixed first, but the code could be reviewed right away.


toddburnside (approved):
LGTM

PR #2665, opened 2026-06-30
Fix gnirs configuration joins

rpiaggio:
Fixes the recently-introduced assertion error for GNIRS configurations.

The long-slit and IFU configurations shared one view with no discriminator, so the wrong type resolved instead of null.

Now each of them declares its FPU column as key so the wrong one correctly resolves to null.


toddburnside (approved):
LGTM

PR #2666, opened 2026-07-01
Gnirs IFU offsets

rpiaggio:
For IFU we don't want on-slit/nod-to-sky offsets, just regular (p, q) offsets.


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -10325,9 +10325,15 @@ type GnirsSpectroscopy {
...
-  explicitTelescopeConfigs: SlitTelescopeConfigs
+  "Long-slit telescope configs. Present iff the FPU is a long slit (`fpuSlit`); null for IFU."
+  telescopeConfigsSlit: SlitTelescopeConfigs
+  defaultTelescopeConfigsSlit: SlitTelescopeConfigs
+  explicitTelescopeConfigsSlit: SlitTelescopeConfigs
+
+  "IFU telescope configs (full p/q offsets). Present iff the FPU is an IFU (`fpuIfu`); null for long slit."
+  telescopeConfigsIfu: [TelescopeConfig!]
+  defaultTelescopeConfigsIfu: [TelescopeConfig!]
+  explicitTelescopeConfigsIfu: [TelescopeConfig!]
```
swalker2m: Now that we have another slit vs IFU mutually exclusive property, we could consider grouping the slit properties into `GnirsSlit` and the IFU properties into `GnirsIfu` and then adding a layer of hierarchy (`slit: GnirsSlit` vs `ifu: GnirsIfu`).

swalker2m (approved):
Looks great.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
rpiaggio: Good idea :+1: I'll implement that before merging.

rpiaggio: I'll actually give both types different shapes. For IFU there's not derived default from other properties, just an initial seed value, so I don't think it's appropriate to have a "default" and "coalesced" value.

I'll just initialize with a seed value depending on the initial resultion mode. Explore uses this presets already, so I'll move them to core.

PR #2667, opened 2026-07-01
Gnirs off slit steps should not contribute to S/N

rpiaggio:
Mirror the Igrins-2 and Flamingos2 logic to prevent off-slit offsets from contributing to the S/N.

Actually, I changed the logic for Igrins-2 so that an offset defined as `NodToSky` with p == 0 and q > slitlength/2 is considered a sky offset (previously it would be considered as on target). If the previous logic was correct @cquiroz please let me know and we can revert with a comment.

### DIFF
--- a/modules/sequence/src/main/scala/lucuma/odb/sequence/gnirs/spectroscopy/Science.scala
+++ b/modules/sequence/src/main/scala/lucuma/odb/sequence/gnirs/spectroscopy/Science.scala
@@ -9,6 +9,7 @@ import cats.Monad
 import cats.data.EitherT
 import cats.data.NonEmptyList
 import cats.data.State
+import cats.syntax.either.*
 import cats.syntax.option.*
 import cats.syntax.order.*
 import eu.timepit.refined.types.numeric.NonNegInt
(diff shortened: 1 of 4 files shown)


modules/sequence/src/main/scala/lucuma/odb/sequence/gnirs/spectroscopy/Science.scala
```diff
@@ -78,25 +81,26 @@ object Science:
...
     // Inline nighttime calibrations (flat + arc).  Empty for telluric sequences.
     cals:         Option[NonEmptyList[ProtoStep[GnirsDynamicConfig]]]
   ):
     /**
      * Cycle count: round up so that we always deliver at least the requested
-     * number of exposures. Sky-only offsets in the cycle (if any) are not
-     * given special treatment — we cycle blindly through whatever the
-     * observing mode defines.
+     * number of on-source exposures. Sky steps (target off the slit) don't
+     * contribute to the S/N, so cycles with sky offsets require extra repeats.
```
cquiroz, line 91: Never thought about that. I think it is correct to apply the same logic to igrins2. What about other longslit instruments?

cquiroz (approved):
LGTM

modules/sequence/src/main/scala/lucuma/odb/sequence/gnirs/spectroscopy/Science.scala
rpiaggio: The only other one that uses sky offsets for the moment is, I think, Flamingos2. The logic is a bit different there and already does count offsets with p==0 and q > slit/2 as sky.

PR #2668, opened 2026-07-01
Use coadds from smart cal for Gnirs arcs and flats

rpiaggio:


modules/service/src/main/scala/lucuma/odb/service/SmartGcalService.scala
```diff
@@ -474,10 +483,14 @@ object SmartGcalService:
...
-    private def selectGcal(tableName: String, where: List[AppliedFragment]): AppliedFragment =
+    private def selectGcal(
+      tableName:  String,
+      where:      List[AppliedFragment],
+      valueCols:  List[String] = List("s.c_exposure_time")
+    ): AppliedFragment =
       sql"""
         SELECT #${gcalColumns("g")},
-               s.c_exposure_time
+               #${valueCols.mkString(",\n               ")}
```
cquiroz: why do we need the extra spaces?

swalker2m: Also, for what it's worth, there's an `encodeColumns` utility that does this (albeit without the extra spaces).

```scala
  #${encodeColumns("s".some, valueCols)}
```

if we strip the `s` from the column names.

swalker2m (approved):
Looks great.

modules/service/src/main/scala/lucuma/odb/service/SmartGcalService.scala
rpiaggio: > ```scala
> #${encodeColumns("s".some, valueCols)}
> ```

Those spaces are the same ones the code has within the triple-quote block. Apparently they were generated so that if the query is printed in a log, everything will align nicely.

rpiaggio: I'll switch to Shane's suggestion.

PR #2672, opened 2026-07-02
Shortcut 9219: Allow exchange partners to override submission deadline

swalker2m:
Normal Gemini partners may specify their own Call For Proposals submission deadline, but until now exchange partners were stuck with the CfP default deadline.  This PR remedies that situation, allowing Keck and Subaru to have independent submission deadlines just like "normal" partners.


toddburnside (approved):
LGTM. I'll update explore to not break.

PR #2675, opened 2026-07-03
[sc-8607] implement chown

tpolecat:
This implements a chown operation that transfers all programs (at this point there only be one) from a guest to a standard user by replacing the guest as PI. It is a REST endpoint to avoid tight coupling between SSO and the ODB. SSO calls this endpoint when a guest user authenticates as an existing user (when a guest authenticates as a *new* user we just turn the guest into a standard user, but that doesn't work in this case).

After this update SSO will require another environment variable `LUCUMA_ODB_HOSTNAME`.


modules/service/src/main/scala/lucuma/odb/graphql/ChownRoutes.scala
```diff
@@ -0,0 +1,141 @@
...
+    calculator:     TimeEstimateCalculatorImplementation.ForInstrumentMode,
+    httpClient:     Client[F],
+    itcClient:      ItcClient[F],
+    gaiaClient:     GaiaClient[F],
+    horizonsClient: HorizonsClient[F]
+  ): HttpRoutes[F] =
+    val dsl = Http4sDsl[F]; import dsl.*
+    given EntityDecoder[F, ChownBody] = jsonOf
+    HttpRoutes.of[F]:
+      case req @ POST -> Root / "chown" =>
```
hugo-vrijswijk, line 88: I get this is mirroring linux's chown, but maybe the endpoint should have a bit more descriptive of a name? Since now the path doesn't mention programs, and assumes linux familiarity.
Also couldn't this be a graphql mutation? Since all other program-related actions are gql

swalker2m (approved):
Looks great to me.  `chown` seems fine to me since it is meant for the SSO service / programmers but your mileage may vary.

PR #2676, opened 2026-07-03
Cors fix

cquiroz:
There is a slight security vulnerability with cors where we only check that the calling host ends with the domain, e.g. `gemini.edu` and that allows domains partially matching to be ok like `somegemini.edu`. 

The fix is simply to request domains ending with a `.gemini.edu` The dot allows e.g. `explore.gemini.edu` but avoids the partially matching bug.

I made a new module `common` to hold the logic and make it reusable in sso/odb. It maybe overkill but we could move a few more bits there in the future

### DIFF
--- a/modules/common-middleware/src/main/scala/lucuma/common/middleware/CorsMiddleware.scala
+++ b/modules/common-middleware/src/main/scala/lucuma/common/middleware/CorsMiddleware.scala
@@ -3,10 +3,19 @@
 
 package lucuma.common.middleware
 
+import cats.*
+import cats.implicits.*
+import org.http4s.HttpRoutes
 import org.http4s.Uri
-import cats.syntax.eq.*
+import org.http4s.Uri.Scheme
+import org.http4s.server.middleware.CORS
+
+import scala.concurrent.duration.*
 
 object CorsMiddleware:
+
+  type Middleware[F[_]] = Endo[HttpRoutes[F]]
+
   /**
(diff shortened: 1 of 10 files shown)


modules/common-middleware/src/main/scala/lucuma/common/middleware/CorsMiddleware.scala
```diff
@@ -0,0 +1,46 @@
...
+   *
+   * @param host The host to check (e.g., "sub.example.com")
+   * @param allowedDomains A list of trusted domains (e.g., List("example.com"))
+   * @return true if the host is allowed, false otherwise
+   */
+  def isAllowed(host: Uri.Host, allowedDomains: List[String]): Boolean =
+    val h = host.value.toLowerCase
+    allowedDomains.exists: d =>
+      val dd = d.toLowerCase
+      h === dd || h.endsWith("." + dd)
```
cquiroz, line 31: adding the `.` is the actual bug fix

hugo-vrijswijk (approved):
Nice, could you add it to Resource too? The new module seems useful

cquiroz:
> Nice, could you add it to Resource too? The new module seems useful

Yes, does it use `cors` as well? you may need to add a variable to include the supported domains

hugo-vrijswijk:
> > Nice, could you add it to Resource too? The new module seems useful
> 
> Yes, does it use `cors` as well? you may need to add a variable to include the supported domains

It does I think it is already fully set up

PR #2679, opened 2026-07-05
Shortcut 7665: Allow selecting which asterism target drives the ITC (signal-to-noise) calculation

swalker2m:
## Summary

When an observation has multiple targets, the ITC previously chose a
representative target automatically — the **brightest** target, i.e. the one
requiring the **shortest** exposure time — to drive the exposure-time
calculation. This PR lets the user explicitly choose which asterism target is
used for the signal-to-noise measurement, while still running the ITC for *all*
targets so saturation can be checked against the whole asterism.

Resolves SC-7665.

## API

`TargetEnvironment` gains a nullable `explicitSignalToNoiseTarget: Target`, and
`TargetEnvironmentInput` gains `explicitSignalToNoiseTargetId: TargetId`.
Following the `explicitBase` convention already in this type, the name signals
that it is an override:

- The selected target is always a member of the observation's asterism.
- `null` means "no override" — the ITC picks the brightest target, preserving
  today's behavior. Existing observations need no migration. The target actually
  used in that case is discoverable via `observation.itc { … selected { targetId } }`.
- On edit the field is nullable: `null` reverts to automatic, absent leaves it
  unchanged.

## Design

The selection is modeled as a boolean flag on the `t_asterism_target` link row
rather than a pointer on `t_observation`. This makes membership a structural
invariant (the flag lives on the membership row), avoids a circular foreign key,
and means removing a target from the asterism removes any selection along with
it. A partial unique index (`WHERE c_is_signal_to_noise_target`) enforces at most
one selected target per observation.

## Behavior

- **Validation:** selecting a target not in the (post-edit) asterism fails with a
  clean `InvalidArgument`. When asterism and selection are edited in the same
  input, validation runs against the *post-edit* asterism, and a contradiction
  rolls back the whole mutation.
- **`updateObservations` / `createObservation`** (replace-the-asterism path): the
  selection is applied after the asterism write; a full asterism replacement
  preserves the selection if the target remains, drops it otherwise.
- **`updateAsterisms`** (diff-based ADD/DELETE path): deleting the flagged target
  clears the selection; adding/deleting other targets leaves it untouched.
- **Cloning** copies the selection.

## Generator / ITC

The flag flows through `v_generator_params` → `GeneratorParamsService` →
`ItcInput` → `ItcService`, where the "selected" result zipper is refocused onto
the chosen target (science only; acquisition still uses the brightest). All
targets are still sent to the ITC, so `ItcResultSet.all` — and the saturation
check — are unaffected. The selection is included in the `ItcInput` hash so
cached sequence/ITC results invalidate correctly when it changes.

Also corrects the `ItcResultSet` schema doc, which incorrectly described the
automatic pick as the "longest observation" (it is the brightest / shortest) and
referenced a non-existent `result` field.

## DB migration

`V1199__signal_to_noise_target.sql`:
- Adds `c_is_signal_to_noise_target` to `t_asterism_target` + the partial unique
  index.
- Recreates `v_observation` to expose the flagged target id (bare scalar
  subquery — no `LIMIT`, so a violated single-selection invariant surfaces as an
  error rather than being masked).
- Recreates `v_generator_params` to carry the flag to the generator.

## Testing

- New `mutation/explicitSignalToNoiseTarget` suite (16 tests): defaults,
  set/clear/move, non-member validation, asterism-replacement preserve/drop,
  both-in-one-input for create and edit (success + contradiction), and the
  `updateAsterisms` diff path (delete selected/other, add, add+delete,
  multi-observation).
- `query/itc`: new assertion that pinning a target flips the ITC `selected`
  result while `all` still lists every target.
- Regression suites (updateAsterisms, blindOffsetTarget, cloneObservation,
  updateObservations, GMOS science generation, time-and-count) pass.

## Follow-ups

- `lucuma-schemas` / frontend client regeneration to expose the new field.
- No default-selection UI logic here: a fresh multi-target asterism stays `null`
  (automatic) until the user picks.
- Separate question worth revisiting: whether the *automatic* default should be
  the faintest target (longest exposure, the worst case that guarantees every
  target reaches the requested S/N) rather than the brightest. Out of scope here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -6182,6 +6182,15 @@ input TargetEnvironmentInput {
...
   asterism: [TargetId!]
 
+  """
+  Explicitly selects which asterism target drives the signal-to-noise (ITC)
+  calculation, overriding the automatic choice. The target must be a member of
+  the observation's asterism (after applying any `asterism` edit in the same
+  input). Assign null to clear the selection and let a representative target be
+  chosen automatically.
+  """
+  explicitSignalToNoiseTargetId: TargetId
```
cquiroz, line 6192: What happens if the target is removed? is this set to null?

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -16809,6 +16820,17 @@ type TargetEnvironment {
...
   ): Target
 
+  """
+  The asterism target explicitly selected to drive the signal-to-noise (ITC)
+  calculation, overriding the automatic choice. This is always one of the
+  observation's asterism targets. When null, no explicit selection has been
+  made and a representative target is chosen automatically; the target that will
+  actually be used in that case is available via the `itc` field's `selected`
+  result. It is set with `TargetEnvironmentInput.explicitSignalToNoiseTargetId`
+  and is cleared automatically if the target is removed from the asterism.
```
cquiroz, line 16840: I guess this answer my previous question

modules/service/src/main/scala/lucuma/odb/graphql/input/TargetEnvironmentInput.scala
```diff
@@ -17,46 +17,50 @@ sealed trait TargetEnvironmentInput
...
+    asterism:                      Nullable[List[Target.Id]],
+    explicitSignalToNoiseTargetId: Nullable[Target.Id],
+    useBlindOffset:                Option[Boolean],
+    blindOffsetTarget:             Nullable[TargetPropertiesInput.Create],
+    blindOffsetType:               BlindOffsetType
   ) extends TargetEnvironmentInput:
     def limitToPreExecution(access: Access): Boolean =
       // staff can edit the blind offset for ongoing observations
-      access <= Access.Pi || asterism.isDefined || explicitBase.isDefined
+      access <= Access.Pi || asterism.isDefined || explicitBase.isDefined || explicitSignalToNoiseTargetId.isDefined
```
cquiroz, line 52: Why is this here? to allow changing the sn target in ongoin observations?

modules/service/src/main/scala/lucuma/odb/service/GeneratorParamsService.scala
```diff
@@ -176,7 +176,7 @@ object GeneratorParamsService {
       )(using Transaction[F]): F[Map[Observation.Id, Either[Error, GeneratorParams]]] =
         for
           paramsRows <- params
-          oms         = paramsRows.collect { case ParamsRow(oid, _, _, _, Some(om), _, _, _, _, _, _, _, _, _, _, _, _) => (oid, om) }.distinct
+          oms         = paramsRows.collect { case ParamsRow(oid, _, _, _, Some(om), _, _, _, _, _, _, _, _, _, _, _, _, _) => (oid, om) }.distinct
```
cquiroz, line 179: Named tuples?

cquiroz (approved):
LGTM

modules/service/src/main/scala/lucuma/odb/graphql/input/TargetEnvironmentInput.scala
swalker2m: As I understand it, if you are a PI or even if you're not a PI but any of the asterism, explicit base, or now the explicit s/n target are being edited then it must be for an observation in a pre-execution state.

PR #2680, opened 2026-07-05
Gnirs imaging phase 0

rpiaggio:
Gnirs can use LongCam or ShortCam. This is derived from the name in each row.


modules/service/src/test/scala/lucuma/odb/graphql/query/imagingConfigOptions.scala
```diff
@@ -173,7 +177,7 @@ class imagingConfigOptions extends OdbSuite {
   }
 
   test("adaptiveOptics false") {
-    val expect = allOptions  // for now, all are AO-free
+    val expect = allOptions.map(_.filter(!_.ao))
```
cquiroz, line 180: Why this change? I guess we are still all AO-free

rpiaggio: The Gnirs mode are declared with AO=yes in the phase0 table. Since we have to support AO for the 2027 release anyway, we decided with Andy to leave the table like that.

The test were thus adapted, I expect this to be their final form. Nothing is done yet with the AO=yes information, just less friction for when we implement it.

cquiroz: So in this case it means we can support AO, not that it always uses AO?
Would that option be hidden in explore for now?

rpiaggio: Yes, it's hidden for now.

cquiroz (approved):
LGTM

PR #2682, opened 2026-07-06
Observations query improvements

cquiroz:
This is another PR about improving performance. As before this was driven examining traces for a large (265 observations) program and focusing on the slowest queries, in particular the one reading observations.

It was discovered that quite few transactions in the traces where due to on a N+1 type of code in the odb. 

From a high-level point of view this PR includes two big changes in queries that can be batched

* `configuration.target` (`ConfigurationMapping`): resolve the batch in a single session, grouping by `referenceTime` and calling the existing batched `TrackingService.getCoordinatesSnapshotOrRegion`.
* `timeCharge` (`ExecutionMapping` + `TimeAccountingService`): added a batched `selectObservations(oids)` and switched the handler from the per-obs `effectHandler` to the batched pattern already used by `digestHandler`. 
`executionMapping` also uses the new `batchedEffectHandler`

Here's the summary of the results before and after:

| Metric (server side) | Before | After | Change |
|---|---:|---:|---:|
| `graphql-query` duration | 17.8 s | **6.6 s** | **−63%** |
| Total spans | 12,005 | **126** | −99% |
| DB sessions (`pool.allocate`) | 533 | **5** | −99% |
| Single-id per-observation executes | 530 | **0** | eliminated |

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/graphql/OdbMapping.scala
+++ b/modules/service/src/main/scala/lucuma/odb/graphql/OdbMapping.scala
@@ -49,6 +49,8 @@ import org.typelevel.otel4s.trace.Tracer
 import scala.concurrent.duration.*
 import scala.io.AnsiColor
 import scala.io.Source
+import java.nio.file.Files
+import java.nio.file.{Path => NIOPath}
 
 object OdbMapping {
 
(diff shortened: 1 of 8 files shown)


swalker2m (approved):
Looks like an excellent improvement.  Are there any more effect handlers we can convert?

cquiroz:
> Looks like an excellent improvement. Are there any more effect handlers we can convert?

I think so, I've been driving this from the point of view of explore queries but it seems there should be more cases like this

tpolecat:
Wow, this is great, thanks. I reviewed the handlers but missed this one (and probably others).

tpolecat:
I do think that in the long run we need to change something in our architecture so we're not doing queries that are a million lines of generated SQL. This got out of control before anyone noticed.

cquiroz:
> I do think that in the long run we need to change something in our architecture so we're not doing queries that are a million lines of generated SQL. This got out of control before anyone noticed.

Yes I'm working on an explore side change that will split the mega query into more maneagable queries. However, we don't control all client, any of our science users could issue on of those huge queries.

PR #2685, opened 2026-07-06
Sec fixes

cquiroz:
Fixes three identified security issues and adds a shared logging middleware

* The SSO server logger's `redactHeadersWhen` was a stub that returned false for every environment, so `Authorization`, `Cookie`, etc. were logged in the clear in all environments. Now redacts `Headers.SensitiveHeaders` everywhere except Local.

* The ORCID OAuth HTTP client had `logBody = true`. Those bodies could carry the ORCID client_secret and user access tokens.

*  The auth flow redirected the browser to a caller-supplied state value after login without validation — an open redirect enabling phishing on the trusted SSO origin. No risk of directly stealing a cookie or token but you could in theory do phishing this way

### DIFF
--- a/modules/sso-service/src/main/scala/Main.scala
+++ b/modules/sso-service/src/main/scala/Main.scala
@@ -226,6 +226,7 @@ object FMain extends AnsiColor {
         jwtWriter = config.ssoJwtWriter,
         publicUri = config.publicUri,
         cookies   = CookieService[F](config.cookieDomain, config.scheme === Scheme.https),
+        cookieDomain = config.cookieDomain,
       ) <+>
       GraphQLRoutes(localClient, pool, channels, SkunkMonitor.noopMonitor[F], wsb, schema)
     }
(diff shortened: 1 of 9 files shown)


modules/sso-service/src/main/scala/Main.scala
```diff
@@ -203,7 +203,7 @@ object FMain extends AnsiColor {
   def orcidServiceResource[F[_]: Async: Trace: Network](config: OrcidConfig) =
     EmberClientBuilder.default[F].build.map(org.http4s.client.middleware.Logger[F](
       logHeaders = true,
-      logBody    = true,
+      logBody    = false, // bodies carry the ORCID client_secret and user access tokens
```
toddburnside: Do we not need to redact sensitive headers here?

toddburnside (approved):
LGTM

modules/sso-service/src/main/scala/Main.scala
cquiroz: your're right I'll fix it before merging

PR #2687, opened 2026-07-07
Gnirs imaging

rpiaggio:
ObservingMode implementation and sequence generation.


swalker2m (approved):
These things always produce giant PRs, as many of my own PRs have been, so a thorough review is difficult but after a quick pass through I didn't see any issues.  Looks great. :+1:

PR #2688, opened 2026-07-07
Gnirs imaging + IFU guiding

rpiaggio:
This PR requires https://github.com/gemini-hlsw/lucuma-core/pull/1496 and a core release with that PR. It won't compile until then.

It also sits on top of https://github.com/gemini-hlsw/lucuma-core/pull/1496


swalker2m (approved):
:+1:

PR #2690, opened 2026-07-07
Also multiply exposures by coadds, not only readouts, in datasetEstimate

rpiaggio:
This seemed to be a bug where we mulitplied readouts by coadds but not exposures.


cquiroz (approved):
LGTM

PR #2694, opened 2026-07-07
Send observation edit events for ghost

cquiroz:
It was discovered that we didn't get `observationEdit` events when editing some ghost fields, like binning or read mode. 

Also we need to invalidate obs clac when ghost observations change


toddburnside (approved):
LGTM

PR #2695, opened 2026-07-08
Shortcut 5184: record original time estimate

swalker2m:
The goal of this PR is to record the time estimate at the moment we begin executing an observation.  Just slewing to an observation doesn't count, but instead we key off the first visit created for / co-opted by Observe.  In the schema, I've added a new `ObservationTimeEstimate` type that combines all the relevant information.

```graphql
type ObservationTimeEstimate {

  "Setup time estimates."
  setup: SetupTime!

  "Expected number of setups."
  setupCount: NonNegInt!

  "Science sequence time estimate."
  science: CategorizedTime!

  "Total time estimate: science time plus full setup time for every setup."
  total: CategorizedTime!

}
```

Much of this overlaps what we already have in `ExecutionDigest`.  Accordingly, I updated `ExecutionDigest` to have an `estimate: ObservationTimeEstimate!` field and deprecated `setup`, `setupCount` and `fullTimeEstimate`.

```graphql
type ExecutionDigest {

  "Time estimate for executing the observation."
  estimate: ObservationTimeEstimate!

  "Setup time calculations."
  setup: SetupTime! @deprecated(reason: "Use `estimate.setup` instead.")

  "Estimated number of setups required to complete this observation."
  setupCount: NonNegInt! @deprecated(reason: "Use `estimate.setupCount` instead.")

  "Acquisition sequence summary."
  acquisition: SequenceDigest

  "Science sequence summary."
  science: SequenceDigest

  "Full time estimate: science time plus full setup time for every setup."
  fullTimeEstimate: CategorizedTime! @deprecated(reason: "Use `estimate.total` instead.")

}
```

This update revealed an issue with trigger ordering and our event tests.  It became necessary to accept expected results in any order for some tests.

```scala
  // Like `subscriptionExpect`, but ignores the order in which events are
  // received.  Use when a single mutation produces events for multiple
  // entities (e.g., editing a target shared by several observations), where
  // the relative notification order is unspecified (it depends on the order
  // in which row-level triggers happen to fire).
  def subscriptionExpectUnordered( ...
```


cquiroz:
Why is this needed? because the estimates would vary as we execute steps?

swalker2m:
> Why is this needed? because the estimates would vary as we execute steps?

Right, the `ExecutionDigest` version is always the remaining time.  This captures the estimate before execution began and doesn't subsequently change.  From the [task request itself](https://app.shortcut.com/lucuma/story/5184/record-the-original-planned-observation-duration):

> It would be useful to store the original planned observation duration so that we can compare with the actual duration after an observation has been executed.  This value is the `timeEstimate` when execution starts.

and

> also the estimated number of acquisitions

cquiroz:
> > Why is this needed? because the estimates would vary as we execute steps?
> 
> Right, the `ExecutionDigest` version is always the remaining time. This captures the estimate before execution began and doesn't subsequently change. From the [task request itself](https://app.shortcut.com/lucuma/story/5184/record-the-original-planned-observation-duration):

Wouldn't you want to keep both remaining and original? Asking because we are dperecating the remaining fields

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -8367,11 +8375,14 @@ enum Existence {
 "Summarizes the execution setup time and sequences."
 type ExecutionDigest {
 
+  "Time estimate for executing the observation."
+  estimate: ObservationTimeEstimate!
```
cquiroz, line 8379: I see we have both the original and the current estimate. I think this was not included on the PR description

swalker2m: Right, I probably didn't make it clear.  The remaining time estimate is still there in the `ExecutionDigest`.  I've just moved it around in there and deprecated the fields that were used to access it before.  The original time estimate is being added and can be found in the `Execution` type next to the time accounting for the observation.

modules/service/src/main/resources/db/migration/V1206__original_time_estimate.sql
```diff
@@ -0,0 +1,87 @@
...
+  ADD CONSTRAINT original_estimate_all_or_none CHECK (
+    num_nonnulls(
+      c_orig_est_full_setup_time,
+      c_orig_est_reacq_setup_time,
+      c_orig_est_setup_count,
+      c_orig_est_sci_non_charged_time,
+      c_orig_est_sci_program_time,
+      c_orig_est_total_non_charged_time,
+      c_orig_est_total_program_time
+    ) IN (0, 7)
```
cquiroz, line 40: is this either 0 nulls or 7 nulls?

modules/service/src/main/resources/db/migration/V1206__original_time_estimate.sql
```diff
@@ -0,0 +1,87 @@
...
+--   * full setup time, reacquisition setup time, and setup count correspond
+--     to the SetupTime and setup count of the execution digest;
+--   * sci(ence) non-charged / program times are the CategorizedTime estimate
+--     for the science sequence;
+--   * total non-charged / program times are the full time estimate: science
+--     time plus full setup time for every setup, charged according to the
+--     science sequence observe class.
+
+ALTER TABLE t_observation
+  ADD COLUMN c_orig_est_full_setup_time        interval NULL CHECK (c_orig_est_full_setup_time        >= interval '0 seconds'),
```
cquiroz, line 21: Can it be enforced that these columns are write-once only?

cquiroz (approved):
LGTM

modules/service/src/main/resources/db/migration/V1206__original_time_estimate.sql
swalker2m: Yes.

modules/service/src/main/resources/db/migration/V1206__original_time_estimate.sql
swalker2m: I think it'd have to be via a trigger. I have a sneaking suspicion that we _will_ have to update these columns at some point when the original estimate is captured before some significant modification.  At any rate, I'll leave it out for now relying on the `UPDATE ... WHERE ... IS NULL` in the code that writes these.

PR #2696, opened 2026-07-08
In Gnirs acquisition S/N ITC, determine brightness and then invoke proper ITC

rpiaggio:
Details here: https://app.shortcut.com/lucuma/story/9291/gnirs-acquisition-s-n-calculations-using-wrong-filter-and-read-mode

Short summary: in Gnirs we must determine the acquisition mode from the target's brightness, which in turn is determined by the exposure time. In S/N mode this is a bit of a chicken-and-egg problem so we invoke ITC twice: first with a standard S/N of 10, which avoids misclassifying the target if a low or high S/N is requested, then with that result we determine the acquisition type and then we do a proper invocation of ITC with the requested S/N.

I've been struggling to determine where to put the logic and I decided to put it in the ItcService. Here it remains nicely isolated but in turn means that the acquisition mode must be returned since it can no longer be inferred from the returned exposure time. This becomes part of the ITC layer result and is easily cached without modifying downstream logic.

### DIFF
--- a/modules/schema/src/main/scala/lucuma/odb/data/Itc.scala
+++ b/modules/schema/src/main/scala/lucuma/odb/data/Itc.scala
@@ -15,6 +15,7 @@ import lucuma.core.data.Zipper
 import lucuma.core.enums.Flamingos2Filter
 import lucuma.core.enums.GmosNorthFilter
 import lucuma.core.enums.GmosSouthFilter
+import lucuma.core.enums.GnirsAcquisitionType
 import lucuma.core.enums.GnirsFilter
 import lucuma.core.model.Target
 import lucuma.core.util.Enumerated
@@ -176,7 +177,13 @@ object Itc:
    */
   case class Spectroscopy(
     acquisition: Zipper[Result],
-    science:     Zipper[Result]
+    science:     Zipper[Result],
+    // The GNIRS acquisition mode resolved by the ITC brightness classification.
+    // Only GNIRS long slit / IFU auto-resolution sets it; other instruments leave
+    // it None.  It pins the mode the sequence uses so it isn't re-derived (and
+    // possibly mis-classified) from the final exposure time.  See the two-pass
+    // acquisition ITC in ItcService.
+    gnirsAcqType: Option[GnirsAcquisitionType] = None
   ) extends Itc:
 
     override def dataType: Type =
@@ -190,7 +197,8 @@ object Itc:
       Eq.by: a =>
         (
           a.acquisition,
-          a.science
+          a.science,
+          a.gnirsAcqType
         )
 
   val spectroscopy: Prism[Itc, Spectroscopy] =
--- a/modules/schema/src/main/scala/lucuma/odb/data/Itc.scala
+++ b/modules/schema/src/main/scala/lucuma/odb/data/Itc.scala
@@ -174,15 +174,13 @@ object Itc:
   /**
    * Spectroscopy results for all instruments. Spectroscopy has separate
    * acquisition and science results.
+   * 
+   * GNIRS acquisition is particular in that it also must compute the acquisition
+   * type in this layer. See the two-pass acquisition ITC in ItcService.
    */
   case class Spectroscopy(
     acquisition: Zipper[Result],
     science:     Zipper[Result],
-    // The GNIRS acquisition mode resolved by the ITC brightness classification.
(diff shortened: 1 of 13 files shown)


modules/schema/src/main/scala/lucuma/odb/data/Itc.scala
```diff
@@ -176,7 +177,13 @@ object Itc:
    */
   case class Spectroscopy(
     acquisition: Zipper[Result],
-    science:     Zipper[Result]
+    science:     Zipper[Result],
+    // The GNIRS acquisition mode resolved by the ITC brightness classification.
```
cquiroz: I don't like very much these long comments in the param itself, should they be part of the class docs instead?

modules/schema/src/main/scala/lucuma/odb/json/itc.scala
```diff
@@ -103,12 +105,23 @@ trait ItcCodec:
...
+    Decoder[String].emap: s =>
+      Enumerated[GnirsAcquisitionType].fromTag(s).toRight(s"Invalid GnirsAcquisitionType: $s")
+
   given Decoder[Itc.Spectroscopy] =
     Decoder.instance: c =>
       for
         acquisition <- c.downField("acquisition").as[Zipper[Itc.Result]]
         science     <- c.downField("spectroscopyScience").as[Zipper[Itc.Result]]
-      yield Itc.Spectroscopy(acquisition, science)
+        // Absent in rows written before the two-pass acquisition ITC → None.
```
cquiroz: Does this refer to the stored itc results in the odb? could we insted just clean the itc stored values table and it should be reconstructed as needed?

modules/sequence/src/main/scala/lucuma/odb/sequence/data/ItcInput.scala
```diff
@@ -90,7 +90,11 @@ object ItcInput:
     science:     SpectroscopyParameters,
     targets:     NonEmptyList[TargetDefinition],
     blindOffset: Option[TargetDefinition],
-    signalToNoiseTargetId: Option[Target.Id]
+    signalToNoiseTargetId: Option[Target.Id],
+    // When set (GNIRS S/N mode with acquisition mode and filter both auto), the ITC
```
cquiroz: Same comment as above

modules/sequence/src/main/scala/lucuma/odb/sequence/gnirs/spectroscopy/Acquisition.scala
```diff
@@ -354,7 +355,11 @@ object Acquisition:
     namespace:     UUID,
     expander:      SmartGcalExpander[F, GnirsStaticConfig, GnirsDynamicConfig],
     config:        Config,
-    time:          Either[OdbError, IntegrationTime]
+    time:          Either[OdbError, IntegrationTime],
+    // The acquisition mode resolved by the ITC brightness classification, when the
```
cquiroz: and here

modules/service/src/main/scala/lucuma/odb/service/ItcService.scala
```diff
@@ -315,17 +327,48 @@ object ItcService {
...
-          case InstrumentMode.GnirsImaging(_, _, _, _, _, _, _) =>
-            go(lucuma.odb.sequence.gnirs.MinAcquisitionExposureTime,
-               lucuma.odb.sequence.gnirs.MaxAcquisitionExposureTime)
+            noType(go(input, flamingos2.MinAcquisitionExposureTime, flamingos2.MaxAcquisitionExposureTime))
+          case InstrumentMode.GnirsImaging(etm, filter, camera, readMode, wellDepth, coadds, port) =>
+            val min: TimeSpan = gnirs.MinAcquisitionExposureTime
+            val max: TimeSpan = gnirs.MaxAcquisitionExposureTime
+            etm match
+              case ExposureTimeMode.SignalToNoiseMode(userSN, at) if autoClassify =>
+                // Two-pass. The acquisition mode (Very Bright / Bright / Faint) is a
```
cquiroz, line 344: Very nice

cquiroz (approved):
Seems very good

modules/schema/src/main/scala/lucuma/odb/json/itc.scala
rpiaggio: Yes, I'm removing the comment.

PR #2698, opened 2026-07-08
SC-8978 SSO api endpoint to revoke a role

tpolecat:
This adds a `deleteRole` mutation to the SSO GraphQL scheme. It can be called only by Admin and Service users. The purpose is to allow us to revoke privileged (i.e., Staff and Admin) roles when people change jobs, etc.

When a role is deleted its active sessions are removed, which will cause the next JWT refresh to fail. The user will be kicked out and will need to log in again, at which point their role will revert to PI.

In the end this is a small change but I split out the GraphQL tests by root mutation/query like in the ODB, which makes the diff larger.


swalker2m (approved):
This looks good to me.  :+1:

PR #2701, opened 2026-07-08
Add minPercentTime to keck and subaru proposals

toddburnside:


modules/service/src/main/resources/db/migration/V1207__proposal_min_percent_exchange.sql
```diff
@@ -0,0 +1,82 @@
...
+-- minimum-percent-time rule.  The ch_proposal_type trigger references the
+-- function by name, so it does not need to be recreated.
+
+CREATE OR REPLACE FUNCTION t_proposal_type_checks()
+RETURNS TRIGGER AS $$
+DECLARE
+  is_gemini boolean           := (NEW.c_observatory = 'gemini');
+  st        e_science_subtype := NEW.c_science_subtype;
+BEGIN
+  -- A science subtype is present iff the proposal is a Gemini proposal.
```
cquiroz, line 15: I suppose all these errors should have a match as an `OdbError` right?

toddburnside: I suppose. But can't guarantee it. These are all existing errors from an old migration. This just changes one condition.

cquiroz (approved):
LGTM

swalker2m (approved):
This looks good to me.  Sorry I didn't notice it before.  :+1:

modules/service/src/main/resources/db/migration/V1207__proposal_min_percent_exchange.sql
swalker2m: The code isn't supposed to allow this to happen, so this just ensures that it doesn't due to a bug or someone doing a manual `UPDATE`.

Unrelated, but the `t_` prefix is supposed to mean "table" so it appears I screwed up calling it `t_proposal_type_checks`.

PR #2702, opened 2026-07-09
Slow query improvements

cquiroz:
A few niceties for slow queries, esentially we can override the time with an obs variable and we can now write the graphql query and sql query to a file for debugging. This is only really useful for local development


tpolecat (approved):
Looks useful 👍🏻

PR #2706, opened 2026-07-09
Fix flaky imaging guideEnvironment tests

toddburnside:
The random generation of offsets for group imaging invariants was changing the vignetting for each run, sometimes selecting a different position angle. They now use an explicit seed for the random generation in the tests.


cquiroz (approved):
Excelent find

PR #2707, opened 2026-07-09
Update f2 setup time for longslit

cquiroz:


modules/service/src/main/resources/db/migration/V1208__f2_longslit_setup_time.sql
```diff
@@ -0,0 +1,13 @@
...
+ WHERE c_tag = 'f2_longslit_setup';
+
+-- The setup time estimate feeds into obscalc, so previously-computed Flamingos 2
+-- results are stale and must be recalculated. 
+UPDATE t_obscalc c
+   SET c_obscalc_state = 'pending'
+  FROM t_observation o
+ WHERE c.c_observation_id = o.c_observation_id
+   AND o.c_instrument = 'Flamingos2'
+   AND c.c_workflow_state NOT IN ('inactive', 'ongoing', 'completed');
```
cquiroz: @swalker2m I'm not sure about the states to exclude. `completed` for sure should be excluded, what about `ongoing` and `completed`

toddburnside (approved):
Seems fine to me

modules/service/src/main/resources/db/migration/V1208__f2_longslit_setup_time.sql
toddburnside: Maybe you only need to do it for `ready` states?

swalker2m (approved):
Looks great

PR #2708, opened 2026-07-09
Shortcut 6776: calibrationRole in WhereObservation

swalker2m:
Adds `calibrationRole` to `WhereObservation`, as we do for targets.


toddburnside (approved):
LGTM

PR #2712, opened 2026-07-10
Add SSO middleware for Resource and move some middlewares to common

hugo-vrijswijk:


cquiroz (approved):
Looks great, thanks for moving more middlewares to the common module

cquiroz:
Could we publish the common middleware module, i could use it in observe and navigate server

hugo-vrijswijk:
> Could we publish the common middleware module, i could use it in observe and navigate server

Good idea, I've enabled publishing for it

PR #2713, opened 2026-07-10
Shortcut 9377: move time accounting

swalker2m:
Moves time accounting updates from execution event handling into the (asynchronous) Obscalc service.  A couple of salutary consequences:

* most importantly, event handling won't be slowed by a synchronous time accounting update
* but also, multiple consecutive updates coalesce in some cases

This turned out to be more ~complicated~ subtle than I expected.  In particular, we don't want to recompute time accounting for existing visits unless something explicitly triggers it: an execution event or a QA state change.  Otherwise, every time we change the time accounting algorithm all the existing time charges and careful manual corrections made by the staff would be compromised.  This PR adds a couple of time accounting timestamp fields to `t_visit` used to detect explicit updates and prevent that situation.

### DIFF
--- a/modules/obscalc/time-accounting-flow.md
+++ b/modules/obscalc/time-accounting-flow.md
@@ -14,7 +14,7 @@ when events arrive. An execution event only marks the affected visit dirty; the
 worker picks the observation up and recomputes the dirty visits. This keeps the
 event-ingestion path (from the "Observe" application) cheap.
 
-See [`obscalc-flow.md`](../obscalc/obscalc-flow.md) for the worker/daemon
+See [`obscalc-flow.md`](obscalc-flow.md) for the worker/daemon
 machinery this rides on.
 
 ## Where it lives
--- a/modules/service/src/main/scala/lucuma/odb/service/ExecutionEventService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/ExecutionEventService.scala
@@ -6,7 +6,6 @@ package lucuma.odb.service
 import cats.effect.Concurrent
 import cats.syntax.applicative.*
 import cats.syntax.applicativeError.*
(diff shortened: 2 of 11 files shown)


modules/obscalc/time-accounting-flow.md
```diff
@@ -0,0 +1,220 @@
+# Time Accounting Flow
```
cquiroz, line 1: Should this file be in `obscalc` module?

modules/service/src/main/scala/lucuma/odb/service/ExecutionEventService.scala
```diff
@@ -127,17 +126,9 @@ object ExecutionEventService:
             case _                        => ().pure
 
         ResultT(insert)
-          .flatMap: (eid, time, vid, wasInserted) =>
+          .flatMap: (eid, time, _, wasInserted) =>
```
cquiroz: so insert returns visit id but we don't need it anymore?

cquiroz (approved):
It is a bit hard to follow for me but in general LGTM

modules/service/src/main/scala/lucuma/odb/service/ExecutionEventService.scala
swalker2m: Good point, I'll fix it.

modules/obscalc/time-accounting-flow.md
swalker2m: It's debatable.  Most of the implementation is in the `TimeAccountingService` which is part of the `service` module. On the other hand, a lot of the doc is devoted to interactions around invoking it which is more of an `obscalc` thing.  I'll move it.  Perhaps we need a shared `doc` module somewhere for these things.

PR #2719, opened 2026-07-13
Provide an api for target visibilityr

cquiroz:
This is a PR endpoint for the sheduler it returns a list of observation and target ids that have change since a parameter timestamp in such a way that visibility should be recalculated.

There are no server-side visibility calculation. Database triggers stamp a single ` c_last_visibility_invalidation` timestamp on the observation or target whose visibility-relevant inputs changed, and the endpoint reads those columns directly.    

 ### Endpoint                                                                                                                          
                                                                                                                                       
 ```                                                                                                                                   
   GET /scheduler/visibility-changes?since=<iso8601-timestamp>                                                                         
 ``` 
### Response: TSV body, one id per line, matching the style of /scheduler/atoms:                                                        
```                                                                                                                                 
  OBSERVATION\t<gid>                                                                                                                
  TARGET\t<gid>                                                                                                                     
```      

It is as simple as it gets, perhaps we should consider returning pairs of obsids,targetids?


modules/service/src/main/resources/db/migration/V1211__visibility_invalidation.sql
```diff
@@ -0,0 +1,126 @@
...
+-------------------------------------------------------
+
+-- A change to any timing window stamps its observation.
+CREATE OR REPLACE FUNCTION visibility_timing_window_stamp()
+RETURNS TRIGGER AS $$
+BEGIN
+  UPDATE t_observation
+  SET c_last_visibility_invalidation = now()
+  WHERE c_observation_id = COALESCE(NEW.c_observation_id, OLD.c_observation_id);
+  RETURN COALESCE(NEW, OLD);
```
swalker2m, line 110: It seems like these functions are going to end up producing extra observation edit events, even though the observation isn't being edited again per se?  That could cause some churn for clients like explore and obscalc.  Perhaps we should consider dedicated last visibility invalidation tables for observations and targets?

modules/service/src/main/scala/lucuma/odb/graphql/SchedulerRoutes.scala
```diff
@@ -98,14 +103,30 @@ object SchedulerRoutes:
       row._3.stepCount.value.toString
     ).intercalate("\t")
 
+  private def visibilityTsv(c: VisibilityChange): String =
+    c match
+      case VisibilityChange.ForObservation(id) => s"OBSERVATION\t${Gid[Observation.Id].show(id)}"
+      case VisibilityChange.ForTarget(id)      => s"TARGET\t${Gid[Target.Id].show(id)}"
```
swalker2m: The GID itself identifies whether it references an observation or a target. It seems like we could get by with just listing the observation and target ids.

modules/service/src/main/resources/db/migration/V1211__visibility_invalidation.sql
cquiroz: That's a good point, didn't think about it. I wanted to avoid extra tables. Could we ignore that column from the triggers that produce `observationEdit`?

modules/service/src/main/scala/lucuma/odb/graphql/SchedulerRoutes.scala
cquiroz: right, that would be shorter, just
o-xxx
t-yyyy

modules/service/src/main/scala/lucuma/odb/service/VisibilityService.scala
```diff
@@ -0,0 +1,70 @@
...
+        FROM t_observation o
+        JOIN t_obscalc oc ON oc.c_observation_id = o.c_observation_id
+        WHERE oc.c_workflow_state IN ('ready', 'ongoing')
+          AND o.c_last_visibility_invalidation >= $core_timestamp
+        ORDER BY o.c_last_visibility_invalidation
+      """.query(observation_id)
+
+    val SelectTargets: Query[Timestamp, Target.Id] =
+      sql"""
+        SELECT DISTINCT t.c_target_id, t.c_last_visibility_invalidation
```
swalker2m: I'm curious why the timestamp is included in the selection.

cquiroz: you're right, this is a leftover of my original design where visibility was deemed unique.

cquiroz: Actually iis needed to sort the output, but that makes me wonder if we need to sort the output at all, scheduler doesn't probably need it. I'll drop that

modules/service/src/main/resources/db/migration/V1211__visibility_invalidation.sql
```diff
@@ -0,0 +1,171 @@
...
+BEGIN
+  INSERT INTO t_target_visibility (c_target_id) VALUES (NEW.c_target_id);
+  RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER visibility_target_seed_trigger
+  AFTER INSERT ON t_target
+  FOR EACH ROW
+  EXECUTE FUNCTION visibility_target_seed();
```
swalker2m, line 59: If the scheduler is already calculating visibility for target create events, I suppose they could drop that and instead rely on the visibility change query. I mean, new targets will be returned by the visibility change query as well so either they rely on it or else they'll end up redoing the visibility calculation.

modules/service/src/test/scala/lucuma/odb/graphql/scheduler/visibilityChanges.scala
```diff
@@ -0,0 +1,343 @@
...
+class visibilityChanges extends SchedulerRoutesSuite with ExecutionTestSupportForGmos:
+
+  private val Epoch:  Instant = Instant.EPOCH
+  private val Future: Instant = Instant.parse("2999-01-01T00:00:00Z")
+
+  private def visibilityRequest(user: User, since: Instant): IO[Request[IO]] =
+    headers(user).map: hs =>
+      Request[IO](
+        method  = Method.GET,
+        uri     = uri"scheduler/visibility-changes".withQueryParam("since", since.toString),
```
swalker2m, line 35: How does the scheduler set the `since` value?  It seems like we somehow need to return the last known visibility change timestamp along with the observations and targets?

modules/service/src/main/scala/lucuma/odb/service/VisibilityService.scala
```diff
@@ -0,0 +1,74 @@
...
+    val SelectObservations: Query[Timestamp, Observation.Id] =
+      sql"""
+        SELECT ov.c_observation_id
+        FROM t_observation_visibility ov
+        JOIN t_observation o ON o.c_observation_id = ov.c_observation_id
+        JOIN t_obscalc oc ON oc.c_observation_id = ov.c_observation_id
+        WHERE oc.c_workflow_state IN ('ready', 'ongoing')
+          AND o.c_existence = 'present'
+          AND ov.c_last_visibility_invalidation >= $core_timestamp
+      """.query(observation_id)
```
swalker2m: Let's say something changes about the observation that impacts observability and makes the observation `ready` (it was sitting at `defined` before the edit).  It is stamped at a given time and obscalc eventually picks it up and starts working on it.  The scheduler makes the query at a given time and because obscalc isn't done yet, misses the observation in this query.  Obscalc finishes its work and the workflow state becomes `ready`.  The scheduler updates its last checked timestamp to whatever the latest invalidation time was and uses it for the next query.  It seems like it will have missed an update?

swalker2m: Maybe that is caught via a normal subscription on edit events.  I'm just trying to think through the asynchronous nature of workflow state updates.

swalker2m: Trying to think of a way make this more reliable.  The problem is that workflow state updates are asynchronous but the last invalidation timestamps are set by synchronous triggers.  It seems like obscalc would need to provide the timestamp to use for the comparison to the provided value from the scheduler and it would need to be set when it settles to a _calculation state_ of `ready`.  But that complicates obscalc.  What if we just don't filter on `ready` and `ongoing` at all and let the scheduler intersect the results of this query with its collection of schedule-able  observations?

cquiroz: would this be a problem in the year 2050 when we have several thousands observations? I guess `since` wil filter them out

modules/service/src/test/scala/lucuma/odb/graphql/scheduler/visibilityChanges.scala
cquiroz: maybe a question for @stroncod . My understanding is that it keeps it in a different db
As discussed on the chat I'll include the timestamp values on the returned list so the scheduler can do a max(timestamp) using odb time values.

swalker2m (approved):
This seems good to me. :+1:  I think we could filter on `ready` and `ongoing` but it would require some `obscalc` updates to manage the timestamps.  If we need to do go that route, I can look into it.  But hopefully the scheduler is already tracking its own idea of ready observations and can just intersect them with the query results.

PR #2720, opened 2026-07-13
Setting selected sn target not working when

cquiroz:
While trying to support this feature on explore I got an error when setting the SN target if one already exists, in the form of
```
   Duplicate key value violates unique constraint "i_asterism_single_sn_target".                                                       
   Key (c_program_id, c_observation_id)=(p-11b9, o-4cee) already exists. 
```

The fx was to split the operation into two statements in the same transaction, reusing existing fragments:                                          
                                                                                                                                       
 1. `clearSignalToNoiseTargets` — set all SN flags false for the observations.                                                           
 2. `applySignalToNoiseFlags` — set exactly the chosen target true.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/AsterismService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/AsterismService.scala
@@ -220,8 +220,15 @@ object AsterismService {
                 val missing = observationIds.toList.filterNot(present.toSet)
                 missing match
                   case Nil =>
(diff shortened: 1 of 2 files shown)


swalker2m (approved):
D'oh, thanks for fixing this.  I guess the test case got lucky and cleared before setting.  :+1:

cquiroz:
> D'oh, thanks for fixing this. I guess the test case got lucky and cleared before setting. 👍

That's right, I only noted when doing a flip of target id doing the explore ui

PR #2721, opened 2026-07-14
Allow introspection without auth for SSO and Resource

hugo-vrijswijk:


modules/common-middleware/src/main/scala/lucuma/common/middleware/IntrospectionMapping.scala
```diff
@@ -0,0 +1,33 @@
...
+import grackle.QueryCompiler.SelectElaborator
+import grackle.Schema
+
+/**
+ * A minimal, read-only Grackle mapping that supports GraphQL schema introspection and nothing else.
+ * It exposes the schema so that clients can fetch it without credentials, but rejects other queries
+ */
+object IntrospectionMapping:
+
+  def apply[F[_]: MonadThrow](loadedSchema: Schema): Mapping[F] =
```
cquiroz, line 18: `MonadThrow as M`

modules/service/src/main/scala/lucuma/odb/graphql/OdbMapping.scala
```diff
@@ -781,41 +782,11 @@ object OdbMapping {
         }
 
   /**
-    * A minimal read-only mapping that only knows how to return introspection metadata. Other queries will
-    * fail with errors.
+    * The full ODB schema, including the dynamically-loaded enums. This is the schema exposed for
```
cquiroz, line 785: do we still have dynamic enums?

cquiroz (approved):
LGTM, great that we have a shared middleware mapping for this

modules/service/src/main/scala/lucuma/odb/graphql/OdbMapping.scala
hugo-vrijswijk: Yes

modules/common-middleware/src/main/scala/lucuma/common/middleware/IntrospectionMapping.scala
hugo-vrijswijk: Not sure how to make this work with the override. A class with `implicit val M` would work, but since this is an anonymous class instantiated it doesn't. Feel free to push a change if you know the magic incantation

hugo-vrijswijk: I think `MonadThrow as M` defines as `private[this] given val M: cats.MonadThrow[F]`, while the override is `implicit val M: MonadThrow[F]`

cquiroz: You'r right I didn't notice that

modules/service/src/main/scala/lucuma/odb/graphql/OdbMapping.scala
toddburnside: We still have several that should be removed, but none are currently used by anyone.

PR #2722, opened 2026-07-14
Shortcut 9429: execution event record deadlock

swalker2m:
This is a fairly frightening, though plausible, explanation and fix (from Claude) for the deadlock encountered by Andy while testing with observe.


modules/service/src/main/resources/db/migration/V1212__async_time_accounting_deadlock.sql
```diff
@@ -0,0 +1,35 @@
...
+-- adding triggers).
+CREATE OR REPLACE PROCEDURE invalidate_visit_time_accounting(
+  visit_id       d_visit_id,
+  observation_id d_observation_id
+) LANGUAGE plpgsql AS $$
+BEGIN
+  -- Lock the observation first.  This is the same lock
+  -- update_execution_information_for_step_event takes, and it precedes the
+  -- t_obscalc lock that invalidate_obscalc acquires, so the order can't invert.
+  PERFORM 1 FROM t_observation WHERE c_observation_id = observation_id FOR UPDATE;
```
cquiroz, line 31: Do any parallel transaction on the same row waits or fails?

cquiroz (approved):
LGTM

modules/service/src/main/resources/db/migration/V1212__async_time_accounting_deadlock.sql
swalker2m: Yes, if it needs to update that observation row it has to wait if this code is executing.

PR #2723, opened 2026-07-15
Decouple visitor science fov and ags diameter

cquiroz:
Visitor will now accept two diameters, one for science and one for ags. This also allows us to treat maroonx like any another visitor.


swalker2m (approved):
Looks good.

PR #2726, opened 2026-07-16
Exchange observations default to Defined, not Completed

toddburnside:
Exchange observations (Keck/Subaru) were automatically showing a Completed workflow status because the digest calculator defaults their execution state to Completed, and execution state overrides everything in the workflow computation. This also masked underlying validation errors.

Handle exchange like visitor observations: ignore the digest's default Completed in getCalculatedWorkflow and exempt exchange from the generator, ITC, and configuration-request validators so the state lands on Defined. Add Completed (declared-complete) to the Defined/Ready transition set and allow reverting Completed -> Defined; Ready is kept for now, Ongoing is naturally excluded. The PI can mark Completed since access control only requires write access plus a valid transition.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/ObservationWorkflowService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/ObservationWorkflowService.scala
@@ -508,17 +508,13 @@ object ObservationWorkflowService {
             case Undefined  => List(Inactive)
             case Unapproved => List(Inactive)
             case Defined    =>
+              // Exchange observations run at Keck/Subaru, not Gemini; they have no
+              // Ready/Ongoing/Completed lifecycle, so Inactive is the only transition.
               List(Inactive) ++
-                Option.when((!info.isOpportunity) && (info.isAccepted || !info.tpe.hasProposal))(Ready) ++
-                Option.when(info.isExchange)(Completed)
-            case Ready      =>
-              List(Inactive, validationStatus) ++
-                Option.when(canUpdateExecutionState)(Ongoing) ++
-                Option.when(info.isExchange)(Completed)
+                Option.when((!info.isExchange) && (!info.isOpportunity) && (info.isAccepted || !info.tpe.hasProposal))(Ready)
+            case Ready      => List(Inactive, validationStatus) ++ Option.when(canUpdateExecutionState)(Ongoing)
             case Ongoing    => List(Completed) ++ Option.when(canUpdateExecutionState)(Ready)
-            case Completed  =>
-              if info.isExchange then List(Defined)            // exchange: allow revert (un-complete)
-              else if info.isDeclaredComplete then List(Ongoing) else Nil
+            case Completed  => if info.isDeclaredComplete then List(Ongoing) else Nil
 
         (state, allowedTransitions)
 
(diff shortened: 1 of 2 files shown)


modules/service/src/main/scala/lucuma/odb/service/ObservationWorkflowService.scala
```diff
@@ -504,10 +507,18 @@ object ObservationWorkflowService {
             case Inactive   => List(executionState.getOrElse(validationStatus))
             case Undefined  => List(Inactive)
             case Unapproved => List(Inactive)
-            case Defined    => List(Inactive) ++ Option.when((!info.isOpportunity) && (info.isAccepted || !info.tpe.hasProposal))(Ready)
-            case Ready      => List(Inactive, validationStatus) ++ Option.when(canUpdateExecutionState)(Ongoing)
+            case Defined    =>
+              List(Inactive) ++
+                Option.when((!info.isOpportunity) && (info.isAccepted || !info.tpe.hasProposal))(Ready) ++
```
cquiroz: Maybe a little comment for future maintainers?

cquiroz:
WDYT about adding a document with a diagram documenting all the possible workflow transitions?

cquiroz (approved):
LGTM

modules/service/src/main/scala/lucuma/odb/service/ObservationWorkflowService.scala
```diff
@@ -504,10 +507,18 @@ object ObservationWorkflowService {
...
+                Option.when(info.isExchange)(Completed)
```
swalker2m: I wonder if even this is necessary.  The idea is that once accepted the proposal is translated into whatever observing system they have at Keck or Subaru.  Maybe exchange proposals should stop at `Defined`.

modules/service/src/main/scala/lucuma/odb/service/ObservationWorkflowService.scala
```diff
@@ -777,6 +788,21 @@ object ObservationWorkflowService {
                       updateDeclaredState(oid, Some(Completed))
                         .as(Result(w.copy(state = state)))
 
+                    // Exchange observations: declare complete directly from Defined/Ready
+                    // (there is no Ready/Ongoing lifecycle at Gemini), and allow reverting.
```
swalker2m: It's a little confusing because the logic only applies to exchange but nothing seems to check that they are exchange observations?

swalker2m (approved):
I have a policy question but I don't think it should hold up merging.  We can always adjust later if necessary.

modules/service/src/main/scala/lucuma/odb/service/ObservationWorkflowService.scala
toddburnside: That was my first thought. But, then I wondered if they would want one more step before entering into Keck or Subaru by setting them to Ready. Then I wondered if somebody would want to mark them as Completed for some kind of reporting. So, I put them in. They are easy to remove.

modules/service/src/main/scala/lucuma/odb/service/ObservationWorkflowService.scala
toddburnside: Huh. You're right. I'll try to figure that out and put a comment or fix it.

toddburnside: I `think` it is because the valid transitions are checked earlier, and these are only valid for exchange. But, maybe it would be simplest to just stop at Defined for exchange programs?

swalker2m: You're probably right that they want some way to mark which observations are accepted.  I guess we need policy clarification at some point at the risk of generating more complexity.

toddburnside: I discussed with Andy and he said we can just stop at Defined.

modules/service/src/main/scala/lucuma/odb/service/ObservationWorkflowService.scala
toddburnside: I discussed with Andy and he said we can just stop at Defined.

toddburnside:
> WDYT about adding a document with a diagram documenting all the possible workflow transitions?

I think we should ask claude to make one after Rob is done with his upcoming changes. It could be a mermaid diagram in the ODB reame.

PR #2732, opened 2026-07-17
Preparation for MOS

cquiroz:
This is just enough to get the lucuma core update, there is no MOS implementation yet


toddburnside (approved):
LGTM

PR #2733, opened 2026-07-17
Shortcut 9429: deadlock on FOR KEY SHARE

swalker2m:
This is another attempt by Claude to solve the deadlock issue.  The description is plausible, except for the inference that there are simultaneous events flowing in from Observe.  I suspect no?  At any rate, presumably the prescription is the same regardless what other thing is also trying to grab the observation `FOR UPDATE` lock.

I didn't expect so much fallout from moving the time accounting calculation from the event insert.

### DIFF
--- a/modules/service/src/main/resources/db/migration/V1217__execution_event_lock_upgrade_deadlock.sql
+++ b/modules/service/src/main/resources/db/migration/V1217__execution_event_lock_upgrade_deadlock.sql
@@ -1,0 +1,47 @@
+-- Fix the remaining execution-event deadlock: a lock UPGRADE on t_observation.
+--
+-- Inserting a row into t_execution_event enforces its foreign key to
+-- t_observation, and to do so PostgreSQL takes a FOR KEY SHARE lock on the
+-- referenced t_observation row.  Crucially, that FK lock is taken as part of the
+-- INSERT itself -- BEFORE any AFTER-INSERT trigger runs.  The AFTER triggers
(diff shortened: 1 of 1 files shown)


modules/service/src/main/resources/db/migration/V1217__execution_event_lock_upgrade_deadlock.sql
```diff
@@ -0,0 +1,47 @@
...
+-- Inserting a row into t_execution_event enforces its foreign key to
+-- t_observation, and to do so PostgreSQL takes a FOR KEY SHARE lock on the
+-- referenced t_observation row.  Crucially, that FK lock is taken as part of the
+-- INSERT itself -- BEFORE any AFTER-INSERT trigger runs.  The AFTER triggers
+-- (event_time_accounting_invalidate -> invalidate_visit_time_accounting, and
+-- update_execution_information_for_step_event) then take FOR UPDATE on the SAME
+-- t_observation row.  So every event insert performs a lock UPGRADE on
+-- t_observation:  FOR KEY SHARE  ->  FOR UPDATE.
+--
+-- FOR KEY SHARE is self-compatible, so two concurrent event inserts that
```
cquiroz, line 12: Not sure if observe would be sending events concurrently, tasks are mostly sequential except when configuring independent parts of the hardware. Maybe it is not waiting for the end of one to start the next step?

toddburnside (approved):
I guess we should give it a try. Maybe claude is right this time.

PR #2735, opened 2026-07-19
Allow configuring the guide state of F2 offsets

cquiroz:
We want to allow changing the guide state for F2 which implies switching to `TelescopeConfigs`. Additionally presets are defined for F2 though not stored.

Depends on https://github.com/gemini-hlsw/lucuma-core/pull/1510

This PR is part of the unification task at [sc-9408] and now F2 and gnirs use the same logic, pending is IGRINS2 and GMOS


toddburnside (approved):
LGTM pending lucuma-core update

PR #2737, opened 2026-07-20
Spectrophotometric calibration S/N wavelength not synced

cquiroz:
When a science observation's S/N wavelength (the λ for S/N) changed, the matching specphot calibration's S/N λ was not updated when the science observation used a non-default GMOS ROI (e.g. FullFrame). The calibration's S/N λ stayed  frozen at its creation-time value, so ITC and sequence generation used a stale wavelength for the calibration.                        

 ### Problem   

At creation, a specphot calibration's ROI is normalized to CentralSpectrum (SpecphotoGmosLS matcher). So the calibration's stored config always carries ROI = CentralSpectrum, even when the science obs that triggered it used FullFrame.                              
                                                                                                                                       
Aditionall `updatePropsAt` only updated `scienceRequirements.exposureTimeMode`. But  `ObservationService.updateObservations` only propagates a requirement edit into the observing mode when an explicit observingMode edit  is supplied. So the mode's own science-role ETM was never updated and ITC uses that ETM for calculations.
                                                                                                                                       
 ### Solution                                                                                                                          
                                                                                                                                       
 In PerProgramPerConfigCalibrationsService:                                                                                            
                                                                                                                                       
 - `calObsProps` now groups by `matcherFor(config, role).normalize(config`) and produces `Map[CalibrationRole, Map[CalibrationConfigSubset, CalObsProps]]`. 
 - `updatePropsAt` now performs two writes: the existing requirement-row ETM sync and a new mode-level science ETM rewrite                                         
 -  Because the mode rewrite now edits the actual executing mode, this must never touch a calibration that has started or completed                                       
 - A migration invalidates obscalc for one science observation per program that has a non-executed specphot calibrations to recalculate the S/N wavelength.


toddburnside (approved):
LGTM

PR #2739, opened 2026-07-20
Fix GNIRS acquisition type leaking user acquisition ExposureTimeMode

toddburnside:
The GNIRS acquisition brightness classification (Very Bright / Bright / Faint) must be determined by a fixed ITC classification pass (S/N=10), independent of the user's acquisition ExposureTimeMode. Previously that pass only ran in SignalToNoiseMode; in TimeAndCountMode it was skipped and the type was re-derived from the user's own acquisition exposure x coadds, so two otherwise-identical observations classified differently and picked different acquisition filters.

Run the two-pass classification for both ETM types. ExposureTimeMode.at is present on both SignalToNoiseMode and TimeAndCountMode, so the classification pass reads etm.at uniformly - no wavelength threading needed. The skip-pass-2 optimization now only short-circuits when the user's ETM is itself the classification S/N request.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/GeneratorParamsService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/GeneratorParamsService.scala
@@ -501,16 +501,14 @@ object GeneratorParamsService {
                 coadds            = gn.coadds
               )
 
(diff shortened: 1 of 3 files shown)


cquiroz (approved):
LGTM

PR #2743, opened 2026-07-21
Make some calibrations steps optional

cquiroz:
Some configurations have a flat but no arc in the smart GCal tables, the arc lamps don't span every wavelength. Sequence generation required both a flat and an arc, so these configs failed with missing Smart GCAL mapping.

This was initially reported for gnirs but also f2 suffers the same fate

After discussing with science it is correct not to have arc steps and only flats, thus they now become optional.


toddburnside (approved):
LGTM

PR #2744, opened 2026-07-21
Dedup observation ids in getAsterisms to avoid parameter-limit overflow

toddburnside:
The configurationRequests/applicableObservations query batches asterism lookups across all observation cursors. Because the same observation can appear many times, getAsterisms could build an IN (...) list exceeding Postgres's 32767 parameter limit, throwing TooManyParametersException.

Deduplicate the ids before building the IN list (as the sibling getSiteAndExplicitBaseCoordinates already does). The result Map is keyed by observation id, so this is behavior-preserving.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/AsterismService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/AsterismService.scala
@@ -270,7 +270,12 @@ object AsterismService {
       override def getAsterisms(
         oids: List[Observation.Id]
       )(using SuperUserAccess): F[Map[Observation.Id, List[(Target.Id, Target)]]] =
-        NonEmptyList.fromList(oids) match
+        // Dedup the ids before building the `IN (...)` list. Callers (e.g. the
+        // configurationRequests/applicableObservations effect batching) may pass the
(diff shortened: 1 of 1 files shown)


modules/service/src/main/scala/lucuma/odb/service/AsterismService.scala
```diff
@@ -270,7 +270,12 @@ object AsterismService {
       override def getAsterisms(
         oids: List[Observation.Id]
       )(using SuperUserAccess): F[Map[Observation.Id, List[(Target.Id, Target)]]] =
-        NonEmptyList.fromList(oids) match
+        // Dedup the ids before building the `IN (...)` list. Callers (e.g. the
```
cquiroz, line 273: Cool, it this something happening in other places?

toddburnside: Possibly. But this is one of the few places using the recursive graphql calls. I'm actually working on another PR that would eliminate getting the duplicate observations in the first place. That is an efficiency change that seemed out of scope for this PR so I'm doing it separately.

cquiroz (approved):
LGTM

PR #2745, opened 2026-07-21
Make igrins2 use the same offsets model as f2

cquiroz:
This is a continuation of #2735 making igrins2 longslit offsets behave like F2 and IGRINS2:

- presets don't get stored on the odb
- We use full telescope configs allowing us to set/unset guiding


toddburnside (approved):
It's a lot to take in, especially the migration, but it seems fine to me.

cquiroz:
Yes the migration is massive I had to review it a few times. Note the changes have been tested with explore on the companion PR

PR #2746, opened 2026-07-21
Fetch program observations once per program in applicableObservations

toddburnside:
queryRequestsAndObservations previously ran a single query that nested program.observations.matches under each configuration request. Since all requests in a program share the same observation set, this re-fetched and re-processed the same observations once per request (O(requests × observations)).

Split it into two sequential queries joined in Scala: SelectRequests fetches each request with its program id, and SelectProgramObservations fetches each distinct program's observations exactly once. The return shape is unchanged, so selectObservations and downstream callers are untouched. The observation sub-selection and Option[Configuration] decoding are preserved verbatim.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/ConfigurationService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/ConfigurationService.scala
@@ -242,11 +242,40 @@ object ConfigurationService {
       if rids.isEmpty then
         ResultT.pure(List.empty)
       else
+        // Fetch the requests (each with its program id), then fetch each distinct
+        // program's observations exactly once, rather than re-fetching the entire
+        // observation list once per request in the same program. The old single
+        // nested query returned O(requests × observations) rows, which for a program
+        // with many requests could also blow the Postgres parameter limit when the
+        // asterism lookups for `configuration.target` were batched.
+        for
+          requests <- queryRequests(rids)
+          obsByPid <- queryProgramObservations(requests.map(_._2).distinct)
+        yield requests.map: (req, pid) =>
+          (req, obsByPid.getOrElse(pid, Nil))
+
+    private def queryRequests(
+      rids: List[ConfigurationRequest.Id]
+    ): ResultT[F, List[(ConfigurationRequest, Program.Id)]] =
+      ResultT:
+        services.runGraphQLQuery(Queries.SelectRequests(rids)).map: r =>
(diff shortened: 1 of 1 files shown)


modules/service/src/main/scala/lucuma/odb/service/ConfigurationService.scala
```diff
@@ -241,12 +241,41 @@ object ConfigurationService {
...
+      else
+        // Fetch the requests (each with its program id), then fetch each distinct
+        // program's observations exactly once, rather than re-fetching the entire
+        // observation list once per request in the same program. The old single
+        // nested query returned O(requests × observations) rows, which for a program
+        // with many requests could also blow the Postgres parameter limit when the
+        // asterism lookups for `configuration.target` were batched.
+        for
+          requests <- queryRequests(rids)
+          obsByPid <- queryProgramObservations(requests.map(_._2).distinct)
```
cquiroz, line 253: is this a `n+1` query? or do we read them all at once?

modules/service/src/main/scala/lucuma/odb/service/ConfigurationService.scala
```diff
@@ -736,58 +759,102 @@ object ConfigurationService {
...
+          pid <- hc.downField("id").as[Program.Id]
+          obs <- hc.downFields("observations", "matches").as(using Decoder.decodeList(using da))
+        yield (pid, obs.flatten)
+
+      given Decoder[Response] = hc =>
+        hc.downFields("programs", "matches")
+          .as(using Decoder.decodeList(using dp))
+          .map(_.toMap)
+
+      def apply(pids: List[Program.Id]) =
```
cquiroz, line 795: why do we query over graphql instead of directly to the db?

modules/service/src/main/scala/lucuma/odb/service/ConfigurationService.scala
toddburnside: We read them all at once via the graphql recursive query.

modules/service/src/main/scala/lucuma/odb/service/ConfigurationService.scala
toddburnside: Rob did that because it makes use of the effect fields instead of having to reproduce the effects here.

cquiroz (approved):
LGTM

PR #2747, opened 2026-07-21
Add top-level result mapping for configurationRequests query

toddburnside:
The top-level configurationRequests query ignored LIMIT/OFFSET because it had no top-level Grackle result mapping. With only the nested, program-scoped mapping registered, the query root fell back to it; its program join and hidden parent-key column distorted the row set and broke the LIMIT+1 / Take / hasMore trimming. Adding a WHERE scoped it to a single program group, which masked the bug.

Register topLevelSelectResultMappingAtPath(QueryType / "configurationRequests") alongside the existing nested mapping, mirroring observations and datasets. Add a regression test that exercises the top-level query with LIMIT/OFFSET and no WHERE.


cquiroz (approved):
LGTM

PR #2749, opened 2026-07-22
Return multislit entries on the phase0 table

cquiroz:
ODB was filtering out multi slot entries on the list of spectroscopy options
This PR lifts that to support MOS selection for gmos


modules/service/src/main/resources/db/migration/V1224__gmos_multislit_null_fpu.sql
```diff
@@ -0,0 +1,6 @@
+-- GMOS MOS spectroscopy options are loaded as separate rows whose focal-plane unit
+-- is a custom mask rather than a builtin aperture.
+-- The builtin FPU column therefore becomes nullable:
+
+ALTER TABLE t_spectroscopy_config_option_gmos_north ALTER COLUMN c_fpu DROP NOT NULL;
+ALTER TABLE t_spectroscopy_config_option_gmos_south ALTER COLUMN c_fpu DROP NOT NULL;
```
toddburnside, line 6: Do we need to prevent it being null when it shouldn't be? For other modesf/

toddburnside (approved):
LGTM with one question.

modules/service/src/main/resources/db/migration/V1224__gmos_multislit_null_fpu.sql
cquiroz: it is tricky, i struggled a bit with this one
i think we could add a check that null is only allowed for multi-slits, wdyt?

toddburnside: It seems we commonly do that sort of thing. And, after all, we don't have nearly enough triggers. 😆

PR #2751, opened 2026-07-22
Shortcut 9415: freeze ITC results

swalker2m:
This turned into a bigger update than I anticipated. The motivation:

* We cache ITC results in the appropriately named `t_itc_result` table and reuse them when a hash of the ITC parameters matches.
* This cache is deleted wholesale whenever the ITC code is updated.
* When an observation's first visit is recorded, we materialize its sequence and faithfully use it thereafter, even if the parameters feeding sequence generation are later changed.
* Yet the ITC results are still looked up — and must be present — whenever anything requires the sequence.
 
The idea, then, is to freeze ITC results so that they stay consistent with the materialized sequence, no longer depend on the ITC version remaining stable, and don't even attempt to call the ITC during execution. What made this bigger than expected is that for acquisition we _do_ need to honor any changes made to the acquisition exposure time mode. For that reason, science and acquisition have been split and are now managed independently.


toddburnside (approved):
It's a lot to absorb, but it seems OK to me.

cquiroz:
The rationale make sense, are we having a "freeze" column in the table?

modules/service/src/main/resources/db/migration/V1225__freeze_itc_result.sql
```diff
@@ -0,0 +1,75 @@
...
+--     execution start, but are additionally re-derived by the resetAcquisition
+--     mutation (the only thing that consults the ITC for acquisition once
+--     execution has begun).
+--
+-- Both parts share the single c_hash (md5 of the full ITC input) for
+-- pre-execution cache validity and the single c_is_frozen flag (both freeze
+-- together at execution start).  t_itc_result is a disposable cache and nothing
+-- released depends on its contents, so we simply wipe it and restructure the
+-- result columns rather than migrate the old fused blob.
+TRUNCATE TABLE t_itc_result;
```
cquiroz, line 26: Would we in the future not truncate but delete non frozen rows?

modules/service/src/main/resources/db/migration/V1225__freeze_itc_result.sql
```diff
@@ -0,0 +1,75 @@
...
+-- On an ITC version change, wipe only the (unfrozen) cache rows and keep frozen
+-- results belonging to executing/executed observations.  Previously this was a
+-- wholesale TRUNCATE.  The obscalc reset below is unchanged (it already skips
+-- ongoing/completed observations).
+CREATE OR REPLACE FUNCTION itc_version_update()
+  RETURNS trigger AS $$
+BEGIN
+  NEW.c_last_update = NOW();
+  IF (OLD.c_version IS DISTINCT FROM NEW.c_version OR OLD.c_data IS DISTINCT FROM NEW.c_data) THEN
+    DELETE FROM t_itc_result WHERE NOT c_is_frozen;
```
cquiroz, line 57: 👍

modules/service/src/main/scala/lucuma/odb/service/ItcService.scala
```diff
@@ -122,6 +127,47 @@ sealed trait ItcService[F[_]] {
...
+   * Freezes the ITC result for an observation, making it durable and
+   * authoritative.  A frozen result is exempt from the wholesale wipe on an ITC
+   * version bump and is returned by all lookups regardless of the current input
+   * hash, so that an executing observation no longer depends on the live ITC
+   * service or on the disposable cache.  Intended to be called when execution
+   * begins (the sequence is materialized), passing the same ITC result that fed
+   * sequence generation.  Idempotent: the first frozen value wins and is never
+   * overwritten by a later call.
+   */
+  def freeze(
```
cquiroz, line 140: Does this freeze acq and science?

modules/schema/src/main/scala/lucuma/odb/data/Itc.scala
```diff
@@ -26,37 +25,75 @@ import monocle.Prism
...
- * ITC result data.  Because the results differ depending upon whether it is
- * imaging or spectroscopy, and which particular instrument is in use, there
- * are specific types for each case.
+ * A single ITC result for one target: the integration time (exposure time and
+ * count) and, when available, the achieved signal-to-noise.  Shared by both the
+ * science ([[ItcScience]]) and acquisition ([[ItcAcquisition]]) results.
+ * Corresponds to `ItcResult` in the GraphQL schema.
  */
-sealed trait Itc:
+case class ItcResult(
```
cquiroz, line 33: Are we using these classes in explore? I"d check before merging

modules/service/src/main/resources/db/migration/V1225__freeze_itc_result.sql
swalker2m: Right, exactly.  Until now it is has just been a disposable cache but now we will be saving the results for executing / executed observations.

modules/service/src/main/scala/lucuma/odb/service/ItcService.scala
swalker2m: Yes, with the caveat that a `resetAcquisition` will compare the hash value and will consult with the ITC if necessary.

modules/schema/src/main/scala/lucuma/odb/data/Itc.scala
swalker2m: The Scala classes are internal to the ODB and the GraphQL schema doesn't change.

cquiroz (approved):
LGTM

modules/schema/src/main/scala/lucuma/odb/data/Itc.scala
swalker2m: Well I make that claim about the Scala classes, but they are in the `schema` module.  Are they actually used elsewhere?

toddburnside: I `think` explore only uses the old `Itc.Type` to help decode SignalToNoiseAt. It seems that just switching that to `ItcScience.Type` would be sufficient.

PR #2752, opened 2026-07-22
WIP: SC-9059 sso api endpoint to query users

tpolecat:
This updates SSO to allow for the full paged query setup used by the ODB, and moves a few things to the shared `odb-schemas` module to allow sharing. It also updates the test setup to use MUnit rather than Weaver, and adds TestContainers to remove the necessity of a running Docker container for testing. I expect this to fix the flaky tests (`flaky()` is stubbed out at the moment).

Oh also it adds the requested `users` query.


modules/sso-service/src/test/scala/FlakyTests.scala
```diff
@@ -4,28 +4,8 @@
...
-              flaky(attempts-1)(x)  
-            }  
-          }  
-        )
-      )
-    }
-  }
-}
\ No newline at end of file
+  def flaky()(x: IO[Unit]): IO[Unit] =
```
hugo-vrijswijk, line 10: You can tag flaky tests in MUnit with `test("testName".flaky) {` which the method here was maybe replicating for Weaver? https://scalameta.org/munit/docs/tests.html#tag-flaky-tests

modules/sso-service/src/test/scala/SsoSuite.scala
```diff
@@ -3,14 +3,28 @@
...
 
-trait SsoSuite extends SimpleIOSuite {
+import simulator.SsoSimulator
+
+trait SsoSuite extends CatsEffectSuite with SsoSimulator {
 
   implicit val logger: Logger[IO] =
     Slf4jLogger.getLogger
 
+  def assertEq[A: Eq](obtained: A, expected: A, clue: => Any = ()): Unit =
```
hugo-vrijswijk: These should probably have:

```suggestion
  def assertEq[A: Eq](obtained: A, expected: A, clue: => Any = ())(using munit.Location): Unit =
```

For better error location reporting

modules/sso-service/src/test/scala/FlakyTests.scala
tpolecat: Yes, I think that was the case. I'm hoping the changes here will make the tests non-flaky so I can remove it entirely. We'll see.

modules/sso-service/src/test/scala/SsoSuite.scala
tpolecat: Ah good thinking. I will add that.

PR #2754, opened 2026-07-22
Shortcut 9430: CAL and ENG time accounting categories

swalker2m:
Adds CAL and ENG time accounting categories to the database.


toddburnside (approved):
LGTM after renumbering the migration

PR #2759, opened 2026-07-23
Add tables to store goa matches and an

cquiroz:
Sorry about the size of this PR, it maybe hard to review.

This is the first step towards doing observation goa duplication checks. It creates tables to get possible duplicates which are populated using GOA searches. Searches are different per observation as it takes into account the coordinates, the instrument family and it gets the radius directly from the mode geometry.

At this stage queries are requested manually via a graphql mutation but future PRs may watn to automate the process via an obscalc-like system. 

A new field in observation can be used to query the results like

```graphql
query archiveDuplication {
  observation(observationId: "o-57e0") {
    archiveDuplication {
      state
      lastCheckedAt
      matches {
        coordinates {
          ra {
            hms
          }
          dec {
            dms
          }
        }
        observationId
      }
    }
  }
}
```

Care is taken not to allow updates after a proposal has been submitted.


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -3527,6 +3527,16 @@ type Mutation {
     input: ReplaceGnirsSequenceInput!
   ): ReplaceGnirsSequenceResult!
 
+  """
+  Re-runs the Archive Duplication Search for an observation, replacing its
+  stored result with what the Gemini Observatory Archive holds now.
+
+  Rejected once the observation's proposal has been submitted.
+  """
```
swalker2m, line 3564: Who is allowed to do this?

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -13714,6 +13744,192 @@ type Observation {
   """
   workflow: CalculatedObservationWorkflow
 
+  """
+  Result of the Archive Duplication Search: what the Gemini Observatory Archive
+  held around this observation's pointing.
+  """
+  archiveDuplication: ArchiveDuplication!
```
swalker2m, line 13841: Is it not possible to have multiple duplications for the same instrument / coordinates?

cquiroz: yes `ArchiveDuplication` can contain multiple matches

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
cquiroz: Anyone can call the mutation on their own programs

cquiroz:
Here's an example of a query result for a single obs

image

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -13714,6 +13744,192 @@ type Observation {
...
+  Coordinates the file was taken at.  Null when the archive records no pointing
+  for the file.
+  """
+  coordinates: Coordinates
+
+  """
+  Archive instrument name, e.g. GMOS-N.  Always present, and always one of the
+  instruments the search queried.
+  """
+  instrument: String!
```
swalker2m: We have an `enum` for this, should we not use it?  Or are we worried that there might be instruments we don't support?

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -13714,6 +13744,192 @@ type Observation {
...
+  """
+  Observation type as the archive reports it.  Always present, and in practice
+  always OBJECT: the search asks the archive for object frames only.
+  """
+  observationType: String!
+
+  """
+  Observation class as the archive reports it, e.g. science.
+  """
+  observationClass: String
```
swalker2m: Same question here.  We have the `ObserveClass` enum.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -13714,6 +13744,192 @@ type Observation {
...
+
+  """
+  Quality assessment the file was given, e.g. Pass or Usable.  Never a failing
+  assessment: the search excludes files that failed QA.
+  """
+  qaState: String
```
swalker2m: And here, `DatasetQaState`.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -13714,6 +13744,192 @@ type Observation {
...
+  May be in the future.
+  """
+  releaseDate: Date
+
+  """
+  Program id as the archive reports it.  The archive holds both OCS- and
+  GPP-era data, so this may be a GPP program id or an OCS one, e.g.
+  GN-2019A-Q-101.
+  """
+  programId: String
```
swalker2m: We call this a `ProgramReference` in this API and a `ProgramId` is the real id `p-123`.  Same for the `observationId` below.  Perhaps the field should be called `programReference` since the OCS version is closer to that than an actual id.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -13714,6 +13744,192 @@ type Observation {
...
+
+}
+
+"The outcome of the most recent Archive Duplication Search attempt."
+enum ArchiveDuplicationState {
+
+  """
+  No search result.  Either the search has never been run, or it ran against an
+  observation the archive cannot be asked about: an instrument it does not know,
+  or no resolvable pointing and no usable target name.
```
swalker2m: It seems like `never been run` vs all the other issues would be valuable to distinguish?

modules/service/src/main/resources/db/migration/V1228__archive_duplication.sql
```diff
@@ -0,0 +1,174 @@
+-- Archive Duplication Search: the per-observation snapshot of what the Gemini
+-- Observatory Archive (GOA) returned around the observation's pointing.
+--
+-- The snapshot lives in its own table rather than in columns on t_observation
+-- deliberately to avoid generating observation edit events.
+--
+-- The values count, saturation, last checked are denormalized here
+-- so the proposal PDF and GraphQL can read them without aggregating over
+-- t_archive_match.
```
swalker2m: Rather than storing this information and worrying about it being in sync, we typically make a view that aggregates it.  Then the Grackle can read the view in a normal way as if it were in a table.  Would that make sense here?

modules/service/src/main/resources/db/migration/V1228__archive_duplication.sql
```diff
@@ -0,0 +1,174 @@
...
+);
+
+COMMENT ON TABLE t_archive_duplication IS
+  'Archive Duplication Search snapshot header, one row per observation that has been searched.';
+
+-- One row per matched archive file.  
+--
+-- The GOA record fields do not all map onto lucuma types, so the archive
+-- instrument name, disperser, filter, QA state, observation type and class, and
+-- the program and observation ids are all kept as text.
```
swalker2m: This is another place where a view, or maybe a generated field, might help.  The underlying storage could be the `text` field and a nullable translation into a schema type calculated from it to be available for the majority (?) of cases where it is possible.

modules/service/src/main/resources/db/migration/V1231__archive_duplication.sql
```diff
@@ -0,0 +1,174 @@
...
+  'Observation id as reported by GOA: a GPP observation id, or an OCS one such as GN-2019A-Q-101-11.';
+
+-- Views the Archive Duplication Search is read through.
+--
+CREATE VIEW v_archive_duplication AS
+  SELECT
+    o.c_observation_id,
+    COALESCE(d.c_state, 'not_checked'::e_archive_duplication_state) AS c_state,
+    COALESCE(d.c_match_count, 0)                                AS c_match_count,
+    COALESCE(d.c_saturated, FALSE)                              AS c_saturated,
```
swalker2m, line 129: Wait, now I'm confused.  This is what I was thinking when I wrote the comment about using a view for this.  But there are also fields for these in `t_archive_duplication`?

modules/service/src/main/scala/lucuma/odb/data/ArchiveDuplication.scala
```diff
@@ -0,0 +1,73 @@
...
+    saturated:     Boolean,
+    lastCheckedAt: Option[Timestamp],
+    error:         Option[NonEmptyString],
+    searchArea:    SearchArea
+  ) derives Eq
+
+  object Summary:
+
+    /** The summary of an observation that has never been searched. */
+    val NeverChecked: Summary =
```
swalker2m: `NotChecked`?

modules/service/src/main/scala/lucuma/odb/data/ArchiveSearchPointing.scala
```diff
@@ -0,0 +1,18 @@
...
+import eu.timepit.refined.types.string.NonEmptyString
+import lucuma.core.math.Coordinates
+
+/**
+ * The single point an Archive Duplication Search is run around.  GOA locates
+ * sidereal pointings by coordinates and moving targets by object name.
+ */
+enum ArchiveSearchPointing derives Eq:
+  case Sidereal(coordinates: Coordinates)
+  case NonSidereal(targetName: NonEmptyString)
```
swalker2m, line 25: Is this the `Nonsidereal` `key` field value or just a (possibly ambiguous) name?

modules/service/src/main/scala/lucuma/odb/service/ArchiveDuplicationSearchService.scala
```diff
@@ -0,0 +1,286 @@
...
+     * submission that lands during the GOA call cannot slip in between this read
+     * and the snapshot write that follows it in the same transaction.
+     */
+    val LockProposalStatus: Query[Observation.Id, ProposalStatus] =
+      sql"""
+        SELECT p.c_proposal_status
+        FROM t_observation o
+        JOIN t_program p ON p.c_program_id = o.c_program_id
+        WHERE o.c_observation_id = $observation_id
+        FOR UPDATE OF p
```
swalker2m: Having been so recently burned by this type of thing, I want to clarify if we need a `FOR UPDATE` lock of the program.  Anything that wants to insert a row with a program id FK reference will be blocked while this lock is held of course.  Is this locked the whole time while a remote service is being called?  If so, we might want to be more optimistic and just check whether there was an update after the GOA call and if so, retry?

modules/service/src/main/scala/lucuma/odb/service/ArchiveDuplicationService.scala
```diff
@@ -0,0 +1,280 @@
...
+import skunk.implicits.*
+
+import java.time.Instant
+import java.time.LocalDateTime
+import java.time.ZoneOffset
+
+/**
+ * Storage for Archive Duplication Search snapshots.  Running the search itself
+ * is the caller's job; this service only reads and replaces what was found.
+ */
```
swalker2m, line 33: A quick glance at this didn't reveal anything requiring a program row lock.  If that's correct, another snapshot could be written while the lock is held anyway.  I'm probably not understanding the situation though.

swalker2m (commented):
This is an impressive amount of work and it looks great to me.  I just have a few questions where I got confused.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
swalker2m: Maybe `instrumentName: String!` and `instrument: Instrument` where the later comes from a generated field or view.

cquiroz: Right we don't really have a full one to one mapping with goa, do we? i think there maybe other instruments there? michelle, etc? I guess this is fine as we will query known instruments

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
cquiroz: Does it match ocs classes one to one?

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
cquiroz: Sounds fine

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
cquiroz: Seems good, i think we will have to add an obscalc mechanism to alays do checks but yes it would be good to distinguish between never run and no results

modules/service/src/main/resources/db/migration/V1228__archive_duplication.sql
cquiroz: right, i think it could help

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
swalker2m: I would expect it to, but we could offer both `observeClassString: String` and `observeClass: ObserveClass` where the later is only defined if it can be mapped.  Maybe nobody cares though and it is just for display.  In which case, no need to bother.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
swalker2m: I guess the obscalc itself could add this to the checks it does, eventually, as long as the timeout is tight.  It probably only applies until an observation is ready at which point, it doesn't much need to be updated anymore?

cquiroz: Actually we only update it after the proposal has been submitted and I think on some workflow changes.
GOA could be a bit slow so we may want to do it separately? On side i don't want to have yet another daemon but otoh maybe we need to, to avoid overcharging obscalc as the goa checks are lower priority

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
cquiroz: One counter example is old IGRINS, which is not present as an enum in lucuma but it is valid in GOA and actually we may want to use. We could add `GoaInstrument` enum that includes thos legacy instruments

swalker2m: I think we have to treat the GOA as essentially an external service.  We have no control over what we might get back from it.  So your original `instrument: String!` seems fine (though maybe we call it `instrumentName`?) if we don't ever do anything with that information or else we return the raw `String` and a parsed `Instrument` if possible?

cquiroz: it is external and doing a sampling i don't think they are fully normalized there, e.g. for qa states I got
CHECK, Fail, Pass, Undefined, Usable

casing is not consistent

swalker2m: Yeah, all those except `CHECK` map to our `DatasetQaState` (where `Undefined` would be simply not set).  Check was an old supposedly transient state that prompted somebody to take a closer look at the dataset and set a definitive state.  I think those map to not set as well.

modules/service/src/main/resources/db/migration/V1228__archive_duplication.sql
cquiroz: you're right we can add a view for count, saturation and last checked are the results of a query thus need to be stored

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
cquiroz: Add a new state for never run

modules/service/src/main/scala/lucuma/odb/service/ArchiveDuplicationSearchService.scala
cquiroz: I'm changing this to use "FOR NO KEY.UPDATE"

modules/service/src/main/scala/lucuma/odb/service/ArchiveDuplicationSearchService.scala
```diff
@@ -279,12 +282,23 @@ object ArchiveDuplicationSearchService:
...
+     * `FOR NO KEY UPDATE` rather than `FOR UPDATE`, for the reason V1227 gives:
+     * `FOR UPDATE` is the only mode conflicting with the `FOR KEY SHARE` that
+     * PostgreSQL takes on `t_program` to enforce the foreign key of every child
+     * row inserted, so holding it would block those inserts for the duration.
+     *
+     * It still serializes against submission, by a wider margin than the mode
+     * name suggests: `c_proposal_status` feeds `c_program_reference`, a STORED
+     * generated column carrying a UNIQUE constraint (V0845), so submission
+     * changes a key column and its `UPDATE` takes `FOR UPDATE` — which conflicts
+     * with every mode, this one included.
```
swalker2m, line 295: What is being said here?  I think it means that submission doesn't happen simultaneously which is good?

modules/service/src/main/scala/lucuma/odb/data/ArchiveSearchPointing.scala
```diff
@@ -12,6 +12,13 @@ import lucuma.core.math.Coordinates
 /**
  * The single point an Archive Duplication Search is run around.  GOA locates
  * sidereal pointings by coordinates and moving targets by object name.
+ *
+ * The non-sidereal name is the GPP target's own name, not its `EphemerisKey`
+ * designation.  GOA matches it against the `object` value the observer recorded
+ * in the file's headers, which is free text of the same kind, so a designation
+ * would match less often rather than more.  Matching moving targets is therefore
+ * by name and inherently approximate: a file archived under a different spelling
+ * of the same body is not found.
```
swalker2m, line 21: Alright, but somewhere in the headers we should probably put the ephemeris key to tighten this up.  Some day.  Hehe.

swalker2m (approved):
Looks good. :+1:

PR #2762, opened 2026-07-24
igrins 2 acquisition slit viewing camera svc

cquiroz:
To fully support igrins2 svc acquisition images we need a few new fields to hold the exposure time and offsets

Note sequence generation has to change when svc is present, that will be done in a separate PR

Don't merge this PR until the explore counter PR is ready


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -10369,6 +10359,70 @@ input Igrins2LongSlitInput {
...
+  explicitTelescopeConfigs: [TelescopeConfig!]
+
+}
+
+"""
+Edit or create IGRINS-2 Slit-Viewing Camera (SVC) configuration. A non-null
+object enables SVC image saving (turning it on if necessary) and applies the
+given sub-field overrides; a null value disables SVC image saving. Omitting the
+field leaves SVC state unchanged.
+"""
```
swalker2m, line 10409: There's a default exposure time and default telescope configs.  What if you want to save the SVC images but you don't want to change either default.  Is it valid to pass empty object?  As in:

```graphql
{
  ...
  svc: {}
  ...
}
```

Probably worth a test if there isn't one.

modules/service/src/main/resources/db/migration/V1228__igrins2_svc.sql
```diff
@@ -0,0 +1,47 @@
...
+ALTER TABLE t_igrins_2_long_slit
+  ADD COLUMN c_svc_exposure          interval NULL,
+  ADD COLUMN c_svc_telescope_configs text     NULL;
+
+-- SVC exposure, when explicitly set, must respect the IGRINS-2 detector limits
+ALTER TABLE t_igrins_2_long_slit
+  ADD CONSTRAINT igrins2_svc_exposure_check
+    CHECK (c_svc_exposure IS NULL
+           OR (c_svc_exposure >= interval '3080 milliseconds'
+               AND c_svc_exposure <= interval '600 seconds'));
```
swalker2m, line 14: Do we need to check / update existing rows first?  Maybe there aren't any.

modules/service/src/test/scala/lucuma/odb/graphql/mutation/updateObservations.scala
```diff
@@ -4562,6 +4559,56 @@ class updateObservations extends OdbSuite with UpdateObservationsOps with Execut
...
+  test("igrins2 SVC toggle: off by default; on/off preserves params while off"):
+    val query = "observations { observingMode { igrins2LongSlit { svc { exposure { microseconds } } } } }"
+    def svcOn(microseconds: Long): Either[String, Json] =
+      Right(json"""{ "updateObservations": { "observations": [{ "observingMode": { "igrins2LongSlit": { "svc": { "exposure": { "microseconds": $microseconds } } } } }] } }""")
+    val svcOff: Either[String, Json] =
+      Right(json"""{ "updateObservations": { "observations": [{ "observingMode": { "igrins2LongSlit": { "svc": null } } }] } }""")
+    multiUpdateTest(pi,
+      List(
+        // off by default; turn ON with defaults (exposure 3.08s)
+        ("observingMode: { igrins2LongSlit: { svc: {} } }", query, svcOn(3080000L)),
```
swalker2m, line 4571: This looks like the case I was mentioning. :+1:

renumber-migrations.sh
```diff
@@ -0,0 +1,167 @@
...
+Usage: renumber-migrations.sh [--dry-run] [base-branch]
+
+Renumbers the Flyway migrations added by the current branch so they follow the
+base branch's highest migration number, contiguously and in their existing
+relative order.
+
+  base-branch   Branch to renumber against (default: main).
+  --dry-run     Report what would change without touching anything.
+EOF
+}
```
swalker2m, line 29: Seems like a useful tool.

swalker2m (approved):
Looks good to me.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
cquiroz: it is valid and that would set the default values, i think i had a test for that but let me review
I expect that mutation to be the one used by default by explore

renumber-migrations.sh
cquiroz: Yeah I got tired of manually renaming them

PR #2763, opened 2026-07-24
retry 'Interrupted system call' failures

swalker2m:
There's an issue with the `concurrentInsertDeadlock` test that bites us on occasion in CI.  The failure is accompanied by an "Interrupted system call" Postgres exception that I don't believe that it is related to deadlocks but rather to the CI environment itself.  A retry, when the test fails with this message should bypass the problem.


cquiroz (approved):
LGTM

PR #2766, opened 2026-07-27
Update igrins2 sequence generation to support SVC

cquiroz:
Will add acquistion steps to igrins if svc config is enabled

### DIFF
--- a/modules/CLAUDE.md
+++ b/modules/CLAUDE.md
@@ -253,7 +253,7 @@ When writing a `Fragment` / `encoder` for an INSERT that maps to a table with `c
 |---|---|---|---|
 | GMOS North/South LongSlit, Flamingos-2 LongSlit | (none inline; comes from observation/defaults) | `t_exposure_time_mode` (role=acquisition) | `t_exposure_time_mode` (role=science) |
 | GNIRS LongSlit | Inline columns in `t_gnirs_long_slit`: `c_acq_type`, `c_acq_coadds`, `c_acq_filter`, `c_acq_sky_offset_p`, `c_acq_sky_offset_q` (the two components together form a single `Option[Offset]` sky offset, enforced by a both-or-neither CHECK) | `t_exposure_time_mode` (role=acquisition) | `t_exposure_time_mode` (role=science) |
-| IGRINS-2 LongSlit | (no acquisition sequence) | — | `t_exposure_time_mode` (role=science) |
+| IGRINS-2 LongSlit | Inline SVC (Slit-Viewing Camera) columns in `t_igrins_2_long_slit`: `c_save_svc_images`, `c_svc_exposure`, `c_svc_telescope_configs`. No acquisition sequence when SVC images aren't saved. | — | `t_exposure_time_mode` (role=science) |
 | GHOST IFU | — | — | Two rows (red + blue) in `t_exposure_time_mode` |
 
 The `check_etm_consistent` trigger function must be updated for each new mode.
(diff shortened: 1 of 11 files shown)


swalker2m (commented):
This all looks great to me, but I want to verify one thing.  We have an `IGRINS_2_SPECTROSCOPY` `ItcType` in the schema:

```graphql
"""
ITC results for IGRINS-2 spectroscopy observations (no acquisition).
"""
type ItcIgrins2Spectroscopy implements Itc {
  itcType: ItcType!
  spectroscopyScience: ItcResultSet!
}
```

Now we will have an optional acquisition?  I get that it isn't a normal acquisition though so maybe there's no need to add it here.  The ITC isn't contacted for this?  We might want at least to update the comment there.

Note also the `Itc` JSON encoder:

```scala
  given (using Encoder[TimeSpan], Encoder[Wavelength]): Encoder[Itc] =
    Encoder.instance: itc =>
      val science = itc.science.asJson
      itc.science match
        case _: ItcScience.Spectroscopy =>
          itc.acquisition match
            case ItcAcquisition.Available(times, _) =>
              science.deepMerge(Json.obj("acquisition" -> times.asJson))
            case _                                  =>
              science.deepMerge(Json.obj("itcType" -> "igrins_2_spectroscopy".asJson))
        case _                          =>
          science
```

I was a little embarrassed by this.  It uses the absence of an acquisition sequence to infer `IGRINS_2_SPECTROSCOPY`.  As is, this continues to work but if we were to add this to the `ItcIgrins2Spectroscopy` type we'd have to do something here.

cquiroz:
> This all looks great to me, but I want to verify one thing. We have an `IGRINS_2_SPECTROSCOPY` `ItcType` in the schema:
> 
> ```graphql
> """
> ITC results for IGRINS-2 spectroscopy observations (no acquisition).
> """
> type ItcIgrins2Spectroscopy implements Itc {
>   itcType: ItcType!
>   spectroscopyScience: ItcResultSet!
> }
> ```
> 
> Now we will have an optional acquisition? I get that it isn't a normal acquisition though so maybe there's no need to add it here. The ITC isn't contacted for this? We might want at least to update the comment there.

Right, no ITC call for acq. igrins2 does acquisition separately so acquisition is not fully correct name here but science requested to treat svc just as a regular acquisition. But you are right about the comment.

> 
> Note also the `Itc` JSON encoder:
> 
> ```scala
>   given (using Encoder[TimeSpan], Encoder[Wavelength]): Encoder[Itc] =
>     Encoder.instance: itc =>
>       val science = itc.science.asJson
>       itc.science match
>         case _: ItcScience.Spectroscopy =>
>           itc.acquisition match
>             case ItcAcquisition.Available(times, _) =>
>               science.deepMerge(Json.obj("acquisition" -> times.asJson))
>             case _                                  =>
>               science.deepMerge(Json.obj("itcType" -> "igrins_2_spectroscopy".asJson))
>         case _                          =>
>           science
> ```
> 
> I was a little embarrassed by this. It uses the absence of an acquisition sequence to infer `IGRINS_2_SPECTROSCOPY`. As is, this continues to work but if we were to add this to the `ItcIgrins2Spectroscopy` type we'd have to do something here.

I guess it would also be a problem if another instrument had no acquisition?

swalker2m:
> I guess it would also be a problem if another instrument had no acquisition?

Right, if there were another _spectroscopy_ observing mode without an acquisition.

swalker2m (approved):
We might just clarify that `(no acquisition)` comment about the `ItcIgrins2Spectroscopy` type in the schema, but this looks good to me. :+1:

PR #2768, opened 2026-07-27
Use RetryFlakyTests trait for SSO and ODB tests

hugo-vrijswijk:
Also remove the `FlakyTests` trait from SSO and mark those tests as flaky instead.

For reviewing, I recommend clicking the cogs and enabling "Hide whitespace"


modules/sso-service/src/test/scala/graphql/mutation/addRole.scala
```diff
@@ -10,122 +10,117 @@ import lucuma.core.enums.Partner
...
-                  .hcursor
-                  .downFields("data", "addRole")
-                  .as[StandardRole.Id]
-                  .isRight
-
-  test("Admin Alice should be able to give Bob an Admin role, and aftewards he should be able to give a Staff role to Alice"):
-    flaky():
-      AsBob
-        .queryIds
+    test(s"$role should not be able to give Bob a role".flaky):
```
cquiroz, line 16: Why are these tests flaky?

hugo-vrijswijk: Unsure, but since they were already marked as such I didn't want to change too much. Determining if a test is flaky is difficult :D

cquiroz: i guess so, weird that it looks like most/many sso tests are flaky

cquiroz (approved):
LGTM

PR #2770, opened 2026-07-27
Add execution event visit index

swalker2m:
Adds an index (or rather replaces one that was inadvertently dropped from previous event tables).  The goal is to improve event ingestion time.  With every new event, we update visit min/max times.  Without an index, we were essentially doing an ever more costly table scan (I think).  This should be an easy, no risk win (hehe).


cquiroz (approved):
LGTM

PR #2772, opened 2026-07-27
Update igrins2 sequence generation

cquiroz:
* No breakpoints
* Support adding acquisition steps when the user requires them


toddburnside (approved):
LGTM

PR #2773, opened 2026-07-27
Allow client-specified event and visit times

swalker2m:
## Update in this PR

This PR adds an optional client time for execution events and visit records.  When used, time accounting would be based on these timestamps instead of event reception time at the database.  As you can see, this is a rather involved update so I'm uncertain as to whether we should accept it and would like your feedback.

## Major Drawback

The main problem with opening up time accounting timestamps to the client is complexity.  The database has no control over the supplied values, which could be anything and could be sent in any order.  I'm guarding against outlandish values by checking that they fall within 5 minutes of what time the database thinks it is, but it makes me nervous regardless.

## Potential Future Benefit

The main benefit that I can see is that it will allow us to batch dataset events (in the next PR).  Instead of sending six separate dataset events they could optionally be batched into a new `addDatasetEvents` mutation where each stage is associated with a client timestamp.  That would be one round-trip and one time-accounting update instead of individual updates for each stage.  (Or perhaps batch everything up to `StartExpose` and then everything from `EndExpose`?)

On the other hand, batching them brings risk of some problem in Observe leading to the events never being stored at all. So I'm not completely convinced.

We could instead eliminate the fine-grained details in `DatasetStage` and simply record start and end dataset events.


swalker2m:
Now that a few updates to event handling have been made, I'd like to re-evaluate whether we need this at all.  If recording an event is now fast enough, there may be no need to take any action.

toddburnside:
> Now that a few updates to event handling have been made, I'd like to re-evaluate whether we need this at all. If recording an event is now fast enough, there may be no need to take any action.

It seems like it's worth waiting for them to test the event recording performance before making a decision on this.

swalker2m:
The expert consensus is that this update is essential, so I'll re-up this PR.  I'm sorry.

cquiroz:
whose experts?

swalker2m:
Postgres doesn't want to start in CI to run the migration tests.

modules/service/src/main/resources/db/migration/V1233__visit_client_created.sql
```diff
@@ -0,0 +1,81 @@
+-- Client-supplied visit creation times.
+--
+-- A visit's recorded time is the moment the ODB first created the row -- either
+-- from a slew event (application "Navigate") or a recordVisit (application
+-- "Observe").  As with execution events, any delay between the visit actually
+-- opening at the telescope and the row being written skews the derived observing
+-- night (which drives the daylight discount).  We now let the creating client
+-- optionally send the real creation time.  The columns mirror the GraphQL API:
```
toddburnside, line 8: Both migrations mention that the columns mirror the GraphQL API. But, client time doesn't seem to be in the graphql query responses. Should it be?

modules/service/src/main/resources/db/migration/V1231__visit_client_created.sql
```diff
@@ -0,0 +1,81 @@
...
+
+  RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER check_visit_client_time_trigger
+  BEFORE INSERT ON t_visit
+  FOR EACH ROW EXECUTE FUNCTION check_visit_client_time();
+
+-- The execution-event time check (V1230) reads the visit's recorded creation time
```
toddburnside: I think the migrations got renumbered?

toddburnside (approved):
LGTM

modules/service/src/main/resources/db/migration/V1233__visit_client_created.sql
swalker2m: We have the `recordedTime`, which is when the event was received and the `effectiveTime` which is the client time when provided, or otherwise the `recordedTime`.  What is missing is just a nullable plain client time.  I could add that if we think it would be useful.  I guess it would round it out.

modules/service/src/main/resources/db/migration/V1231__visit_client_created.sql
swalker2m: Right.  In the time it takes to write a PR description migrations usually need to be renumbered.  I'll fix it.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -11,6 +11,13 @@ input AddDatasetEventInput {
   "Dataset execution stage."
   datasetStage: DatasetStage!
 
+  """
+  Time the event occurred, if the client wishes to supply it.  When omitted, the
+  time the event is received by the ODB is used instead.  A supplied time that
+  falls outside the visit's expected timeframe is rejected.
+  """
+  time: Timestamp
```
cquiroz: I suppose there are always UTC?

modules/service/src/main/resources/db/migration/V1231__visit_client_created.sql
cquiroz: and i see v1231 and v1233 not a v1232

modules/service/src/main/scala/lucuma/odb/service/ClientTimeError.scala
```diff
@@ -0,0 +1,26 @@
...
+ * dedicated SQLSTATE when a client-supplied event/visit time is outside the
+ * accepted range.  Matching on that exact code (rather than "any raised
+ * exception") guarantees we only translate *this* error into a typed OdbError and
+ * let every other database error propagate untouched.
+ *
+ * Kept in sync with the `USING ERRCODE` clauses in the V1230 and V1231 migrations.
+ */
+object ClientTimeError:
+
+  // Custom code; class 'OD' is not used by PostgreSQL, so nothing else raises it.
```
cquiroz, line 19: is this  custom error code? user supplied?

modules/service/src/main/resources/db/migration/V1234__execution_event_client_time.sql
```diff
@@ -0,0 +1,202 @@
...
+  RETURNS TRIGGER AS $$
+DECLARE
+  visit_created timestamp;
+BEGIN
+  IF NEW.c_client_time IS NOT NULL THEN
+    SELECT c_created INTO visit_created
+      FROM t_visit
+     WHERE c_visit_id = NEW.c_visit_id;
+
+    IF NEW.c_client_time < visit_created - interval '5 minutes'
```
cquiroz, line 64: I see this baked in the odb. who supplies `visit_created`?

modules/service/src/main/scala/lucuma/odb/service/ClientTimeError.scala
toddburnside: It is the code used by the `RAISE EXCEPTION`s in the migrations.

modules/service/src/main/resources/db/migration/V1233__visit_client_created.sql
swalker2m: Exposed `clientTime` as queryable field.

modules/service/src/main/resources/db/migration/V1231__visit_client_created.sql
swalker2m: I missed one in the renumbering and scrambled the order.  Now fixed.

modules/service/src/main/resources/db/migration/V1234__execution_event_client_time.sql
swalker2m: This was scrambled in the migration renumbering but now should correctly reflect the client-supplied visit time if it was given.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
swalker2m: Right, always UTC `Z` zero meridian in the format.

```
"""
Timestamp of time in ISO-8601 representation in format '2011-12-03T10:15:30Z'
"""
scalar Timestamp
```

PR #2776, opened 2026-07-28
Support pending masks

cquiroz:
This PR adapts odb to use the new ADT from lucuma core for pending MOS masks
We are still not supporting defining mos observations but we can tell when a custom mask is yet to be defined in phase 2

For ITC we just always asume it is `ToBeDefined.`


toddburnside (approved):
LGTM

PR #2777, opened 2026-07-28
increase input value depth; add test

tpolecat:
This increases the allowed complexity of input values for SSO's graphql interface. (Andy ran into this).

### DIFF
--- a/modules/sso-service/src/test/scala/graphql/query/users.scala
+++ b/modules/sso-service/src/test/scala/graphql/query/users.scala
@@ -153,3 +153,65 @@ class users extends GraphQLSuite with SsoSuite with Fixture with OrcidIdGenerato
           }
         """
     )
+
+  test("Andy issue ('exceeded maximum input value depth').".flaky):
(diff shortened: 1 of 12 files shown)


modules/sso-service/src/test/scala/graphql/query/users.scala
```diff
@@ -153,3 +153,65 @@ class users extends GraphQLSuite with SsoSuite with Fixture with OrcidIdGenerato
           }
         """
     )
+
+  test("Andy issue ('exceeded maximum input value depth').".flaky):
```
toddburnside: Is this test really `flaky`?

toddburnside (approved):
LGTM

modules/sso-service/src/test/scala/graphql/query/users.scala
tpolecat: The tests used to be flaky but they shouldn't be anymore so I'll remove those and see what happens.

PR #2779, opened 2026-07-29
Pick up GNIRS IFU signal-to-noise fix from the OCS ITC

toddburnside:
Refreshes the ocslib jars to include the OCS GnirsRecipe change that computes signalToNoiseAt for the IFU branch. Without it, GNIRS IFU configurations returned no S/N at the requested wavelength, which left the S/N column of the Explore spectroscopy modes table blank in time-and-count mode.

The existing "gnirs IFU" legacy tests did not catch this: they iterate with `foreach`, which discards the IO that `assertIOBoolean` returns, so the assertions never ran. Both are switched to `traverse_` and now assert the S/N through the new `containsValidResultsWithSNAt` helper. `containsValidResults` is left alone, since an absent signalToNoiseAt is legitimate for other modes. Verified in both directions: the two tests fail against the previous jars and pass against the rebuilt ones.

Note the same `foreach` pattern remains in roughly forty other cases across the legacy suites, which are likewise not asserting anything.

Also adds a GNIRS IFU case to spectroscopyTimeAndCountSuite covering the `ifu` arm of the @oneOf GnirsFpuSpectroscopyInput, which no existing case exercised. That suite runs against MockItc, so it is input-wiring coverage only.

Claude-Session: https://claude.ai/code/session_01Q8HPfo5nknxUYmpQkYZVnb


itc/legacy-tests/src/test/scala/lucuma/itc/legacy/LegacyITCGnirsSpecExpTimeSuite.scala
```diff
@@ -99,15 +99,18 @@ class LegacyITCGnirsSpecExpTimeSuite extends CommonITCLegacySuite:
         ).asJson.noSpaces
       assertIOBoolean(result.map(_.fold(allowedErrors, containsValidResults)))
 
+  // Note `traverse_`, not `foreach`: `assertIOBoolean` returns an IO, and `foreach` discards it,
+  // which would make the assertion a no-op.
   test("gnirs IFU".tag(LegacyITCTest)):
-    Enumerated[GnirsFpuIfu].all.foreach: ifu =>
+    Enumerated[GnirsFpuIfu].all.traverse_ { ifu =>
```
toddburnside, line 105: I will be sending another PR which fixes all the tests that are using `foreach` improperly.

cquiroz: Great catch, i was under the impression that the assert by themselves would make the test fail

cquiroz (approved):
LGTM

PR #2780, opened 2026-07-29
Make GNIRS/F2 science flats optional

cquiroz:
image

### DIFF
--- a/modules/sequence/src/main/scala/lucuma/odb/sequence/SmartGcalExpander.scala
+++ b/modules/sequence/src/main/scala/lucuma/odb/sequence/SmartGcalExpander.scala
@@ -4,10 +4,11 @@
 package lucuma.odb.sequence
 
 import cats.Applicative
-import cats.Functor
+import cats.Monad
 import cats.data.NonEmptyList
 import cats.syntax.applicative.*
 import cats.syntax.either.*
+import cats.syntax.flatMap.*
 import cats.syntax.functor.*
 import lucuma.core.enums.ObserveClass
 import lucuma.core.enums.SmartGcalType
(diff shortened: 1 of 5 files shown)


modules/sequence/src/main/scala/lucuma/odb/sequence/SmartGcalExpander.scala
```diff
@@ -44,14 +45,22 @@ trait SmartGcalExpander[F[_], S, D]:
...
+    arc:    ProtoStep[D]
+  )(using Monad[F]): F[Either[String, NonEmptyList[ProtoStep[D]]]] =
+    for
+      f <- expandStep(static, flat)
+      a <- expandStep(static, arc)
+    yield (f, a) match
+      case (Right(fs), Right(as)) => (fs ::: as).asRight
+      case (Right(fs), Left(_)  ) => fs.asRight
+      case (Left(_),   Right(as)) => as.asRight
+      case (Left(m),   Left(_)  ) => s"$m (a flat or an arc is required)".asLeft
```
swalker2m: I don't think I wrote `expandStepOptional` but it seems like a useful concept.  Maybe we could write `expandFlatAndOrArc` (seems it should contain "or" in the name?) in terms of it instead of replacing it?

swalker2m (approved):
Looks good.  Seems useful to keep `expandStepOptional` but either way :+1:

modules/sequence/src/main/scala/lucuma/odb/sequence/SmartGcalExpander.scala
swalker2m: I mean something like

```scala
for
  fs <- expandStepOptional(static, flat)
  as <- expandStepOptional(static, arc)
yield
  NonEmptyList
    .fromList(fs ::: as)
    .toRight("A flat or an arc is required, but neither has a SmartGcal mapping for this configuration.")

```

cquiroz: I added `expandStepOptional` for a similar bug where no arcs made sequence generation fail. This is the same bug for flats

PR #2781, opened 2026-07-29
Make the ITC legacy tests actually assert

toddburnside:
Every looping test in these suites iterated with `foreach`:

    Enumerated[GnirsFpuIfu].all.foreach: ifu =>
      val result = localItc.calculate(...)
      assertIOBoolean(result.map(_.fold(allowedErrors, containsValidResults)))

`assertIOBoolean` returns an `IO`, and `foreach` expects `Unit`, so the `IO`
was discarded and never run. The assertion was a no-op and the test passed
unconditionally. That is how the GNIRS IFU signal-to-noise gap went unnoticed:
the "gnirs IFU" test passed against both the broken and the fixed OCS jars.

This affected 53 sites. Fifteen were in the shared generators (testConditions,
testSEDs, testBrightnessUnits, testPowerAndBlackbody), which twelve suites call,
so roughly 180 test instances were asserting nothing.

Replaces the pattern with an `assertAllValid` helper that runs the body for
every value and fails listing every value that was rejected along with why.
Collecting all offenders rather than failing fast matters here: this is the
first time the assertions have run, so a single run has to be a triage pass.
The three `.traverse:` sites in LegacyITCFlamingos2Suite were already correct
but are folded in too, so only one idiom remains to copy.

Turning the assertions on surfaced nine failures, all now resolved:

- Seven were stale `allowedErrors` patterns. OCS 41f8bdd14 (REL-4806) rewrote
  GmosRecipe's messages to "Insufficient signal at N nm" and "N exposures
  required for this configuration", but the list still matched the old wording.
  Note GnirsRecipe and Flamingos2Recipe were left on the older phrasing, so
  both forms are needed. Dropped "Invalid S/N", which matches nothing in either
  the OCS or this repo any more.
- Two were a bad fixture: the GNIRS 10 l/mm grating cannot be used with the
  0.15"/pix short camera. Added `gnirsCameraForGrating`, mirroring the existing
  `gnirsCameraForIfu`, so D10 is exercised rather than excused.
- One was GNIRS imaging with the Order1 / PAH filters reporting "Could not find
  best read mode", accepted as a legitimate rejection.

Also raises munitIOTimeout. These tests previously did nothing but build JSON
strings; they now make around 3000 real ITC calls per run, and the largest
single test (158 StellarLibrarySpectrum values) exceeded the 30s default. The
legacy-tests workflow will go from seconds to roughly twenty minutes.

Full suite: 301 tests, 0 failed, 299 passed, 2 skipped.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01Q8HPfo5nknxUYmpQkYZVnb

### DIFF
--- a/itc/legacy-tests/src/test/scala/lucuma/itc/legacy/CommonITCLegacySuite.scala
+++ b/itc/legacy-tests/src/test/scala/lucuma/itc/legacy/CommonITCLegacySuite.scala
@@ -32,11 +32,13 @@ import lucuma.itc.service.ItcObservingConditions
 import lucuma.itc.service.Main.ReverseClassLoader
 import lucuma.itc.service.ObservingMode
 import lucuma.itc.service.TargetData
+import munit.Location
 import munit.Tag
 
 import java.io.File
 import java.io.FileFilter
 import scala.collection.immutable.SortedMap
+import scala.concurrent.duration.*
 
 object LegacyITCTest extends Tag("LegacyItcTest")
 
@@ -46,6 +48,11 @@ import munit.CatsEffectSuite
  * This is a common trait for tests of the legacy ITC code
  */
 trait CommonITCLegacySuite extends CatsEffectSuite:
+
+  // Each of these tests makes one real (and slow) call into the legacy ITC per enum value, and
+  // some of those enums are large — StellarLibrarySpectrum alone has 158. The default 30s is not
+  // enough now that the assertions actually run.
+  override def munitIOTimeout: Duration = 15.minutes
 
   // Common validation functions
   def containsValidResults(r: IntegrationTimeRemoteResult): Boolean =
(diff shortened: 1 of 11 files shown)


itc/legacy-tests/src/test/scala/lucuma/itc/legacy/CommonITCLegacySuite.scala
```diff
@@ -57,23 +64,59 @@ trait CommonITCLegacySuite extends CatsEffectSuite:
...
-      err.exists(_.contains("target is too bright")) ||
-      err.exists(_.contains("Invalid SignalToNoise value"))
+  /**
+   * Messages the legacy ITC produces when a configuration simply is not calculable, rather than
+   * because something is wrong.
+   *
+   * The wording is not consistent across instruments: REL-4806 rewrote the messages in GmosRecipe
+   * but left GnirsRecipe and Flamingos2Recipe on the older phrasing, so both forms are listed here.
+   */
+  private val notCalculablePatterns: List[String] = List(
```
cquiroz, line 74: 👍

cquiroz (approved):
LGTM

PR #2789, opened 2026-07-30
Shortcut 9747: ToO Trigger First Class Object

swalker2m:
Submitted for discussion, this PR develops a concrete example of the concept of a first class ToO trigger object. It envisions ToO triggers as objects with an ID and a short lifecycle.  It is agnostic about _how_ a ToO observation is created (copied from a template, using supplied pre-approved configurations, etc.).  As of yet it makes no restrictions about which observations can be triggered (no ToO types or checks along those lines).  This would be added in subsequent work.  What it does provide is:

* __Lifecycle.__  A PI sends a `request` and may later `withdraw` the request if it has not been acted upon.  Staff may `accept`, or `deny` with a reason. The only non-terminal state is `requested`.  If the request is denied, withdrawn or even accepted but the observation isn't executed, a new request would have to be made (presumably after edits to the observation).

* __Chron.__ Keeping up with the history of a ToO request (who did what and when) is an explicit requirement.  The chronicle tables we've developed in a few cases seem like a good template for this, so I've added that.

* __Events.__ The database sends ToO trigger events. Clients, such as the scheduler and observe, may subscribe to them.

This is big, but largely mechanical.  It is probably enough to look at the `OdbSchema.graphql` and the `V1233` SQL migration to get a sense of it.  There are sanity checks but I didn't go into great detail since this is all very much subject to change.


toddburnside:
So, TooTriggers are requested for "real" observations (non-ToO targets) which have been created as "implementations" of ToO observations? Not the ToO observation itself? 

For high priority ToOs, is the plan to have an automated accept or something of that nature?

swalker2m:
> So, TooTriggers are requested for "real" observations (non-ToO targets) which have been created as "implementations" of ToO observations? Not the ToO observation itself?

There are a few scenarios for how the actual observation gets created.  I wanted to remain agnostic on those options.  One of them would be to replace the ToO target in an observation with a real target and trigger that.  Another would be to copy the ToO observation, slot in the now known target and trigger that.  Another would be to just supply an approved configuration and a target and have the system create the observation and trigger that.  None of this is covered by the PR.

So there may be mutations that do more than the minimal trigger creation mutation provided here.  At the very least, this is a building block.

> For high priority ToOs, is the plan to have an automated accept or something of that nature?

No, I think the idea is for the observer to be notified in `Observe` and take a decision.  I don't believe we'll automatically switch.

modules/service/src/main/scala/lucuma/odb/data/TooTriggerStatus.scala
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * This deliberately excludes observation-level concerns.  `Accepted` is terminal
+ * regardless of whether the observation ultimately executes, and there is no
+ * `Expired` state -- expiration is a function of the observation's timing window,
+ * which lives on the observation's own workflow, not here.
+ *
+ * Defined here in the odb project for now; move to lucuma-core once the design
+ * settles.
+ */
+sealed abstract class TooTriggerStatus(val tag: String) extends Product with Serializable
```
toddburnside: Do we still need to extend `Product with Serializable`? And could this be an `enum` instead of a class?

toddburnside (approved):
It's a lot to take in, but it seems fine to me.

swalker2m:
Because it is not hooked up to anything yet, it might make sense to explore what actually happens when a new ToO request is made.  The creation of the observation itself will be its own phase prior to the trigger as mentioned above.  Assuming it now exists, its workflow state should probably not be allowed to advance beyond `Defined`.  Only when the observer accepts the request should it move to `Ready`.  At that point it is up to the scheduler and observer to figure out when to do the execution, be it immediately or in due course.  Does this sound reasonable?

modules/service/src/main/scala/lucuma/odb/data/TooTriggerStatus.scala
swalker2m: Good point, I'll switch it.

toddburnside:
> Does this sound reasonable?

Well, that depends on your definition of `reasonable`. 😆 Tying into the workflow state is, of course, going to be complicated. But, that's probably inevitable.

modules/service/src/main/scala/lucuma/odb/data/TooTriggerStatus.scala
swalker2m: Switched to enum.

PR #2790, opened 2026-07-30
Support replacing ghost sequence

cquiroz:
Somehow me missed enabling manual editing of ghost


toddburnside (approved):
Interesting. It was in the sequence service, but never wired up, eh? LGTM

cquiroz:
Yeah, some oversight

PR #2791, opened 2026-07-31
Batch sequence insert statements

cquiroz:
Investigating observe events via tracing, one of them seems to be taking a bit longer. Record visit only happens once per sequence, so we impact is not as big, but still we can improve it by doing batch insertion of steps

This PR batches rows into groups of 256 and issues multi-row INSERT ... VALUES (...), (...), ... statements instead

Measured on a RecordVisitMutation trace (13-atom sequence, 52 steps igrins2):

| Metric           | Before | After           |
|------------------|--------|-----------------|
| Server time      | 310 ms | 151 ms (**-51%**) |
| SQL exec time    | 187 ms | 71 ms (-62%)    |
| # SQL statements | 138    | 24 (-83%)       |

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/Flamingos2SequenceService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/Flamingos2SequenceService.scala
@@ -118,6 +118,15 @@ object Flamingos2SequenceService:
           $flamingos_2_dynamic
       """.command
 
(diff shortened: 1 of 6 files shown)


swalker2m (approved):
The first record visit is more expensive because it writes the heretofore dynamic sequence into the database, as you've seen.  This seems like a very worthy improvement.  Maybe the batch size could even be a bit bigger.  :+1:

cquiroz:
yeah it is a pity we don't send this more often to get more out of this fix :D Not sure about the batch size i guess 256 covers most sequences?

PR #2792, opened 2026-07-31
Shortcut 9748: ToO execution requirement

swalker2m:
This PR adds an `ExecutionRequirement` scheduling constraint.  It applies to all observations, not just ToOs, because it is describing what the scheduler may do to a running observation, say when a ToO arrives.  The PR reworks the existing `isSplittable` boolean scheduling constraint into a three-valued enumeration:

* `Unconstrained` - the scheduler may break the sequence into visits at atom boundaries, and interrupt it between atom boundaries if necessary
* `Contiguous` - the scheduler may interrupt the observation, but it will only have a single atom so it cannot be split across visits without redoing any part already observed
* `Uninterruptible` - this observation is `Contiguous` (not splittable) and may not be interrupted to switch to another observation

I've left `isSplittable` as a deprecated field, mapping to the execution requirement appropriately.


toddburnside (approved):
LGTM. I'll just have to find out Explore should handle this. We currently allow the pi to set whether or not the observation can be split. I don't know how `uninterruptible` fits into the picture.

swalker2m:
> LGTM. I'll just have to find out Explore should handle this. We currently allow the pi to set whether or not the observation can be split. I don't know how uninterruptible fits into the picture.

Things are being sorted out, but I don't think just anybody can do it.  Your proposal will have to have been accepted with a "time critical" status or some such.  Probably for now you could just map the boolean to the enum value.

bryanmiller:
"Contiguous" is a bit unclear. Maybe use "No splitting"? That is at least more explicit about the impact.

PR #2793, opened 2026-07-31
Email notification of proposal submission

toddburnside:
See https://app.shortcut.com/lucuma/story/7549/email-notification-of-proposal-submission for details.

The deletion of columns from ProposalView is unrelated, but I figured we didn't really need yet another PR for something so small... These columns were removed from t_proposal some time ago, but ProposalView was not updated at that time.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/graphql/table/ProposalView.scala
+++ b/modules/service/src/main/scala/lucuma/odb/graphql/table/ProposalView.scala
@@ -21,8 +21,6 @@ trait ProposalView[F[_]] extends BaseMapping[F]:
     // those rows, so the placeholder is never rendered or discriminated upon.
     val ScienceSubtype  = col("c_gemini_science_subtype", science_subtype)
 
-    val Title           = col("c_title",    text_nonempty.opt)
-    val Abstract        = col("c_abstract", text_nonempty.opt)
     val Category        = col("c_category", tag.opt)
 
     val TooActivation   = col("c_too_activation", too_activation)
--- a/modules/service/src/main/scala/lucuma/odb/service/ProposalService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/ProposalService.scala
@@ -15,6 +15,7 @@ import grackle.syntax.*
 import lucuma.core.data.EmailAddress
 import lucuma.core.enums.ConsiderForBand3
 import lucuma.core.enums.ExchangePartner
+import lucuma.core.enums.GeminiCallForProposalsType
 import lucuma.core.enums.ObservationWorkflowState
 import lucuma.core.enums.Observatory
 import lucuma.core.enums.Partner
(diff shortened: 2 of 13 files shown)


cquiroz:
bye bye p1-monitor

Should emails use some sort of template?

toddburnside:
> Should emails use some sort of template?

Probably, but that is out of scope for this PR.

modules/service/src/main/scala/lucuma/odb/graphql/table/ProposalView.scala
```diff
@@ -21,8 +21,6 @@ trait ProposalView[F[_]] extends BaseMapping[F]:
     // those rows, so the placeholder is never rendered or discriminated upon.
     val ScienceSubtype  = col("c_gemini_science_subtype", science_subtype)
 
-    val Title           = col("c_title",    text_nonempty.opt)
```
cquiroz, line 24: Were these unused?

modules/service/src/main/scala/lucuma/odb/service/ProposalService.scala
```diff
@@ -183,6 +188,32 @@ object ProposalService {
...
+  private[odb] def timeRequestedText(cv: Option[CalculatedValue[CategorizedTimeRange]]): String =
+    def hours(t: TimeSpan): String = f"${t.toHours}%.2f"
+
+    cv.map(_.value).filter(_.max.programTime.toMicroseconds > 0L) match
+      case None                                               => "Not available"
+      case Some(r) if r.min.programTime === r.max.programTime => s"${hours(r.max.programTime)} hours"
+      case Some(r)                                            => s"${hours(r.min.programTime)} - ${hours(r.max.programTime)} hours"
+
+  /** Escapes the characters that would otherwise be markup in an html message. */
+  private[odb] def escapeHtml(s: String): String =
```
cquiroz, line 214: I'd bet something like this already exists on the std lib

cquiroz (approved):
LGTM

swalker2m (approved):
Looks brilliant. :+1:

modules/service/src/main/scala/lucuma/odb/graphql/table/ProposalView.scala
swalker2m: Yeah, it's mentioned in the PR description.  I think I'm the one who failed to remove them.

toddburnside: Yeah. They used to be columns in t_proposal, but awhile back we switched to using the program name and description from t_program instead and the columns were removed from t_proposal.

modules/service/src/main/scala/lucuma/odb/service/ProposalService.scala
toddburnside: I didn't find anything at first, but it looks like `scala.xml.Utility.escape()` will escape the basic characters that are escaped here as well as double quotes.

toddburnside: Otherwise we have to import something like apache commons

toddburnside: Actually, I'd need to add the scala-xml dependency. I'm inclined to think it isn't worth it, but I can do it.

swalker2m: Doesn't seem worth adding a dependency for this to me, fwiw.

cquiroz: i agree not worth it, mayb on http4s uri?

toddburnside: http4s doesn't seem to have anything, either.

PR #2794, opened 2026-07-31
Add support to create and read gmos mos

cquiroz:
This is a fairly large PR, however it's mostly based on the existing GMOS long slit implementation. In fact some of the code was refactored to be used by MOS and long slit.

The pull request includes not only creation and edition but also deletion and cloning.


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -1954,6 +1954,27 @@ input GmosCustomMaskInput {
   slitWidth: GmosCustomSlitWidth!
 }
 
+"""
+GMOS MOS custom mask input parameters.
+
+Distinct from `GmosCustomMaskInput`, which carries a *complete* mask (slit width
+required) as the FPU of a dynamic sequence step. 
+"""
```
swalker2m: ~Remind me why it is different for long slit with a custom mask?  It says `(slit width required)` as though that distinguishes `GmosMosCustomMaskInput` from `GmosCustomSlitWidth` but then below for the `slitWidth` field it says `Required when creating the mode`.  I wonder if we should just remove the `!` from the `GmosCustomMaskInput` `slitWidth` field to make it optional in both cases and then have only the one type.~

cquiroz: I think we can do that, in fact i wasn't sure if maybe we should make MOS just a particular case of longslit though there are enough differences in terms of ags and acquisition to justify separate modes.

swalker2m: I got a bit confused by this, but I think I understand now.  Long slit always uses a builtin FPU of course.  The existing input is for any GMOS dynamic step so it has to allow either kind of FPU but needs to know the slit width.

swalker2m: So I think we could get away with just using `GmosCustomMaskInput`, if we were willing to always require a `slitWidth` even if it isn't changing?  I wonder if that would be worth it to avoid the confusion of two inputs that are so similar.

cquiroz: let's do that we need slitWidth at least for ITC calls. I agree having two very similar inputs is confusing

swalker2m (approved):
It's too much to comb through in detail, but it all looks like fairly standard stuff.  The only doubt I have is about the two custom mask inputs.  I think I'd vote to just have one and always require the slit width.

PR #2795, opened 2026-08-03
Batch events

swalker2m:
Adds an `addEventBatch` mutation which accepts a list of execution events.  This can be used by Observe to save round trips to the database for a series of events. For example, all the events related to a single step could be sent at once.

## Caveat 1

Though it is permitted, it seems unlikely that grouping events for multiple steps would be wise.  Until an `END_STEP` event is received the database does not consider a step complete.  For example, if Observe executes a step without sending events and then requests the next atom, it will still contain the already executed step.

## Caveat 2

At the moment, the implementation simply loops through the events and adds them one at a time.  That means that the chief savings from using this method is the round trip and grackle processing time.  We could at some point attempt to produce a single SQL `INSERT` statement that does all the events at once, but it will require sorting out all the logic surrounding each insertion of the different kinds of events and making sure that it is still correct in this context.

### DIFF
--- a/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -122,6 +122,45 @@ type AddStepEventResult {
 }
 
 """
+A single event within an 'addEvents' batch.  Exactly one of the fields must be
+set, identifying the event type.  Unlike the singular mutations, every event in
+a batch must supply both a 'clientTime' and an 'idempotencyKey' (see 'addEvents').
+"""
+input AddEventInput {
+  dataset:  AddDatasetEventInput
+  sequence: AddSequenceEventInput
+  slew:     AddSlewEventInput
+  step:     AddStepEventInput
+}
+
+"""
+Input to the 'addEvents' mutation, recording a batch of execution events in a
+single request.
+
+Because a batch shares one database transaction, its events would otherwise all
+receive the same recorded time; every event must therefore supply its own
+'clientTime' (the mutation fails otherwise).  Every event must also supply a
+distinct 'idempotencyKey' so the batch can be retried safely.  All events must
+belong to the same observation.
+"""
+input AddEventsInput {
+  events: [AddEventInput!]!
+}
+
+"The result of adding a batch of events."
+type AddEventsResult {
+  "The new events that were added, in the order they were recorded."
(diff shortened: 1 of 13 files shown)


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -121,6 +121,46 @@ type AddStepEventResult {
   event: StepEvent!
 }
 
+"""
+A single event within an 'addEventBatch'.  Exactly one of the fields must be
+set, identifying the event type.  Unlike the singular mutations, every event in
+a batch must supply both a 'clientTime' and an 'idempotencyKey' (see
```
cquiroz, line 127: I guess this is enforced in the code not in the schema

modules/service/src/main/scala/lucuma/odb/graphql/input/AddEventBatchInput.scala
```diff
@@ -0,0 +1,35 @@
...
+  val Binding: Matcher[AddEventBatchInput] =
+    ObjectFieldsBinding.rmap:
+      case List(
+        AddEventBatchEntryInput.Binding.List("events", rEvents)
+      ) =>
+        rEvents.flatMap: events =>
+          NonEmptyList
+            .fromList(events)
+            .toRight("An 'addEventBatch' batch must contain at least one event.")
+            .flatMap: nel =>
```
cquiroz, line 25: Should we also enforce that all events have the same stepId?

cquiroz (approved):
LGTM

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
swalker2m: Yeah the input is explicitly validated in code.  We do the same thing in other places.  Unfortunately there is no interface type that can be extended for input objects or else it could be an array of the interface type with implementations for each kind.

modules/service/src/main/scala/lucuma/odb/graphql/input/AddEventBatchInput.scala
swalker2m: It should work if you wanted to send all the events for a sequence at once, even if it isn't advisable.  For one thing, you'd see no progress until the events are finally sent.  Since this is for service use, I didn't try to constrain it beyond what is strictly necessary though.

PR #2797, opened 2026-08-03
Fix three proposal time-request bugs

toddburnside:
Three independent bugs in how a proposal's time request — Gemini partner splits, or an assignment to a Keck/Subaru exchange partner community — is written and validated. Each commit stands alone; see the commit messages for detail.

Swapping splits for an exchange partner in one edit was rejected. check_proposal_time_request fires immediately and forbids a proposal from holding both, but updateProposal always wrote t_proposal before t_partner_split, so a valid edit setting exchangePartner alongside partnerSplits: null tripped it. Write order is now direction-dependent. The contract is unchanged: setting an exchange partner while leaving the splits in place is still an error.

Exchange requests ignored the per-community submission deadline. ProposalContext resolved the deadline to the call's default, ignoring the override introduced with t_gemini_cfp_exchange_partner in V1201 — which the API has been advertising all along as CallForProposalsExchangePartner.submissionDeadline. A call could show a Keck PI one deadline and enforce another. Now resolved from v_gemini_cfp_exchange_partner; identical for calls without an override.

A request naming a community the call doesn't invite was accepted. It silently borrowed the call's default deadline. Now rejected at submission with a message that says so, mirroring missingPartners for Gemini partners.

Testing

Five new tests, plus 227 existing across the proposal, CfP and program-user suites. Two pre-existing exchange-partner tests were asserting on a setup the ODB should never have accepted — a Regular Semester call with no exchangePartners declared, then a Keck request against it — and only passed because of the fallbacks removed here; their calls now invite the community they request.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/graphql/input/ProposalPropertiesInput.scala
+++ b/modules/service/src/main/scala/lucuma/odb/graphql/input/ProposalPropertiesInput.scala
@@ -5,6 +5,7 @@ package lucuma.odb.graphql.input
 
 import cats.syntax.all.*
 import grackle.syntax.*
+import lucuma.core.enums.ExchangePartner
 import lucuma.core.enums.Observatory
 import lucuma.core.enums.Partner
 import lucuma.core.enums.ScienceSubtype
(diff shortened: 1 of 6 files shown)


cquiroz:
> A request naming a community the call doesn't invite was accepted.

I'm a bit confused by this. i think it was done on the second commit,could you elaborate?

toddburnside:
> > A request naming a community the call doesn't invite was accepted.
> 
> I'm a bit confused by this. i think it was done on the second commit,could you elaborate?

If the PI was a Keck PI, a proposal for a CfP that didn't invite keck would be accepted without error. It was the third commit and if you look at that commit, you'll see a lot of additional detail in the message.

cquiroz (approved):
LGTM

PR #2799, opened 2026-08-03
Test for graphql diamond fragment issue

toddburnside:
Grackle is incorrectly reporting fragment cycles for shared fragments under some circumstances (see story).

A PR has been submitted to grackle to fix this: https://github.com/typelevel/grackle/pull/886

If/when this PR is accepted and a new grackle release is made, the failing test should pass.

FIxed in grackle 0.30.0.


hugo-vrijswijk (commented):
Is merging this still desired? Usually we don't really add tests for every downstream library bug. Useful to verify it was actually fixed though

toddburnside:
> Is merging this still desired? Usually we don't really add tests for every downstream library bug. Useful to verify it was actually fixed though

True. I guess I'll just close it. Easier than rebasing several times. 😆

PR #2800, opened 2026-08-03
GMOS Twilight sequence should be unguided

toddburnside:


cquiroz (approved):
Seems quite logical, we missed it when generating them originally. LGTM

PR #2801, opened 2026-08-03
Add mos acq types

cquiroz:
We need to record a flag to tell which type of acquisition we'll use and that will generate different type of sequences.


modules/service/src/main/resources/db/migration/V1237__gmos_mos_acquisition_type.sql
```diff
@@ -0,0 +1,18 @@
...
+
+-- Reconstruct the views
+DROP VIEW v_gmos_north_mos;
+DROP VIEW v_gmos_south_mos;
+
+CREATE VIEW v_gmos_north_mos AS
+  SELECT m.* FROM t_gmos_north_mos m;
+
+CREATE VIEW v_gmos_south_mos AS
+  SELECT m.* FROM t_gmos_south_mos m;
```
swalker2m, line 18: These views don't add anything not already in the table.  I guess we're anticipating adding something there though?

swalker2m (approved):
Looks good. :+1:

modules/service/src/main/resources/db/migration/V1237__gmos_mos_acquisition_type.sql
cquiroz: Right,  just followed the convention of using views for modes

PR #2805, opened 2026-08-04
MOS science sequence generation

cquiroz:
Add support to generate MOS science sequences. They are almost identical to long slit except for the fpu thus we also have refactored the common code for sequence generation. Also by default offsets are empty

Acquisition sequences will be add in a separate PR

image


swalker2m (approved):
Looks good. :+1:

PR #2806, opened 2026-08-05
Load ODB schema with Schema stitcher loader

hugo-vrijswijk:
Verifies the schema at compile-time, moves some unsafe code out of OdbMapping, and adds safe handling and logging for loading on startup

### DIFF
--- a/modules/binding/src/main/scala/lucuma/odb/graphql/schema/SchemaStitcherMacros.scala
+++ b/modules/binding/src/main/scala/lucuma/odb/graphql/schema/SchemaStitcherMacros.scala
@@ -19,6 +19,15 @@ private object SchemaStitcherMacros:
   inline def fromResources(name: String) =
     ${ SchemaStitcherMacros.fromResourcesImpl('name) }
 
+  private val chunkSize = 16384
+
+  /** Splits `s` into constants small enough for the constant pool and rejoins them at runtime. */
+  private def stringExpr(s: String)(using Quotes): Expr[String] =
+    if s.length <= chunkSize then Expr(s)
+    else
+      val chunks = Expr.ofSeq(s.grouped(chunkSize).toSeq.map(Expr(_)))
+      '{ $chunks.mkString }
+
   /**
    * Macro to do some compile time checking for the schema. It will:
    *   - Check that the resource exists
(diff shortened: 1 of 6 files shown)


modules/binding/src/main/scala/lucuma/odb/graphql/schema/SchemaStitcherMacros.scala
```diff
@@ -19,6 +19,15 @@ private object SchemaStitcherMacros:
   inline def fromResources(name: String) =
     ${ SchemaStitcherMacros.fromResourcesImpl('name) }
 
+  private val chunkSize = 16384
+
+  /** Splits `s` into constants small enough for the constant pool and rejoins them at runtime. */
```
cquiroz, line 24: Was the macro breaking?

modules/obscalc/src/main/scala/lucuma/odb/obscalc/Main.scala
```diff
@@ -267,6 +267,7 @@ object CalcMain extends MainParams:
       _          <- Resource.eval(pool.use: s =>
                       Services.asSuperUser(UserService.fromSession(s).canonicalizeUser(user))
                     )
+      schema     <- Resource.eval(OdbMapping.loadSchema[F])
```
cquiroz, line 270: Why does obscalc need the schema?

cquiroz:
Why do we want to load the schema via schemastitcher?

modules/binding/src/main/scala/lucuma/odb/graphql/schema/SchemaStitcherMacros.scala
hugo-vrijswijk: Yes, the schema is too big to fit in a single string constant 😅

modules/obscalc/src/main/scala/lucuma/odb/obscalc/Main.scala
hugo-vrijswijk: It calls the OdbMapping

hugo-vrijswijk:
> Why do we want to load the schema via schemastitcher?

It gives compile-time schema validation and errors, removes some unsafe code in OdbMapping and adds logging through the logger. There's no stitching, so the name is maybe a little confusing

cquiroz (approved):
LGTM

PR #2807, opened 2026-08-05
Add mos acquisition sequence generation

cquiroz:
The acquisition sequence for MOS will depend on the acquisition type. With `Maskn` we take n steps with the mask. with `MaskOut` mode we take first an image without the mask and then continue with the mask.

Unlike long slit, we always use full frame ROI. And the ETM is always time and count. 

MaskIn:
image

MaskOut:
image


swalker2m (approved):
:+1:

PR #2808, opened 2026-08-05
Shortcut 9783: Add `Interrupting` ToO activation value, proposal ToO ceiling, observation ToO activation

swalker2m:
This PR makes several Target of Opportunity properties updates (~and relies on as of yet [unmerged changes in lucuma-core](https://github.com/gemini-hlsw/lucuma-core/pull/1539)~).

### Adds Observation-level ToO Activation Property

Adds a ToO activation property, `tooActivation`, to the `SchedulingConstraints` of observations.  This makes it possible to identify how each observation should behave wrt ToO triggers.  As mentioned below, the ToO activation selected for an observation cannot exceed what is in an _accepted_ proposal.

### Adds Proposal-level ToO Activation _Ceiling_ 

Changes the `tooActivation` field on the proposal types that support it to the familiar triplet of default/explicit/effective values and treats it as a cap on observation-level ToO activation:

* **`defaultTooActivationCeiling: TooActivation!`**. The default is computed from the observation-level ToO activations.  It is always the _most disruptive_ activation amongst the program's observations (`NONE` < `STANDARD` < `RAPID` < `INTERRUPTING`).
* **`explicitTooActivationCeiling: TooActivation`**. This is optional before the proposal is accepted, but explicitly set (to the default) upon acceptance if not already defined.  Once accepted, only NGO or staff or better may change this value.
*  **`tooActivationCeiling: TooActivation!`**. This is the effective value, which is the explicit value when defined but otherwise the default. Any observation whose ToO activation exceeds the effective ceiling for the proposal cannot be made `READY`.

### Changes Execution Requirement Handling

Splits the observation `executionRequirement` field into the familiar triplet of default/explicit (optional)/effective values:

* **`defaultExecutionRequirement: ExecutionRequirement!`**.  Here the default depends on the ToO activation. `RAPID` and `INTERRUPTING` ToOs default to an `UNINTERRUPTIBLE` execution requirement while `NONE` and `STANDARD` default to `UNCONSTRAINED`.
* **`explicitExecutionRequirement: ExecutionRequirement`**. The explicit value requested by the PI, if any.  The explicit value is respected if it is a tighter (more disruptive) value than the default.  In other words, the default is a _floor_ on the execution requirement. For example, a `RAPID` ToO has a default execution requirement of `UNINTERRUPTIBLE` so any explicit setting less than this is ignored. A `STANDARD` ToO has a default execution requirement of `UNCONSTRAINED` so any explicit setting above this is respected.
* **`executionRequirement: ExecutionRequirement!`**.  The effective value the scheduler should use.  Explicit when defined, the default otherwise, with the caveat that the default value is a hard floor on the effective value.


toddburnside (approved):
Looks very confusing and complicated to me, but I don't see any issues with it.

swalker2m:
> Looks very confusing and complicated to me, but I don't see any issues with it.

TLDR

* There’s a ToO status at the observation level.
* It can limit how permissive the execution requirement is.
* There’s a derived ToO status at the proposal level, until it is accepted and then it is fixed and is a hard limit.

PR #2810, opened 2026-08-05
Add GNIRS imaging acquisition customization

toddburnside:
GNIRS spectroscopy lets the user override the ITC brightness classification and its inputs — acquisition type, filter, exposure time mode, coadds and (for Faint) the sky offset. GNIRS imaging had none of that: its acquisition type was always ITC-classified and its Faint field offset was hardcoded to (0", +10").

Adds the same customization to GNIRS imaging:

- `AcquisitionConfig` moves out of `gnirs/spectroscopy/Config.scala` to `sequence/gnirs/`, shared by both modes. `selectedFilter` is generalized to take the mode-specific automatic filter, since spectroscopy derives it from the central wavelength while imaging uses the first (wavelength-ordered) science filter.
- The GraphQL acquisition input, including its three validations, moves to `GnirsAcquisitionInput`; both mode inputs alias it.
- `gnirs/imaging/Acquisition` resolves the mode, filter, coadds and Faint sky offset from the config, falling back to the previous behavior when nothing is set. The VeryBright field image stays on H (Order4).
- The imaging branch of `GeneratorParamsService` passes the explicit acquisition filter, ETM and coadds to the ITC, and only runs the two-pass classification when both the mode and filter are automatic.
- V1242 adds the five acquisition columns to `t_gnirs_imaging` with the same both-or-neither and faint-only CHECKs as the spectroscopy table, rebuilds `v_gnirs_imaging`, and requires an acquisition ETM row.

Also fixes a latent bug in `GnirsImagingService.update`: editing the filter list ran `deleteCurrentFiltersAndEtms`, which drops every acquisition-role ETM, then re-derived one from the science ETM. That was harmless while the acquisition ETM was invisible, but it would have silently discarded a user's setting, so imaging now uses a science-only variant of that statement.

Claude-Session: https://claude.ai/code/session_01Lyt6Er1f7k4DDcNBoZEUGS


cquiroz (approved):
LGTM

PR #2814, opened 2026-08-06
Add test for replacing gmos MOS sequences

cquiroz:
materialization is available already for MOS but we didn't have tests, some are included on this PR


toddburnside (approved):
LGTM

PR #2820, opened 2026-08-07
Trigger on Ready

swalker2m:
This PR changes the way that ToO observations are triggered, favoring the `ocs` way of inferring it from state changes.

* Removes the explicit trigger mutations `requestTooTrigger`, `acceptTooTrigger` and `withdrawTooTrigger`.  Instead, triggering a ToO is inferred from seeing (via a _database_ trigger) its workflow state transition to `READY`.  That action records a new row in the `t_too_trigger` table and sends events.  Withdrawing is done similarly by changing the workflow state back to `DEFINED`.  There is no explicit acceptance, though the observer is free to observe (or not) a rapid or interrupting ToO whenever the scheduler suggests he or she do so.  They may explicitly `decline` a trigger though, with a reason, and doing so moves the workflow state back to `DEFINED` so that it may be triggered again in the future.

* Flags an observation carrying an opportunity placeholder as `Undefined` when it declares no ToO activation, or when it has been set `READY` without the placeholder having been resolved.


swalker2m:
I had to add `expected-breaking-change` because I deleted a few ToO mutations, renamed one, and changed the ToO trigger states.  Nobody is using this yet though so nothing should, you know, actually _break_.  😬

toddburnside (approved):
Seems good to me.

PR #2825, opened 2026-08-10
Add a mask name entry for mask attachments

cquiroz:
Observe needs a mask name to tell the instrument with a specific format.

In OCS this is solved because staff adds the mask but in gpp anybody can upload a mask. This PR adds a field for the mask name that, for the moment, is derived from the file name, but should in the future be read directly from the mask file. 

The field is read only


toddburnside (approved):
Seems fine to me.

PR #2826, opened 2026-08-10
Shortcut 9812: Support multiple GNIRS central wavelengths

toddburnside:
NOTE: One bit I'm unsure of is in Acquisition.scala. It is using the first (lowest) wavelength and coadds for something, and I'm not sure if that is correct.

What follows is claude's commit message:
The GNIRS higher-resolution modes cover only part of the spectral band, so observers take spectra at several central wavelengths in one observation. Unlike GMOS wavelength dithers, which all contribute to a single S/N, each central wavelength here is a separate configuration: its own exposure time mode, coadds, ITC calculation and smart calibrations.

GnirsSpectroscopy's single centralWavelength / exposureTimeMode / coadds become a list of (centralWavelength, exposureTimeMode, coadds), stored in a new t_gnirs_spectroscopy_wavelength child table modeled on t_gnirs_imaging_filter.  The ITC fans out one call per wavelength, mirroring the per-filter imaging fan out, and returns results keyed by wavelength through the previously unused GNIRS_SPECTROSCOPY ItcType.  The science sequence runs each wavelength as a contiguous segment closed by its own flats and arcs, round-robining between them as the GMOS wavelength-dither generator does; the visit-length budget is shared out, so a single-wavelength observation generates exactly the sequence it did before.

Two adjustments outside GNIRS were required.  check_etm_consistent no longer fixes the science exposure time mode count for gnirs_long_slit / gnirs_ifu, since those are now per wavelength and per row version, as for the imaging modes.  PerScienceObservationCalibrationsService now updates the telluric exposure time modes in place rather than deleting and reinserting them: the wavelength rows reference their ETM with ON DELETE CASCADE, so replacing it would take the observing mode's wavelengths with it.

### DIFF
--- a/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -11332,11 +11332,48 @@ type GnirsSpectroscopyAcquisition {
 }
 
 """
+A GNIRS spectroscopy science configuration: a central wavelength together with
+the exposure time mode and coadds that apply at that wavelength.
+"""
+type GnirsSpectroscopyWavelength {
+
+  centralWavelength: Wavelength!
+
+  """
+  Exposure time mode for this central wavelength.
+  If not specified, it is taken from the observation's requirements.
+  """
+  exposureTimeMode: ExposureTimeMode!
+
+  coadds: PosInt!
+
+}
+
+"""
+Defines a GNIRS spectroscopy central wavelength to use along with its exposure
+time mode and coadds.
+"""
+input GnirsSpectroscopyWavelengthInput {
+
+  centralWavelength: WavelengthInput!
+
+  """
+  Exposure time mode for this central wavelength.
+  If not specified, it is taken from the observation's requirements.
+  """
+  exposureTimeMode: ExposureTimeModeInput
+
+  """
+  Coadds for this central wavelength.  Defaults to 1 when not specified.
+  """
+  coadds: PosInt
+
+}
+
+"""
 GNIRS Spectroscopy mode (long slit or IFU, distinguished by the FPU).
 """
 type GnirsSpectroscopy {
-
-  exposureTimeMode: ExposureTimeMode!
 
   grating: GnirsGrating!
   explicitGrating: GnirsGrating
@@ -11346,8 +11383,18 @@ type GnirsSpectroscopy {
   explicitPrism: GnirsPrism
   initialPrism: GnirsPrism!
 
-  centralWavelength: Wavelength!
-  initialCentralWavelength: Wavelength!
+  """
+  The central wavelengths at which spectra are taken, in increasing wavelength
+  order.  Each one is a separate configuration with its own exposure time mode,
+  coadds, ITC calculation and calibrations.
+  """
+  centralWavelengths: [GnirsSpectroscopyWavelength!]!
+
+  """
+  The central wavelengths as they were configured when the observing mode was
+  created.
+  """
+  initialCentralWavelengths: [GnirsSpectroscopyWavelength!]!
 
   camera: GnirsCamera!
   initialCamera: GnirsCamera!
@@ -11366,8 +11413,6 @@ type GnirsSpectroscopy {
 
   filter: GnirsFilter!
   initialFilter: GnirsFilter!
-
-  coadds: PosInt!
 
   decker: GnirsDecker!
   explicitDecker: GnirsDecker
@@ -11482,9 +11527,12 @@ Edit or create GNIRS Spectroscopy configuration (long slit or IFU).
 """
 input GnirsSpectroscopyInput {
 
-  exposureTimeMode: ExposureTimeModeInput
-
-  coadds: PosInt
+  """
+  The central wavelengths at which spectra are taken.  Required on create, and
+  must contain at least one entry with no duplicated wavelength.  When given on
+  edit, it replaces the existing list wholesale.
+  """
+  centralWavelengths: [GnirsSpectroscopyWavelengthInput!]
 
   filter: GnirsFilter
 
@@ -11499,8 +11547,6 @@ input GnirsSpectroscopyInput {
   grating: GnirsGrating
 
   prism: GnirsPrism
-
-  centralWavelength: WavelengthInput
 
   explicitDecker: GnirsDecker
 
@@ -14509,6 +14555,15 @@ type ItcGnirsImagingResultSet {
 }
 
 """
+Combines a GNIRS spectroscopy central wavelength with an `ItcResultSet`. In
+other words, ITC results for all targets but a single central wavelength.
+"""
+type ItcGnirsSpectroscopyResultSet {
+  centralWavelength: Wavelength!
+  results:           ItcResultSet!
+}
+
+"""
 ITC results for a particular observation, including relevant instrument
 configurations and targets.  There are specific instances for each `ItcType`.
 """
@@ -14561,6 +14616,16 @@ GNIRS imaging ITC results.  Here each filter is paired with its result set.
 type ItcGnirsImaging implements Itc {
   itcType: ItcType!
   gnirsImagingScience: [ItcGnirsImagingResultSet!]!
+}
+
+"""
+GNIRS spectroscopy ITC results.  Each central wavelength is a separate
+configuration and is paired with its own result set.
+"""
+type ItcGnirsSpectroscopy implements Itc {
+  itcType: ItcType!
+  acquisition: ItcResultSet!
+  gnirsSpectroscopyScience: [ItcGnirsSpectroscopyResultSet!]!
 }
 
 """
--- a/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -11335,7 +11335,7 @@ type GnirsSpectroscopyAcquisition {
 A GNIRS spectroscopy science configuration: a central wavelength together with
 the exposure time mode and coadds that apply at that wavelength.
 """
-type GnirsSpectroscopyWavelength {
+type GnirsCentralWavelengthConfig {
 
   centralWavelength: Wavelength!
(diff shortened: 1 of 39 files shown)


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -11332,12 +11332,49 @@ type GnirsSpectroscopyAcquisition {
...
 
 """
-GNIRS Spectroscopy mode (long slit or IFU, distinguished by the FPU).
+A GNIRS spectroscopy science configuration: a central wavelength together with
+the exposure time mode and coadds that apply at that wavelength.
 """
-type GnirsSpectroscopy {
+type GnirsSpectroscopyWavelength {
+
+  centralWavelength: Wavelength!
```
cquiroz, line 11340: Could this instead be part of the ETM and carry multiple etms?

toddburnside: Well, it is only applicable to GNIRS spectroscopy (at the moment, at least) so I think it would be messy to pollute ETM with it. Also, this is analogous to how we handled the multiple filters for imaging.

cquiroz: Right, i was thinking that's why we added multiple etms, if central wavelength were in ETM as an optional attribute maybe it would be simpler?
I'm not sure if i have the full picture though

toddburnside: We'd also need coadds. Then for imaging we'd need instrument specific filters. It seems like it might be better to keep the ExposureTimeMode as an exposure time mode. Not an exposure time mode plus extra info.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -11332,12 +11332,49 @@ type GnirsSpectroscopyAcquisition {
 }
 
 """
-GNIRS Spectroscopy mode (long slit or IFU, distinguished by the FPU).
+A GNIRS spectroscopy science configuration: a central wavelength together with
+the exposure time mode and coadds that apply at that wavelength.
 """
-type GnirsSpectroscopy {
+type GnirsSpectroscopyWavelength {
```
cquiroz: Still this type seems odd to me, it is not really a `Wavelength` as it contains coadds and etms too. maybe the name could be improved?

toddburnside: I guess the reference here are the ones named like `Flamingos2ImagingFilter` both here and in explore. Maybe `GnirsSpectroscopyCentralWavength`? Or `GnirsSpectroscopyCentralWavelengthAndEtmAndCoadds`? Seems a bit long. Any other suggestions?

cquiroz: the comment says
`A GNIRS spectroscopy science configuration`
GnirsSpectroscopyScienceConfiguration

though i guess configuration is a broader concept

toddburnside: `GnirSpectroscopyWavelengthConfig`?

cquiroz: Maybe, what do you think?

toddburnside: Doesn't really matter to me. I can have claude update the odb and explore PRs if we want that.

cquiroz: That sounds fine. it just seemed odd to call a type XXWavelength and it contains several things besides a wavelength

toddburnside: How about `GnirsCentralWavelengthConfig`? It only applies to spectroscopy, but maybe that doesn't need to be in the name.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -11482,9 +11527,12 @@ Edit or create GNIRS Spectroscopy configuration (long slit or IFU).
 """
 input GnirsSpectroscopyInput {
 
-  exposureTimeMode: ExposureTimeModeInput
-
-  coadds: PosInt
+  """
+  The central wavelengths at which spectra are taken.  Required on create, and
+  must contain at least one entry with no duplicated wavelength.  When given on
```
cquiroz, line 11532: is this enforced on the input bindings? i think you can't enforce `at least one` in the schema, can you?

modules/schema/src/main/scala/lucuma/odb/data/Itc.scala
```diff
@@ -95,13 +96,27 @@ sealed trait ItcScience:
...
+   * wavelength).  Saturates rather than overflowing: the value only feeds the
+   * sequence-size guard, which any saturated total will trip anyway.
+   */
+  private def sumExposureCounts[A](
+    science: NonEmptyMap[A, Zipper[ItcResult]]
+  ): PosInt =
+    PosInt.unsafeFrom:
+      science.foldLeft(0): (cnt, z) =>
+        val n = z.focus.value.exposureCount.value
+        if cnt > Int.MaxValue - n then Int.MaxValue else cnt + n
```
cquiroz, line 110: i guess it could be clamped on the final result

modules/sequence/src/main/scala/lucuma/odb/sequence/data/ItcInput.scala
```diff
@@ -135,6 +135,53 @@ object ItcInput:
...
+        // differ only in an extra central wavelength must not share a cache key.
+        a.science.toList.foreach: params =>
+          bld.addAll(params.hashBytes)
+        bld.addAll(hashTargets(a.blindOffset.fold(a.targets)(_ :: a.targets)))
+        bld.addAll(a.signalToNoiseTargetId.hashBytes)
+        bld.addAll(a.gnirsAcqAutoClassify.hashBytes)
+        bld.result()
+
+  val gnirsSpectroscopy: Prism[ItcInput, ItcInput.GnirsSpectroscopy] =
+    Prism[ItcInput, ItcInput.GnirsSpectroscopy] {
```
cquiroz: `GenPrism`?

modules/sequence/src/main/scala/lucuma/odb/sequence/gnirs/spectroscopy/Science.scala
```diff
@@ -164,62 +170,117 @@ object Science:
...
+     * expansion, since the flat and arc lookups are keyed on the wavelength.
+     */
+    def computeAll[F[_]: Monad](
+      config:   Config,
+      times:    NonEmptyList[(ScienceWavelength, IntegrationTime)],
+      static:   GnirsStaticConfig,
+      expander: SmartGcalExpander[F, GnirsStaticConfig, GnirsDynamicConfig],
+      calRole:  Option[CalibrationRole]
+    ): EitherT[F, String, NonEmptyList[StepDefinition]] =
+      times.traverse((sw, t) => compute(config, sw, t, static, expander, calRole))
```
cquiroz, line 199: nice

modules/sequence/src/main/scala/lucuma/odb/sequence/gnirs/spectroscopy/Science.scala
```diff
@@ -164,62 +170,117 @@ object Science:
...
+  )
+
+  /**
+   * Generates the science sequence across every central wavelength.
+   *
+   * Each wavelength is a separate configuration whose flats and arcs are looked
+   * up by wavelength, so its exposures run as a contiguous segment followed by
+   * its own calibrations.  The segments are then round-robined -- λ1, λ2, ... λN,
+   * λ1, ... -- until every wavelength has met its goal, following the GMOS
+   * wavelength-dither generator.  Running each wavelength to completion instead
```
cquiroz, line 222: Does it mean we have a single common acquisition and then science at each central wavelength?

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
swalker2m: I suppose you could require a head and accept an optional tail, but I think it's probably less confusing to just document and validate.

modules/schema/src/main/scala/lucuma/odb/data/Itc.scala
swalker2m: "Saturates rather than overflowing" means that the exposure count is artificially capped at `MaxInt` so the exposure times are correspondingly longer?  Anyway, if the exposure count is that big it probably isn't happening regardless.

swalker2m (approved):
This seems reasonable.  It's like having an ETM per filter in imaging but applied to wavelengths instead.  The suggested name change for the wavelength config type seems like a good idea to me.

modules/sequence/src/main/scala/lucuma/odb/sequence/gnirs/spectroscopy/Science.scala
toddburnside: Yes.

modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
toddburnside: Yes, this is validated in GnirsSpectroscopyInput. I don't think there is an `at least one` annotation for graphql, so this sort of validation is the same as we do elsewhere.

modules/schema/src/main/scala/lucuma/odb/data/Itc.scala
toddburnside: I think claude is just very thorough about not overflowing Ints. I guess it could be done on the final value if the accumulator was a Long, but I suspect that would not be much of an optimization.

PR #2830, opened 2026-08-11
SS-9725 initial support for warnings take2

tpolecat:
This factors out the validation code from ObservationWorkflowService and moves each validator into its own source file. It's completely mechanical. I did a while back but I got too out of sync with main so I did it again. Fingers crossed.


tpolecat:
Since this is mechanical I'm just going to merge when it passes.

PR #2832, opened 2026-08-11
Fix SSO users query filtered by orcidId

toddburnside:
WhereUser.orcidId is declared as WhereOptionString, so the predicate bound a String constant. Grackle binds Const values with the column's own encoder, and the SSO's orcid_id column is typed as OrcidId, so EQ/NEQ/IN/NIN threw a ClassCastException at execution time, producing a 500 with an empty body.

Add WhereOptionString.bindingAs, which parses the equality-style values with a supplied matcher, and use it with OrcidIdBinding in the SSO. LIKE/NLIKE remain string comparisons, which Grackle always encodes as strings.

### DIFF
--- a/modules/binding/src/main/scala/lucuma/odb/graphql/binding/WhereOptionString.scala
+++ b/modules/binding/src/main/scala/lucuma/odb/graphql/binding/WhereOptionString.scala
@@ -5,6 +5,7 @@ package lucuma.odb.graphql
 
 package binding
 
+import cats.Eq
 import cats.syntax.all.*
 import eu.timepit.refined.cats.*
 import grackle.Path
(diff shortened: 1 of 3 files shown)


modules/binding/src/main/scala/lucuma/odb/graphql/binding/WhereOptionString.scala
```diff
@@ -17,13 +18,23 @@ import lucuma.odb.graphql.binding.*
...
+    bindingAs(path, NonEmptyStringBinding)
+
+  /**
+   * As `binding`, but parses the equality-style values (EQ, NEQ, IN, NIN) with the
+   * given matcher. Use this when the mapped column's Scala type is not `String`,
+   * since the constants are bound into the query with the column's own codec.
+   * LIKE and NLIKE remain string comparisons, which is always correct because
+   * Grackle encodes them with a string encoder regardless of the column codec.
+   */
+  def bindingAs[A: Eq](path: Path, value: Matcher[A]): Matcher[Predicate] =
```
cquiroz, line 30: Sounds like this could be generic enough to use in other places too

cquiroz (approved):
LGTM

modules/binding/src/main/scala/lucuma/odb/graphql/binding/WhereOptionString.scala
toddburnside: Yep.

PR #2834, opened 2026-08-12
Derive the mask name from the ODF file name convention

cquiroz:
This PR updates how we derive mask names from the files and also requires filenames to follow either ocs or gpp conventions

We need ocs ones to do testing in the interim period

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/service/AttachmentFileService.scala
+++ b/modules/service/src/main/scala/lucuma/odb/service/AttachmentFileService.scala
@@ -106,11 +106,6 @@ object AttachmentFileService {
       // does not contain the dot.
       def extName: Option[NonEmptyString] =
         NonEmptyString.from(Path(fileName.value.value).extName.drop(1).toLowerCase).toOption
-
-      // the whole name when there is no extension to strip.
-      def baseName: NonEmptyString =
-        val name = fileName.value.value
-        NonEmptyString.from(name.stripSuffix(Path(name).extName)).getOrElse(fileName.value)
   }
 
   extension [F[_], A](fe: F[Either[AttachmentException, A]])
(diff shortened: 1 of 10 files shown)


toddburnside (approved):
LGTM

modules/service/src/main/scala/lucuma/odb/service/AttachmentFileService.scala
```diff
@@ -151,14 +146,35 @@ object AttachmentFileService {
...
+   * the program reference with its dashes removed,
+   * `GYYYY(A|B)PPPP<type>-XX_ODF.fits`.
+   *
+   * The mask name is the file name with the `_ODF.fits` suffix removed.
+   * Names are matched case-insensitively but stored upper case, so the identifier
+   * handed to observe is canonical however the file was named.
    */
+  private val OcsMaskFileName =
+    raw"(?i)(G[NS]\d{4}[AB](?:ENG|CAL|COM|DD|DS|SV|LP|FT|Q|C)\d{3}-\d{2})_ODF\.fits".r
+  private val GppMaskFileName = raw"(?i)(G\d{4}[AB]\d{4}[CDFLPQSV]-\d{2})_ODF\.fits".r
```
swalker2m, line 160: If the idea is to match a `ProgramReference` with an `_ODF.fits` suffix, maybe we should use `ProgramReference.fromString` instead of a separate regex?  Of if it is only science program references there's a parser for `ProgramReference.program`.

swalker2m (approved):
Looks great.  We might want to consider using the `ProgramReference` / `ObservationReference` parsers if possible so that we don't have the parsing in two independent places.

modules/service/src/main/scala/lucuma/odb/service/AttachmentFileService.scala
cquiroz: Good idea

cquiroz: Actually this is not correct, the mask names don't follow exactly the program reference name, from the ICD:

```
The resulting MOS MDF name is based on a compressed form of the program id: GYYYYSNNNNT-NN, where NN is a unique two-digit running number within a program.

Examples:
G2026A0007Q-03
```

cquiroz:
The regex way is correct as mask names have a different format, though perhaps we should add a parser to lucuma-core for the new format
I'll merge as it is now

PR #2835, opened 2026-08-12
Move GNIRS imaging coadds into the per-filter list

toddburnside:
Coadds were a single mode-level field applied to every filter, while each filter already carried its own exposure time mode.  That is wrong in two ways: the legacy OCS ITC discards the caller's coadds entirely under a signal-to-noise ETM (ImagingMethodExptime hardcodes 1 and GnirsRecipe recomputes its own), and one value cannot serve a filter list whose members mix ETMs.

So coadds now live on GnirsImagingFilter, forced to 1 when that filter's ETM is signal-to-noise, and the mode-level field is gone.  This mirrors the GNIRS spectroscopy central wavelength configurations.

The read path moves off the shared s_filter_exposure_time_mode composite type (which has no coadds slot and is shared with GMOS/Flamingos-2) to the row-per-key join used by GNIRS spectroscopy.

### DIFF
--- a/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -11163,6 +11163,12 @@ type GnirsImagingFilter {
   """
   exposureTimeMode: ExposureTimeMode!
 
+  """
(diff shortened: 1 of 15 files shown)


swalker2m (approved):
I can find no faults.  Looks great. :+1:

cquiroz (approved):
LGTM

PR #2836, opened 2026-08-12
Postgres error setting reverting imaging modes

cquiroz:
Accepting or reverting a configuration fails: the mutation reads the full observingMode, and grackle compiles it into a single SQL statement with ~1720 columns — over Postgres's 1664-entry target-list limit.

The width came from the offset-generator mappings, repeated at 16 paths (4 imaging instruments × 2 variants × object/sky): each generator sub-shape self-joined v_offset_generator (a fresh aliased copy per join), and every nested singleton (center, cornerA/B, p, q, size) declared 2–3 key columns, hoisted per level per path.

The fix is on the grackle mappings: map the sub-shapes without a join — they live on the same row, and nullability still comes from the synthetic CASE WHEN keys — and keep a single key column on the singletons. The statement drops from 1719 to 1391 entries.

ShortCut_9893 replays the failing selection and now passes; all imaging, execution, and clone suites unchanged. Verified in dev using tracing. The previously failing mutation completes in ~3s.

We can also do some improvements on the front end but the fixes on the backend are also quite critical


modules/service/src/main/scala/lucuma/odb/graphql/mapping/OffsetMapping.scala
```diff
@@ -111,25 +111,29 @@ trait OffsetMapping[F[_]] extends EnumeratedOffsetView[F]
...
-      offsetComponentMappingAtPath(RandomPath / "q", TelescopeConfigGeneratorView.CenterOffsetQ, TelescopeConfigGeneratorView.Random.ObservationId, TelescopeConfigGeneratorView.Random.Role),
-
-      offsetMappingAtPath(SpiralPath, TelescopeConfigGeneratorView.Spiral.ObservationId, TelescopeConfigGeneratorView.Spiral.Role),
-      offsetComponentMappingAtPath(SpiralPath / "p", TelescopeConfigGeneratorView.CenterOffsetP, TelescopeConfigGeneratorView.Spiral.ObservationId, TelescopeConfigGeneratorView.Spiral.Role),
-      offsetComponentMappingAtPath(SpiralPath / "q", TelescopeConfigGeneratorView.CenterOffsetQ, TelescopeConfigGeneratorView.Spiral.ObservationId, TelescopeConfigGeneratorView.Spiral.Role),
+      // Singleton offsets under the telescope-config generators carry a
+      // single key column: every key = true column is hoisted once per
+      // object level and per generator path in the generated SQL, which
+      // contributes to the Postgres 1664-column target-list limit (sc-9893).
```
swalker2m, line 117: Say what?

swalker2m (approved):
I don't understand it, but if it works ¯\_(ツ)_/¯

modules/service/src/main/scala/lucuma/odb/graphql/mapping/OffsetMapping.scala
cquiroz: Don't expect me to explain grackle but as I understand it we are re-reading the same column over an over with the kind of joins on the previous mapping. we mostly discard those columns so they are deadweight but hit a postgress limit on how many columns you can get

PR #2844, opened 2026-08-14
Acquire GNIRS thermal-IR science on the blue camera

toddburnside:
A GNIRS observation at 5µm produced an ITC plot but could not generate a sequence, failing with "No Gnirs spectroscopy filter available for wavelength: 5000000". The acquisition filter was derived by strict wavelength containment, and the acquisition filters cover only 1.030-2.490µm plus PAH's narrow 3.266-3.321µm, so every M-band config and every L-band config outside the PAH band had no acquisition filter. The derivation is in GeneratorParamsService, so it aborted before any sequence code ran.

The lookup is now total in lucuma-core, which is most of the fix. On this side:

  - The acquisition steps use the blue camera at the science pixel scale, per the GNIRS instrument scientists (Config.acquisitionCamera). Short and long are never mixed. The HR-IFU alignment flat keeps the science camera, being taken at the science configuration. The ITC acquisition parameters use the same camera as the sequence, as they must.
  - The two filter-derivation sites collapse now that the lookup cannot fail. GNIRS imaging needs no camera change: its filters are all blue-camera.

A manual override to acquire on the red camera is deliberately left for later; it needs a new column, GraphQL input/output, mapping and Explore UI.

Tests cover L and M on both pixel scales, asserting H on the blue counterpart, plus the reported fully automatic 5µm case and a blue-camera config to show it is unaffected. The SmartGCAL fixture gains Deep rows, which the red cameras default to and which no test previously needed.


cquiroz (approved):
LGTM

PR #2848, opened 2026-08-14
Expose phase0 custom slit width

cquiroz:
Phase0 already contains the custom slit with in pixels. We were before calculating it from geometry, but it is not necessary as we can read it directly. 

GMOS is also updated. The problem is not really present for gmos right now because custom slit widths match the built-in sizes, but for uniformity it is better to expose the custom width in the schema as well


toddburnside (approved):
LGTM

PR #2850, opened 2026-08-14
Target-of-Opportunity resolution and SchedulingMode

swalker2m:
An insufferable ODB companion to the [core PR](https://github.com/gemini-hlsw/lucuma-core/pull/1549). It will glaze your eyes but it comes down to supporting the same two things as the core PR:

* Scheduling Mode and derived ToO Activation
* "Forever" ToOs that resolve with real target tracking rather than transmogrify into sidereal or nonsidereal

See the last file in the commit, an [overview Markdown document](https://github.com/gemini-hlsw/lucuma-odb/blob/bdd42b839ef9e08c2520b5739782c29491c6aabb/notes/too-scheduling-overview.md) if you inexplicably want more context.

##  Claude Commit Message

Replaces the two observation-level ToO properties -- `ExecutionRequirement` and a declared `TooActivation`, coupled by a floor where raising one silently raised the other -- with a single `SchedulingMode`, and makes an opportunity target keep its identity after it is resolved.  The two changes are inseparable: the first is only coherent because of the second.

An opportunity target survives resolution.  Previously "resolving" a ToO rewrote the row from `c_type = 'opportunity'` to `'sidereal'` and nulled the region, so every trace that this had ever been a Target of Opportunity was destroyed at exactly the moment the observation became one operationally -- and the approved search region evaporated on first use.  `c_type` now records what kind of target this is and a new `c_resolved_type` records what it turned out to be, with a generated `c_target_tracking_type` deriving how it moves.  That type is a new `e_target_tracking_type` (`'sidereal'`, `'nonsidereal'`) rather than `e_target_type`, since a ToO cannot resolve to another ToO; it mirrors lucuma-core's `TrackType`.

`SchedulingMode` is a single ladder -- `Unconstrained`, `NoSplitting`, `Uninterruptible`, `Interrupting` -- each rung keeping every restriction below it.  `TooActivation` becomes derived rather than stored, from the scheduling mode plus whether the asterism holds an opportunity target, so ToO-ness is the single predicate `hasOpportunityTarget` and there is no second field to keep in step.  On `t_observation`, `c_too_activatio`n is now `GENERATED ALWAYS` over `c_has_too_target` and `c_scheduling_mode`, so it cannot drift from its inputs.  Interrupting is the one rung reserved to ToOs: an observation carrying it with an ordinary asterism is `Undefined` rather than meaningful.

The region an opportunity target is approved for is now optional on input and defaults to the whole sky, and it outlives resolution.  Nothing in this change checks a resolved target against it -- what that check should ask needs science staff input -- but the configuration approval already does so at the reference coordinates, and now that the region survives, that keeps working across resolution and any later edit.

Migrations V1255-V1258.


notes/too-scheduling-overview.md
```diff
@@ -0,0 +1,291 @@
...
+
+---
+
+## The idea in one paragraph
+
+A Target of Opportunity is an observation that waits for an **alert** rather than for
+the queue. Usually the target itself is unknown when the proposal is written — a
+supernova, a gamma-ray burst, a gravitational-wave counterpart — and all that can be
+said in advance is the region of sky it might appear in. Sometimes, though, the target
+is known perfectly well and what is unknown is *when* it will do something worth
```
cquiroz, line 13: Nice, this is a key piece of information, we are separating the time and the location, they can be set separately

notes/too-scheduling-overview.md
```diff
@@ -0,0 +1,291 @@
...
+**opportunity target**. What makes it special is not that its position is unknown, but
+that it belongs to an observation waiting on an alert. It carries a **region**: an arc
+of right ascension and an arc of declination bounding where the event may occur, which
+is part of what the TAC approves. A target that names no region — or names an arc in
+only one axis — is approved for **the whole sky** in what it left out. That is the
+weakest thing it can say rather than a missing answer, so there is no such thing as a
+ToO without a region.
+
+An opportunity target is either **unresolved** — region only, position not yet known —
+or **resolved**, carrying real coordinates or an ephemeris. Both are ordinary states:
```
cquiroz, line 68: Only coordinates or can it be a full target with PM, mags, etc?

notes/too-scheduling-overview.md
```diff
@@ -0,0 +1,291 @@
...
+
+An opportunity target may resolve to a sidereal *or* a nonsidereal target. Nothing in
+the design assumes sidereal.
+
+---
+
+## 3. What makes an observation a ToO
+
+Nothing is declared. An observation is a Target of Opportunity **exactly when its
+asterism contains an opportunity target**. Any observation is allowed to have one.
```
cquiroz, line 127: does it mean we can mix sidereal and toos in one asterism?

notes/too-scheduling-overview.md
swalker2m: It's resolved with the entire sidereal tracking information, so including proper motion.  The magnitudes, I think, are part of the source profile, which is set in anticipation before the target is even resolved. All targets, be they sidereal, nonsidereal or opportunity have a source profile.

notes/too-scheduling-overview.md
swalker2m: You could.  I'm not sure whether there is a use case for that.  Some ToOs are for objects which are already identified, in which case close by objects could conceivably be included in the same asterism.  This will be validated, if necessary, in a future PR.

toddburnside (approved):
Seems to work with my Explore changes, so LGTM.

PR #2851, opened 2026-08-15
Expose itc max adu result

cquiroz:
We can calculate the ADU after getting the results from ITC and expose them in the schema. Explore wants to display adu values.

### DIFF
--- a/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
+++ b/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
@@ -14528,6 +14528,16 @@ type ItcResult {
   the exposure time reported alongside it.
   """
   peakPixelFlux: Float
+
+  """
+  The maximum, across the CCDs, of each CCD's peak pixel flux converted to ADU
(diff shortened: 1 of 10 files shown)


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -14528,6 +14528,16 @@ type ItcResult {
...
+
+  """
+  The maximum, across the CCDs, of each CCD's peak pixel flux converted to ADU
+  using that CCD's own amplifier gain.  Because the gain differs per CCD this
+  is not in general `peakPixelFlux` divided by a single gain; the per-CCD
+  maximum is what matters for saturation.  Null under exactly the same
+  conditions as `peakPixelFlux`, and likewise computed for the unclamped
+  acquisition exposure time.
+  """
+  peakPixelAdu: Int
```
rpiaggio: Nitpick: Would it be more natural to have `peakPixel { flux adu}` instead of 2 separate properties?

rpiaggio (approved):
LGTM

PR #2854, opened 2026-08-17
Assign bands to all callibrations

rpiaggio:
## Assign a science band to calibration observations

Fixes [sc-7684]. Calibration observations must get the highest priority band of
the science observations they serve. Today a program with time in two or more
bands often ends up with no band on its calibrations, and the scheduler skips
them.

### Why it happened

When a program has time in only one band, `setAllocations` copies that band onto
every observation, so calibrations got a band by accident. With two or more
bands nothing does that, and three gaps showed up:

1. The band for GMOS calibrations was picked with `min` over `Option[ScienceBand]`.
   In that ordering "no band" sorts before every real band, so a single science
   observation without a band pulled the calibration down to no band at all.
2. Twilight observations were never given a band when created. They only picked
   one up later, on the next recalculation.
3. Tellurics and daytime pinhole flats were created with no band and never
   updated.

### Changes

- `PerProgramPerConfigCalibrationsService`: pick the lowest band among the
  science observations that actually have one, ignoring the rest. Band 1 wins
  over Band 2, which wins over Band 3, so a configuration shared by Band 2 and
  Band 3 gives a Band 2 calibration.
- `CalibrationsUtils`: twilight observations now get their band at creation,
  the same way spectrophotometric ones already did. The lookup that finds the
  science properties for a calibration configuration is now shared by both
  (`propsForConfig`), since it has to match configurations the way the role
  does, not by exact key.
- `PerScienceObservationCalibrationsService`: `syncObservationConfiguration`
  now copies `c_science_band` from the science observation. That covers
  tellurics and daytime pinhole flats, and because the sync runs on every
  recalculation, a later band change on the science observation follows through.

### Tests

- `perProgramPerConfigCalibrations`: three GMOS-N observations sharing one
  configuration, with Band 3, Band 2, and no band at all. Both calibrations
  (spectrophotometric and twilight) come out as Band 2. Moving one science
  observation to Band 1 then moves both calibrations to Band 1.
- `perScienceObservationCalibrations`: an F2 telluric takes its science
  observation's band, and follows it when the band changes.

Note on the daytime pinhole flat: it now inherits a band too, since it goes through the same sync. It's a DayCal and non-chargeable so the band should not affect time accounting. If anyone suspects this is an issue, we can special-case it to leave it unassigned.

### DIFF
--- a/modules/service/src/test/scala/lucuma/odb/graphql/feature/perProgramPerConfigCalibrations.scala
+++ b/modules/service/src/test/scala/lucuma/odb/graphql/feature/perProgramPerConfigCalibrations.scala
@@ -20,7 +20,9 @@ import lucuma.core.enums.GmosAmpReadMode
 import lucuma.core.enums.GmosRoi
 import lucuma.core.enums.ObservationWorkflowState
 import lucuma.core.enums.ObservingModeType
+import lucuma.core.enums.ScienceBand
 import lucuma.core.enums.Site
+import lucuma.core.enums.TimeAccountingCategory
 import lucuma.core.math.Coordinates
 import lucuma.core.math.Wavelength
 import lucuma.core.model.CloudExtinction
@@ -31,9 +33,11 @@ import lucuma.core.model.ProgramReference.Description
 import lucuma.core.model.Target
 import lucuma.core.model.User
 import lucuma.core.syntax.string.*
+import lucuma.core.syntax.timespan.*
 import lucuma.odb.data.EditType
 import lucuma.odb.graphql.OdbSuite
 import lucuma.odb.graphql.TestUsers
+import lucuma.odb.graphql.input.AllocationInput
 import lucuma.odb.graphql.input.ProgramPropertiesInput
 import lucuma.odb.graphql.query.ExecutionQuerySetupOperations
 import lucuma.odb.graphql.query.ExecutionTestSupportForGmos
@@ -63,7 +67,7 @@ class perProgramPerConfigCalibrations
 
   val DefaultSnAt: Wavelength = Wavelength.fromIntNanometers(510).get
 
-  override val validUsers = List(pi, service)
+  override val validUsers = List(pi, service, staff)
 
   def scienceRequirements(pi: User, oid: Observation.Id, snAt: Wavelength = DefaultSnAt): IO[Json] =
     query(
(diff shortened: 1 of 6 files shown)


modules/service/src/test/scala/lucuma/odb/graphql/feature/perProgramPerConfigCalibrations.scala
```diff
@@ -304,6 +328,39 @@ class perProgramPerConfigCalibrations
     }
   }
 
+  test("calibrations take the highest priority band of the science observations") {
+    // More than one allocated band, so no band is assigned program-wide on allocation.
```
cquiroz: wdym? the calibration gets no bad?

cquiroz:
It seems we could enforce in the db that calibrations have a science band, at least if the science obs have a band. It may require a one time migration to consider the existing observations.

rpiaggio:
> It seems we could enforce in the db that calibrations have a science band, at least if the science obs have a band. It may require a one time migration to consider the existing observations.

I ran it through Claude:

❯ would it be possible to enforce at the db level that calibrations have a band whenver their science observation have a band? do all calibrations have an associated science observation?

Let me check the actual model before answering.

  Searched for 4 patterns, read 1 file, ran 2 shell commands

Short answer: a constraint could cover tellurics and pinholes, but not the GMOS specphots — which is the case the ticket is about.

→ Tellurics and daytime pinholes do have one science observation, and the DB can see it. The link is the system group: one science observation plus its calibrations. check_obs_calibration_group_observations (V1180) is already a deferred constraint trigger enforcing that shape, so a band rule could be modeled directly on it. One wrinkle: that trigger says at most one science observation, because cleanup moves the science obs out of the group before deleting it. Zero is a legal transient state, so the constraint would have to tolerate it.

→ Spectrophotometrics and twilights have no parent observation at all. One specphot serves every science observation in the program whose configuration matches after role-specific normalization (specphot ignores ROI, twilight doesn't). It's many-to-many and computed in Scala by CalibrationConfigMatcher. In the database it's just a row sitting in the program's "Calibrations" group with nothing pointing at it. To constrain it, Postgres would have to re-implement the config-subset and normalization logic across every observing mode table.

→ And Photometric exists in the enum but nothing generates it. A calibration role is just a column value, so it can be stamped on any observation without either flow being involved. "Calibration" doesn't imply "came from the calibration service".

→ If you want the guarantee anyway, propagate instead of reject. A trigger on t_observation that copies the band down to the calibrations in the same group when a science band changes gets you the same invariant for tellurics and pinholes, without the failure mode: a rejecting constraint aborts whichever transaction trips it, and calibration recalculation writes a lot in one transaction. It would also make the band independent of the obscalc → recalculation round trip. The cost is that the rule then lives in both SQL and Scala.

→ My recommendation: leave it in Scala. The DB can only protect the half that was already easy, the half it can't protect is the one that broke, and a second implementation in SQL is one more thing to keep in sync. If you want a safety net, a periodic query that reports mismatches is cheaper than a constraint that can fail writes.

modules/service/src/test/scala/lucuma/odb/graphql/feature/perProgramPerConfigCalibrations.scala
rpiaggio: Not sure I understand the question. The calibration takes the highest priority band, that's what this test checks.

It creates a program with multiple band allocations, otherwise when the program has a single band all the observations in the program get that single band.

cquiroz: the comment is not very understandable

rpiaggio: I made it more verbose now.

cquiroz:
i was thinking along the lines of enforcing that calibrations have a band, not that the db would calculate it

rpiaggio:
> i was thinking along the lines of enforcing that calibrations have a band, not that the db would calculate it

Yes, that's what I asked Claude and it seems to have answered. The bottom line is that at the DB it's not always trivial to know what science observation(s) originated the calibrations.

cquiroz (approved):
LGTM

PR #2862, opened 2026-08-18
Shortcut 9948: explicit time request

swalker2m:
Adds an optionally specified explicit time request for a proposal.  This is purely additive and shouldn't break clients.  It's unclear whether there should be warnings if an explicit time request differs from the default sum (particularly for non-ToO proposals).  If so, that will be handled in a separate PR.

From the GPP requirements document:

> **Specify the total proposed time without defining all observations.** PIs should not have to pre-define every ToO observation at the proposal stage.

> Add an option in the Proposal tab where the user can override the sum of the defined observations with the final time that they are proposing for.

Claude's commit message below.

## Explicit proposal time request

Until now a proposal's time request was derived and nothing else: the sum of the time estimates of the program's observations, folded up the group tree in `TimeEstimateService`.  An observation without an estimate contributes nothing to that sum, which is fine while every observation is defined and wrong as soon as one isn't -- a Target of Opportunity proposal cannot write down the observations that will consume most of its time before the targets are known.

So the request becomes explicit-or-derived, following the ToO ceiling of V1245: a nullable `c_time_request` on `t_proposal`, set through `ProposalPropertiesInput.explicitTimeRequest` and exposed on `Proposal` as the explicit / default / effective triple.  It applies to every kind of proposal, so it sits at the top level of the input rather than inside `gemini`/`keck`/`subaru`.

The effective value keeps the shape of the derived one, a `CalculatedCategorizedTimeRange`, so the submission email and any client rendering the estimate read one field either way; an explicit request is a settled, degenerate range charged to program time.  Unlike the ToO ceiling it is not frozen on acceptance: the ceiling is enforced against observations at execution time, while the request authorizes nothing. `Program.timeEstimateRange` is left alone -- it means remaining execution time and is used well after phase 1.


cquiroz:
> Adds an optionally specified explicit time request for a proposal.
Time as in requested time or as execution time (though i guess that would be obs specific)

swalker2m:
> Time as in requested time or as execution time (though i guess that would be obs specific)

It's referring to the time allocation request for a proposal as a whole.  The observation and proposal time estimates are unchanged and execution time accounting is unchanged.

swalker2m:
Note also that a time request does not imply any constraints on how much time is actually allocated by partners in the ITAC.  It could be more or less than the request, just as it is today.

cquiroz (approved):
LGTM

PR #2868, opened 2026-08-19
Fix wavelength <= 0 error

cquiroz:


itc/model/src/main/scala/lucuma/itc/ItcGraph.scala
```diff
@@ -35,10 +35,15 @@ enum GraphType(val tag: String) derives Enumerated:
...
-  assert(start >= 0, "Wavelength <= 0 received in ITC graph data.")
-
   val step: Double = (end - start) / (count - 1)
 
+  // Drop the first n samples, keeping the sample spacing and the remaining wavelengths intact.
+  def drop(n: Int): ItcXAxis = ItcXAxis(start + n * step, end, count - n)
+
+  // Number of leading samples that are at or below 0 nm, i.e. have no valid Wavelength.
+  def nonPositiveCount: Int =
+    if start > 0 then 0 else (((-start) / step).floor.toInt + 1).min(count)
```
swalker2m: This seems to be the money change.  What is going on here?  Where does `start` come from?

cquiroz: in certain calculations the series start with negative or 0 wavelengths, This is an itc artifact, probably due no numeric approximation. The fix is to remove the non positive steps

We could consider also changing it in the ocs itc but we may want to have this check anyway in lucuma-itc

toddburnside: The X axis information is coded as the starting value, the ending value, and the number of points. Values for points between start and end are calculated. So, this code figures out what the first non-zero value would be based on that calculation. The `drop` method updates the start and count values to eliminate the first n points. The `fromLegacy` function also drops those points from the Y values list.

toddburnside (approved):
LGTM once CI passes.

PR #2870, opened 2026-08-19
Extract more information about masks on upload

cquiroz:
When a mos_mask attachment is uploaded, the file is now parsed with `MosMaskReader` and the extracted design is stored on the attachment as jsonb column.

`MaskDefinition` carries the header (instrument, pixel scale, pointing, position angle, dispersion direction) and the slit list (id, coordinates, geometry, offsets, tilt, placement priority), plus precomputed summary fields: science/acquisition slit counts and average science-slit width (alignment boxes excluded).


modules/schema/src/main/scala/lucuma/odb/data/MaskDefinition.scala
```diff
@@ -0,0 +1,123 @@
+// Copyright (c) 2016-2025 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.odb.data
```
toddburnside: I don't remember, is lucuma.odb.data published? If not, do these need to be in core so that explore can use them?

toddburnside (approved):
Seems fine to me.

PR #2871, opened 2026-08-19
F2 MOS acquisition

cquiroz:
image

It also remove the offset preset field to be consistent with the other modes


toddburnside (approved):
If copilot says it's OK, it's fine with me.

PR #2872, opened 2026-08-20
sc-9915 for review status

tpolecat:
This adds a `ForReview` state to the obs workflow, as well as new warnings for low S/N and low conditions probability. The change works as follows:

- Observations that are otherwise ok but have warnings cannot normally transition to `Ready`; instead they can transition to `ForReview`. 
- Users with `Staff` access or better **are** allowed to transition these observations to `Ready` via a special case in access-control checking, but `Ready` will **not** appear in the allowed transition set. The reason for this is that the caching code in the database has no real concept of who the user is, and needs to have a stable transition set.
- UI code thus needs a special case that says if a transition to `ForReview` is allowed *and* the user is staff or better, then also allow a transition to `Ready`.
- The UI should paint warnings with a "dismissed" indication if the observation has passed the `ForReview` state.

Note that many observations in test code were triggering the conditions warning, which broke a lot of tests. So I changed the default obs to ask for weaker conditions, which broke a few other tests that were dependent on very good conditions. I added `updateObservationConditions` and `setBestConditions` on the mixin `UpdateObservationsOps` trait as a way for tests to deal with this in the future.


modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql
```diff
@@ -23511,6 +23514,7 @@ enum ObservationWorkflowState {
   READY
   ONGOING
   COMPLETED
+  FOR_REVIEW
```
toddburnside: Should the order of the enum values in scala, graphql and postgres match?

tpolecat: Yeah that makes sense. I'll change it here and in the SQL.

swalker2m (approved):
This seems great to me. :+1:

modules/service/src/main/scala/lucuma/odb/service/workflow/ObservationValidator.scala
```diff
@@ -62,18 +62,21 @@ object ObservationValidator:
...
+      GhostVMagnitudeValidator       |+|
+      TooActivationValidator         |+|
+      OpportunityTargetValidator     |+|
+      OtherConfigErrorValidator      |+|
+      ConditionsProbabilityValidator
 
     val scienceValidator2: ObservationValidator =
-      ItcValidator(itcFor) |+| AcquisitionValidator(itcFor)
+      ItcValidator(itcFor)                |+| 
+      AcquisitionValidator(itcFor)        //|+|
```
toddburnside: Should the `|+|` be commented out?

cquiroz:
Could we update the documentation for the workflow states to consider the `ForReview` case?

tpolecat:
> Could we update the documentation for the workflow states to consider the `ForReview` case?

Oh, yes. i didn't see the doc file. Will do.

modules/service/src/test/scala/lucuma/odb/graphql/mutation/setObservationWorkflowState.scala
```diff
@@ -403,12 +406,60 @@ class setObservationWorkflowState
...
+  test("[Exchange]    Defined    <-> Inactive (proposal accepted)"):
     for
       po        <- createExchangeKeckObs(accepted = true)
       (pid, oid) = po
       _         <- assertIO(queryObservationWorkflowState(oid), Defined)
       _         <- testTransitions(pid, oid, Defined, Inactive)
     yield ()
 
+  // sorry
+  var HACK_ITC = false
```
toddburnside: 😬 Couldn't you just make a new local function to call when you need hacked results rather than use a var?

toddburnside (approved):
LGTM
One thing that isn't clear to me is what happens if a staff or better user sets it to Ready, thereby `accepting` the warnings, but then the observation is updated and different warnings are generated?

modules/service/src/test/scala/lucuma/odb/graphql/mutation/setObservationWorkflowState.scala
tpolecat: It's called from the shared itc client that's used to construct the test server, so it's not easy to work around. I'll see if I can improve it.

toddburnside: If it's a problem, then no worries.

modules/service/src/main/scala/lucuma/odb/service/workflow/ObservationValidator.scala
tpolecat: doh

cquiroz:
Are warning shown already at proposal time? I'd think PIs may want to see them

tpolecat:
> Are warning shown already at proposal time? I'd think PIs may want to see them

Yes, they show up as non-fatal vaildation errors.

cquiroz:
> Oh, yes. i didn't see the doc file. Will do.

Forgot the file has an svg image, we could generate it after the PR is merged

tpolecat:
> One thing that isn't clear to me is what happens if a staff or better user sets it to Ready, thereby accepting the warnings, but then the observation is updated and different warnings are generated?

At this point nothing happens, which I guess is a problem. I'm not sure how much people mess with observations once they're set to `Ready`. I think I was assuming that Explore disabled edits at that point. I'll think about it and add an issue.

toddburnside:
Another question. What in the API indicates an ObservationValidation is a warning vs an error? I only see `code` and `messages` in ObservationValidation.

tpolecat:
> Another question. What in the API indicates an ObservationValidation is a warning vs an error? I only see `code` and `messages` in ObservationValidation.

The code has an associated severity, which is `NonFatal` for warnings.

toddburnside:
> > Another question. What in the API indicates an ObservationValidation is a warning vs an error? I only see `code` and `messages` in ObservationValidation.
> 
> The code has an associated severity, which is `NonFatal` for warnings.

OK. I see. I get that from the scala enum, but not from the API.

toddburnside:
I very small piece of this PR, including part of the migration, have been incorporated into https://github.com/gemini-hlsw/lucuma-odb/pull/2896 in order to unblock progress.

swalker2m:
From ToO documentation (from Bryan):

> Some ToO observations may need custom sequences that must be checked manually. We require a way to specify that these are checked and but not ready to execute. 
>
> Related to the warnings and ‘For Review’ discussion. The checked observations could be set to Defined, or we may need to bring back the On Hold workflow state. A PI may then duplicate a Defined/On Hold observation, or make a new observation with the same config, and then set it to Ready.

I don't know if we want to address this now as part of this PR, or leave it to another iteration.  As is, we wouldn't support a "checked but not ready" state if I understand it correctly.

PR #2873, opened 2026-08-20
Shortcut 9747: Trigger update

swalker2m:
This PR makes two updates to ToO triggers.

## Superseded

Updates ToO triggers with a new status: `SUPERSEDED`.  This allows the scheduling mode of an observation to change after a trigger, resulting in superseding an existing trigger.  In particular, a `Rapid` ToO can turn into an `Interrupting` ToO by changing the observation's `SchedulingMode` to `Interrupting`.  When the scheduling mode of an already triggered observation is changed, an outstanding ToO request is superseded by the new request.  Old requests are not deleted but are no longer considered live.  A full audit trail of past triggers is available by following the `supersedes` reference on `TooTrigger`.

From the ToO documentation:

> The interrupt occurs when an observation/group with the interrupt flag selected is set to Ready, or when an observation that is Ready has its interrupt flag set.  An interrupt will cause the system to prompt the observer to abort the current observation and slew immediately to the new observation/group (see the Inception document). A new observing plan is then generated.

Here setting the "interrupt flag" is changing the `SchedulingMode` to `Interrupting`.  ToO triggers generate events which subscribers may respond to (e.g., to "slew immediately").

## Withdrawn  

Plugs a hole in which the ToO ceiling for a program could be reset by staff and yet leave outstanding triggers at a higher activation value.  When the ceiling is changed, existing triggers for the program that violate the ceiling are now withdrawn.

### DIFF
--- a/modules/service/src/main/resources/db/migration/V1267__too_trigger_activation.sql
+++ b/modules/service/src/main/resources/db/migration/V1267__too_trigger_activation.sql
@@ -1,0 +1,240 @@
+-- Records, on the trigger itself, the ToO activation it was requested at.
+--
+-- The UI needs trigger events to carry the activation and to be filterable on
+-- it, and needs a change of activation on an outstanding request to produce an
+-- event of its own.  All three fall out of putting the activation on the row.
+--
+-- A CHANGE OF ACTIVATION IS A NEW REQUEST
+--
+-- Triggering a standard ToO, a rapid one and an interrupting one provoke very
+-- distinct notification behaviour downstream: who is told, how fast, and what
+-- they are expected to drop.  They are effectively different things, so a
+-- request's identity does not survive a change in its character.  The
+-- outstanding row is superseded and a successor takes its place, with a new id
+-- and a new c_requested_at -- which is correct rather than lossy, since the clock
+-- that matters is how long *this* request, at *this* activation, has been
+-- outstanding.  c_supersedes chains them, so the root of a chain still answers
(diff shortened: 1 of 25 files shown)


modules/service/src/main/resources/db/migration/V1267__too_trigger_activation.sql
```diff
@@ -0,0 +1,240 @@
...
+COMMENT ON COLUMN t_too_trigger.c_supersedes IS
+  'The request this one replaced, null for a first request. The root of the chain '
+  'is when the observation first went live at any activation.';
+
+COMMENT ON TABLE t_too_trigger IS
+  'One row per attempt to activate a ToO observation, maintained by '
+  'too_trigger_track_ready() from the observation user state and its derived '
+  'activation. At most one row per observation is requested at a time '
+  '(i_too_trigger_active); declined, withdrawn and superseded attempts accumulate '
+  'as history.';
```
tpolecat: Thank you for adding comments on these objects. I need to remember to do this.

tpolecat (approved):
I am in awe of your ability to get through eye-watering complexity with no outward degenerative effects. This looks really good. 👍🏻

toddburnside (approved):
LGTM
It doesn't seem like this will affect explore at this point, right?

swalker2m:
> It doesn't seem like this will affect explore at this point, right?

I don't think so.  This one was more about the trigger objects themselves and the events they produce.

